PR #75, opened 2022-12-07
Fix use of CredentialUsage.Provisioner for locally built plugins

jpcoenen:
There are a few problems with how the `Provisioner` field of the `CredentialUsage` struct is currently exposed for locally built plugins:
- The `Executable` field of `ProvisionerID` is a pointer, which means it has a different value for each invocation of a plugin. This is especially problematic because `ProvisionerID` is used as the key of a map.
- `ProvisionerID` contains an unused `Plugin` field.
- `Credential` in `ProvisionerID` is not guaranteed to be unique.
- The `Provisioner` field of `CredentialUsage` is not set to `nil` before transferring it over RPC.
- Whether or not the `Provisioner` field is set is not encoded in `proto.GetPluginResponse`. This

This all combined led to plugins using this field panicking on startup and the `Provision` field not being accessible by the CLI.

To fix this, the `ProvisionerID` has been rewritten to not use any pointers, but use slice indices instead and `proto.GetPluginResponse` now includes `CredentialUsageHasProvisioner` which encodes whether the `Provisioner` field is set.

Because these changes are not backwards-compatible, `proto.Version` has been bumped to `3`.

### DIFF
--- a/sdk/rpc/proto/protocol.go
+++ b/sdk/rpc/proto/protocol.go
@@ -27,18 +27,32 @@ func (c CredentialID) String() string {
 	return fmt.Sprintf("plugin.Credentials[%d]", c)
 }
 
+// CredentialUsageID uniquely identifies a CredentialUsage within a plugin.
+type CredentialUsageID struct {
+	Executable ExecutableID
+	Usage      int
+}
+
+func (c CredentialUsageID) String() string {
+	return fmt.Sprintf("%s.Uses[%d]", c.Executable, c.Usage)
+}
+
 // ProvisionerID uniquely identifies a provisioner within a plugin.
 type ProvisionerID struct {
-	Plugin     string
-	Credential sdk.CredentialName
-	Executable *ExecutableID
+	// IsCredentialProvisioner is set to true if the ProvisionerID identifies the DefaultProvisioner of a credential.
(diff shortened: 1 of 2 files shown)


sdk/rpc/proto/protocol.go
```diff
@@ -27,18 +27,32 @@ func (c CredentialID) String() string {
...
+}
+
 // ProvisionerID uniquely identifies a provisioner within a plugin.
 type ProvisionerID struct {
-	Plugin     string
-	Credential sdk.CredentialName
-	Executable *ExecutableID
+	// IsCredentialProvisioner is set to true if the ProvisionerID identifies the DefaultProvisioner of a credential.
+	// It is set to false if the ProvisionerID identifies the Provisioner of an executable's CredentialUsage.
+	IsCredentialProvisioner bool
```
Marton6, line 44: Technically all provisioners are "credential provisioners". Therefore, I'd say we name this `IsDefaultProvisioner`

Marton6 (commented):
Thanks for the fix, @jpcoenen!

Could we make sure we merge this to a `protocol-v3` branch which we will merge to main when we release the corresponding protocol update for the CLI?

sdk/rpc/proto/protocol.go
jpcoenen: Makes sense :+1:

Marton6 (approved):

PR #85, opened 2022-12-08
Created shell plugin for Tugboat.

markdorison:
Created shell plugin for [Tugboat CLI](https://docs.tugboatqa.com/tugboat-cli/).

### DIFF
--- a/plugins/tugboat/access_token.go
+++ b/plugins/tugboat/access_token.go
@@ -2,7 +2,6 @@ package tugboat
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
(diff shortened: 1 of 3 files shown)


SimonBarendse (approved):
Thank you Mark! This looks excellent

plugins/tugboat/access_token.go
```diff
@@ -0,0 +1,77 @@
...
+					Length: 32,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.TempFile(tugboatConfig, provision.AtFixedPath("~/.tugboat.yml")),
```
florisvdg, line 35: One downside of using the file provisioner with a fixed path is that it users won't be able to test drive this plugin without overwriting their existing configuration.

Seems that Tugboat also reads from the [`TUGBOAT_API_TOKEN` environment variable](https://github.com/TugboatQA/docker-tugboat-cli#read-the-token-using-an-environment-variable), so if that works well then that may be the more suitable option.

florisvdg (commented):
Thanks for the contribution!

plugins/tugboat/access_token.go
markdorison: @florisvdg I was having trouble finding that documentation; thanks for sharing! Updating with your suggestion.

DCKcode:
Thanks for your contribution 😄

DCKcode (approved):

markdorison:
My pleasure; this was fun!

SimonBarendse (approved):

PR #87, opened 2022-12-09
hcloud: init plugin

shyim:
Adds support for [hcloud](https://github.com/hetznercloud/cli) from [Hetzner](https://www.hetzner.com/de/) well known hosting provider in Germany also expanding to US right now

### DIFF
--- a/plugins/hcloud/api_token_test.go
+++ b/plugins/hcloud/api_token_test.go
@@ -1,0 +1,26 @@
+package hcloud
+
+import (
+	"testing"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func TestImportingTheHetznerConfig(t *testing.T) {
+	expectedFields := map[sdk.FieldName]string{
+		fieldname.Token: "dcAuOpQaCNvjzsNPmeGXvegHBdq4Zamx8QjI8ibxfErzy34fjL4ZOITFvdP5SKct",
+	}
+
+	plugintest.TestImporter(t, APIToken().Importer, map[string]plugintest.ImportCase{
+		"Hcloud config file": {
+			Files: map[string]string{
+				"~/.config/hcloud/cli.toml": plugintest.LoadFixture(t, "hcloud.toml"),
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{Fields: expectedFields, NameHint: ""},
+			},
+		},
+	})
+}
--- a/plugins/hcloud/api_token.go
+++ b/plugins/hcloud/api_token.go
@@ -1,0 +1,71 @@
+package hcloud
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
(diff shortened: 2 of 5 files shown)


plugins/hcloud/api_token_test.go
```diff
@@ -0,0 +1,25 @@
...
+		fieldname.Token: "dcAuOpQaCNvjzsNPmeGXvegHBdq4Zamx8QjI8ibxfErzy34fjL4ZOITFvdP5SKct",
+	}
+
+	plugintest.TestImporter(t, APIToken().Importer, map[string]plugintest.ImportCase{
+		"Hcloud config file": {
+			Files: map[string]string{
+				"~/.config/hcloud/cli.toml": plugintest.LoadFixture(t, "hcloud.toml"),
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{Fields: expectedFields, NameHint: "default"},
```
florisvdg, line 21: Small thing: it would be nice for the importer to omit the name hint if it's `default`. It will get added to the 1Password item title, but in most cases, having the word "default" there doesn't provide a lot of value.

florisvdg (commented):
Nice! And thanks for adding a test as well.

plugins/hcloud/api_token_test.go
shyim: In my example it is `default`.

In hcloud you can have multiple "context" which is similar to profile in AWS config. Each context can have an own token, so It made sense for me to show that name to the user

DCKcode (approved):

plugins/hcloud/api_token_test.go
florisvdg: Yes, so to clarify: it's perfect that you're populating the name hint with the profile name! Only suggestion is to omit the name hint if the profile name is `default`. Does that make sense?

SimonBarendse: You can use https://github.com/1Password/shell-plugins/pull/105 for this when it's merged.

SimonBarendse: #105 is merged.

shyim:
Is fixed. Sorry for the delay :)

DCKcode (approved):

SimonBarendse (approved):

notDavid:
Hi there, noob question;

when following the steps [here](https://developer.1password.com/docs/cli/shell-plugins/hetzner-cloud/#step-3-use-the-cli) i run:
```
$ hcloud context create my-project
The HCLOUD_TOKEN environment variable is set. Do you want to use the token from HCLOUD_TOKEN for the new context? (Y/n):
$ y
```
and then my API key gets permanently stored **in plain text** in `/Users/david/.config/hcloud/cli.tom`

Is this expected / intended behavior ?

shyim:
yes. you need to import the key into 1password vault and it overrides it then. `op plugin init hcloud`

notDavid:
@shyim Yep, i had done that, as i followed the steps in the guide i had linked to:

`rm /Users/david/.config/hcloud/cli.tom`

Guide:
```
op signin
op plugin init hcloud
# search in Password > "Prompt me for each new terminal session"
source ~/.config/op/plugins.sh
echo "source ~/.config/op/plugins.sh" >> ~/.bashrc && source ~/.bashrc
```

(So, it still saves the password in plain text.)

notDavid:
For the record, it's the last step that saves the password in plain text:

```
hcloud context create my-project
# i then get prompted for TouchID
```

notDavid:
@shyim Should i create a new issue for this?

PR #88, opened 2022-12-09
run ci also for pull requests

shyim:
Currently CI is not running for Pull Requests

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,5 +1,11 @@
 name: Test and Validate
-on: [push]
+on:
+  push:
+    paths:
+      - "**.go"
+  pull_request:
+    paths:
+      - "**.go"
 
 jobs:
   build:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,11 +1,13 @@
 name: Test and Validate
 on:
   push:
-    paths:
-      - "**.go"
+    paths-ignore:
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/**'
   pull_request:
-    paths:
-      - "**.go"
+    paths-ignore:
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/**'
 
 jobs:
   build:


.github/workflows/test.yaml
```diff
@@ -1,5 +1,11 @@
 name: Test and Validate
-on: [push]
+on:
+  push:
+    paths:
+      - "**.go"
+  pull_request:
+    paths:
+      - "**.go"
```
florisvdg, line 8: There are a few other relevant files where you'd want this to trigger, such as `.github/workflows/test.yaml`, `.golangci.yml`, `go.mod`, `go.sum`, `Makefile`, so maybe easier to [reverse](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) the pattern?

```suggestion
on:
  push:
    paths-ignore:
      - '**.md'
      - '.github/ISSUE_TEMPLATE/**'
  pull_request:
    paths-ignore:
      - '**.md'
      - '.github/ISSUE_TEMPLATE/**'
```

SimonBarendse (approved):
Thanks @shyim!

.github/workflows/test.yaml
```diff
@@ -1,5 +1,13 @@
 name: Test and Validate
-on: [push]
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+      - '.github/ISSUE_TEMPLATE/**'
+  pull_request:
```
SimonBarendse, line 7: We can run on pull requests, also from forks, because there are no secrets required to run this job.

If we were to add tests that do require secrets in the future (e.g. integration or end-to-end tests) we need to figure out a strategy of how we run those, as [secrets are not passed to the runner when a workflow is triggered from a forked repository](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow). We can come up with a strategy when we get there.

SimonBarendse: If we were to add tests that require a secret without updating this rule, the job would simply fail on the missing secret.

shyim: You can't run those tests together with forks. It would make sense to have then two workflow files.

But maybe mocking can avoid that 🤔

florisvdg (approved):

.github/workflows/test.yaml
DCKcode: This is all quite alright yeah. Thanks for the contribution 😄

DCKcode (approved):

PR #94, opened 2022-12-09
feat: add wrangler

shyim:
Adds support for Wrangler

- Fine gained PAT (API Token) works
- API Key with Email IDK? I can't choose it currently


plugins/wrangler/api_token.go
```diff
@@ -0,0 +1,51 @@
...
+
+func APIToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIToken,
+		DocsURL:       sdk.URL("https://developers.cloudflare.com/workers/wrangler/system-environment-variables/"),
+		ManagementURL: sdk.URL("https://dash.cloudflare.com/profile/api-tokens"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.AccountID,
+				MarkdownDescription: "The account ID for the Workers related account, can be found in the Cloudflare dashboard, can usually be inferred by Wrangler.",
```
SimonBarendse, line 20: > can usually be inferred by Wrangler

This suggest we should add `Optional: true`

SimonBarendse (commented):
Thank you yet again, @shyim! :raised_hands:

I have left one small question for your consideration.

DCKcode (approved):

plugins/wrangler/api_token.go
shyim: done

plugins/wrangler/api_token.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIToken,
+		DocsURL:       sdk.URL("https://developers.cloudflare.com/workers/wrangler/system-environment-variables/"),
```
florisvdg, line 15: I think this docs page would make more sense for the Cloudflare API token:
```suggestion
		DocsURL:       sdk.URL("https://developers.cloudflare.com/fundamentals/api/get-started/create-token/"),
```

plugins/wrangler/api_token.go
```diff
@@ -0,0 +1,52 @@
...
+		ManagementURL: sdk.URL("https://dash.cloudflare.com/profile/api-tokens"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.AccountID,
+				MarkdownDescription: "The account ID for the Workers related account, can be found in the Cloudflare dashboard, can usually be inferred by Wrangler.",
+				Secret:              true,
```
florisvdg, line 21: I don't think this is secret, right? Making it secret will also make it concealed in the 1Password GUI.
```suggestion
				Secret:              false,
```

plugins/wrangler/api_token.go
```diff
@@ -0,0 +1,52 @@
...
+			{
+				Name:                fieldname.Token,
+				MarkdownDescription: "The API token for your Cloudflare account, can be used for authentication for situations like CI/CD, and other automation.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
```
florisvdg, line 40: Don't these have a fixed length? Same for the account ID.

plugins/wrangler/api_token.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMappingApiToken),
+		Importer:           importer.TryEnvVarPair(defaultEnvVarMappingApiToken),
+	}
+}
+
+var defaultEnvVarMappingApiToken = map[string]sdk.FieldName{
+	"CLOUDFLARE_ACCOUNT_ID": fieldname.AccountID,
+	"CLOUDFLARE_API_TOKEN":  fieldname.Token,
+}
```
florisvdg, line 52: Nit: In other plugins we've been using `defaultEnvVarMapping` as the var name. Any particular reason to deviate from that?

If not, then I'd suggest:
```suggestion
		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
		Importer:           importer.TryEnvVarPair(defaultEnvVarMapping),
	}
}

var defaultEnvVarMapping = map[string]sdk.FieldName{
	"CLOUDFLARE_ACCOUNT_ID": fieldname.AccountID,
	"CLOUDFLARE_API_TOKEN":  fieldname.Token,
}
```

florisvdg (commented):
Another great addition, thanks!

plugins/wrangler/api_token.go
shyim: It's nothing documented, So I don't wanted to add a limitation

florisvdg: I'd be okay with using what we know now. Nothing bad will happen if it turns out that the length is incorrect.

SimonBarendse: > Nothing bad will happen if it turns out that the length is incorrect.

This is probably worth documenting. @florisvdg shall we add this to https://developer.1password.com/docs/cli/shell-plugins/contribute/#credential-information-and-schema ?

DCKcode (approved):

PR #95, opened 2022-12-10
Add Lacework plugin

colinbarr:
Adds support for the [Lacework CLI](https://docs.lacework.com/cli/), including import from the config file produced by `lacework configure`.

```
➜  shell-plugins git:(lacework) make lacework/validate
# BETA NOTICE: The plugin ecosystem is in beta and is subject to change.
# You may have to update or recompile your local builds every now and then to keep them
# compatible with the 1Password CLI updates.

go run cmd/contrib/main.go lacework/validate
# Plugin: lacework

✔ Has plugin name set
✔ Plugin name only using lowercase characters or digits
✔ Plugin name not longer than 20 characters
✔ Has platform name set
✔ Has platform homepage URL set
✔ Has a credential type or executable defined
✔ Has no more than one credential type defined. Plugins with multiple credential types are not supported yet
✔ Has no more than one executable defined. Plugins with multiple executables are not supported yet

# Credential: API Key

✔ Has name set
✔ Name is using title case
✔ Has documentation URL set
✔ Has management URL set
✔ Has at least 1 field
✔ All fields have name set
✔ All field names are using title case
✔ All fields have a description set
✔ All specified value compositions are valid
✔ Has at least 1 field that is secret
✔ Has a provisioner set
✔ Has an importer set

# Executable: Lacework CLI

✔ Has name set
✔ Has documentation URL set
✔ Has specified which commands need authentication
✔ Has executable command set
✔ Has a credential type defined
```

```
➜  shell-plugins git:(lacework) make test
go test ./...
?   	github.com/1Password/shell-plugins/cmd/contrib	[no test files]
?   	github.com/1Password/shell-plugins/cmd/contrib/build	[no test files]
ok  	github.com/1Password/shell-plugins/plugins	(cached)
ok  	github.com/1Password/shell-plugins/plugins/aws	(cached)
ok  	github.com/1Password/shell-plugins/plugins/circleci	(cached)
ok  	github.com/1Password/shell-plugins/plugins/datadog	(cached)
ok  	github.com/1Password/shell-plugins/plugins/digitalocean	(cached)
?   	github.com/1Password/shell-plugins/plugins/fossa	[no test files]
ok  	github.com/1Password/shell-plugins/plugins/github	(cached)
ok  	github.com/1Password/shell-plugins/plugins/gitlab	(cached)
ok  	github.com/1Password/shell-plugins/plugins/heroku	(cached)
ok  	github.com/1Password/shell-plugins/plugins/lacework	(cached)
ok  	github.com/1Password/shell-plugins/plugins/mysql	(cached)
ok  	github.com/1Password/shell-plugins/plugins/okta	(cached)
ok  	github.com/1Password/shell-plugins/plugins/postgresql	(cached)
ok  	github.com/1Password/shell-plugins/plugins/sentry	(cached)
ok  	github.com/1Password/shell-plugins/plugins/snyk	(cached)
ok  	github.com/1Password/shell-plugins/plugins/stripe	(cached)
?   	github.com/1Password/shell-plugins/plugins/tugboat	[no test files]
ok  	github.com/1Password/shell-plugins/plugins/twilio	(cached)
ok  	github.com/1Password/shell-plugins/plugins/vault	(cached)
ok  	github.com/1Password/shell-plugins/sdk	(cached)
ok  	github.com/1Password/shell-plugins/sdk/example	(cached)
?   	github.com/1Password/shell-plugins/sdk/importer	[no test files]
?   	github.com/1Password/shell-plugins/sdk/needsauth	[no test files]
ok  	github.com/1Password/shell-plugins/sdk/plugintest	(cached)
?   	github.com/1Password/shell-plugins/sdk/provision	[no test files]
?   	github.com/1Password/shell-plugins/sdk/rpc/proto	[no test files]
?   	github.com/1Password/shell-plugins/sdk/rpc/server	[no test files]
ok  	github.com/1Password/shell-plugins/sdk/schema	(cached)
?   	github.com/1Password/shell-plugins/sdk/schema/credname	[no test files]
?   	github.com/1Password/shell-plugins/sdk/schema/fieldname	[no test files]
```

### DIFF
--- a/plugins/lacework/api_key.go
+++ b/plugins/lacework/api_key.go
@@ -1,0 +1,109 @@
+package lacework
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+	"github.com/BurntSushi/toml"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://docs.lacework.com/console/api-access-keys"),
+		ManagementURL: sdk.URL("https://login.lacework.net/ui?redirectUrl=/investigation/settings/apikeys"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Account,
+				MarkdownDescription: "The subdomain used to access your Lacework account.",
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Specific:  []rune{'.'},
+					},
+				},
+			},
+			{
+				Name:                fieldname.APIKeyID,
+				MarkdownDescription: "The API key used to authenticate to Lacework.",
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Uppercase: true,
+						Digits:    true,
+						Specific:  []rune{'-'},
+					},
+				},
+			},
+			{
+				Name:                fieldname.APISecret,
+				MarkdownDescription: "The API secret used to authenticate to Lacework.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 33,
+					Prefix: "_",
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryEnvVarPair(defaultEnvVarMapping),
+			TryLaceworkConfigFile(),
+		)}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"LW_ACCOUNT":    fieldname.Account,
+	"LW_API_KEY":    fieldname.APIKeyID,
+	"LW_API_SECRET": fieldname.APISecret,
+}
+
+func TryLaceworkConfigFile() sdk.Importer {
+	return importer.TryFile("~/.lacework.toml", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		parsedFile := make(map[string]toml.Primitive)
+		metaData, err := toml.Decode(string(contents), &parsedFile)
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		for profileName, rawConfig := range parsedFile {
+			var config ProfileConfig
+			err = metaData.PrimitiveDecode(rawConfig, &config)
+			if err != nil {
+				continue // skip sections that don't define credentials
+			}
+
+			if profileName == "default" {
+				profileName = ""
+			}
+
+			out.AddCandidate(sdk.ImportCandidate{
+				Fields: map[sdk.FieldName]string{
+					fieldname.Account:   config.Account,
+					fieldname.APIKeyID:  config.APIKey,
+					fieldname.APISecret: config.APISecret,
+				},
+				NameHint: profileName,
+			})
+		}
+	})
+}
+
+type Config struct {
+	Profiles map[string]ProfileConfig
+}
+
+type ProfileConfig struct {
+	Account   string `toml:"account"`
+	APIKey    string `toml:"api_key"`
+	APISecret string `toml:"api_secret"`
+}
--- a/plugins/lacework/api_key_test.go
+++ b/plugins/lacework/api_key_test.go
@@ -1,0 +1,71 @@
+package lacework
+
+import (
+	"testing"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func TestAPIKeyImporter(t *testing.T) {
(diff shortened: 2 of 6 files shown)


florisvdg (commented):
Great contribution, thanks a lot! Left a few small comments.

plugins/lacework/api_key.go
```diff
@@ -0,0 +1,106 @@
...
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Specific:  []rune{'.'},
+					},
+				},
+			},
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "The API key used to authenticate to Lacework.",
```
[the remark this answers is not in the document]
colinbarr: Really good point, and one that I couldn't make my mind up on easily either. I ended up going for field names that matched the contents of the CLI config file, for example:
```
[default]
  account = "example"
  api_key = "EXAMPLE_1234567890ABCDE1EXAMPLE1EXAMPLE123456789EXAMPLE"
  api_secret = "_aaaaaaaaaaaaaaaaaaaaaaaaaexample"
  version = 2
```
The CLI itself uses slightly different terminology again:
```
➜  ~ lacework configure
▸ Account: example
▸ Access Key ID: EXAMPLE_1234567890ABCDE1EXAMPLE1EXAMPLE123456789EXAMPLE
▸ Secret Access Key: *********************************
```
Happy to go with your suggestion - does the above change your opinion?

plugins/lacework/api_key.go
```diff
@@ -0,0 +1,106 @@
...
+				continue // skip sections that don't define credentials
+			}
+
+			out.AddCandidate(sdk.ImportCandidate{
+				Fields: map[sdk.FieldName]string{
+					fieldname.Account:   config.Account,
+					fieldname.APIKey:    config.APIKey,
+					fieldname.APISecret: config.APISecret,
+				},
+				NameHint: profile,
```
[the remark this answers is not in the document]
colinbarr: Good suggestion, I've done that now.

plugins/lacework/api_key.go
```diff
@@ -0,0 +1,106 @@
...
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://docs.lacework.com/console/api-access-keys"),
+		ManagementURL: sdk.URL("https://login.lacework.net/ui?redirectUrl=/investigation/settings/apikeys"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Account,
+				MarkdownDescription: "The account ID (subdomain) used to access Lacework.",
```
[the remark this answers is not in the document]
colinbarr: Updated, thanks!

plugins/lacework/api_key_test.go
```diff
@@ -0,0 +1,71 @@
...
+			ItemFields: map[sdk.FieldName]string{
+				fieldname.Account:   "example",
+				fieldname.APIKey:    "EXAMPLE_1234567890ABCDE1EXAMPLE1EXAMPLE123456789EXAMPLE",
+				fieldname.APISecret: "_aaaaaaaaaaaaaaaaaaaaaaaaaexample",
+			},
+			ExpectedOutput: sdk.ProvisionOutput{
+				Environment: map[string]string{
+					"LW_ACCOUNT":    "example",
+					"LW_API_KEY":    "EXAMPLE_1234567890ABCDE1EXAMPLE1EXAMPLE123456789EXAMPLE",
+					"LW_API_SECRET": "_aaaaaaaaaaaaaaaaaaaaaaaaaexample",
```
florisvdg, line 66: Would be nice to use more realistic values here. We have a Makefile command to easily generate those:
```
make lacework/example-secrets
```

colinbarr: This works, in part, for the API secret since it has a fixed/known length (32 hex characters, prefixed with `_`), but this isn't true for the other two fields. I've taken the API key (ID) value from [here](https://docs.lacework.com/cli/#configure-the-cli), but the prefix before the `_` is variable so the total length isn't deterministic.

For the secret, I've got an "expected" format, with the exception of the `example` suffix not be hex, but I was following the pattern that `%/example-secrets` produces to try to be consistent.

```
➜  shell-plugins git:(lacework) make lacework/example-secrets
go run cmd/contrib/main.go lacework/example-secrets
API Key:
  Account:
  API Key:
  API Secret: _6wz23kv3ry45eg78xwd6m6_vqexample
```

Is there another way to handle this within the current framework?

florisvdg: I think the account and API key values you're using now are fine. If you can update the API secret test values with this example value then we should be good to go!

colinbarr: Perfect, thanks! What do you think about using something like this instead, so that the example is more correct/representative? The `_` in the middle of the previous example isn't quite correct since it's only used as a prefix (nor are some of the characters in `example`, but I think that's clear enough).

```
➜  shell-plugins git:(lacework) echo -n '_'; openssl rand -hex 16 | cut -c8- | tr -d '\n'; echo 'example'
_72a101a5e009abc47c32c6ff9example
```

florisvdg: Yep, looks good to me. And FYI, if the `_` always is the prefix, you can add `Prefix: "_"` to the struct.

colinbarr: Great, thank you. I've got that there already and I also included `Specific:  []rune{'_'}` in the charset. I assumed this was required since it was used in the prefix, but perhaps it's not?

florisvdg: Ah right, then you don't need the `Specific` entry.

colinbarr: Will remove that and push the newly generated secrets too.

colinbarr:
> Great contribution, thanks a lot! Left a few small comments.

Thanks for the review! All comments resolved now.

SimonBarendse (approved):

DCKcode (approved):

PR #98, opened 2022-12-11
Fix charset validation for `ValueComposition`

micnncim:
It'll be possible that a credential consists of all possible types of characters. The validation should report an error if none is allowed instead.


florisvdg (approved):

jpcoenen (approved):
Thanks for addressing @micnncim! 🙌

PR #99, opened 2022-12-12
sentry-cli file importer

volodymyrZotov:
This PR adds `senry-cli` file importer.

The config file is looked for upwards from the current path and defaults from `~/.sentryclirc` are always loaded.
The config file uses standard INI syntax.


plugins/sentry/test-fixtures/.sentryclirc
```diff
@@ -0,0 +1,7 @@
+[defaults]
+url = https://mysentry.invalid/
```
florisvdg, line 2: May be nicer to use:
```suggestion
url = https://sentry.acme.com/
```

plugins/sentry/test-fixtures/.sentryclirc
```diff
@@ -0,0 +1,7 @@
...
+project = my-project
+org = my-org
```
florisvdg, line 4: Here too:
```suggestion
org = acme
```

florisvdg (approved):

SimonBarendse (approved):
Thank you! :raised_hands:

PR #100, opened 2022-12-12
Add template for credential importer and provisioner tests

hculea:
This PR adds tests for credential the default credential provisioner and credential importers when generating a new plugin using `make new-plugin`.

For testing, run `make new-plugin` two times. First, omit the credential example, second time, add it. The template should be able to gracefully handle both cases. 

I have a few open questions about wrapping up this MR, which I'll be adding in comments.

### DIFF
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -471,7 +471,7 @@ func Test{{ .CredentialNameUpperCamelCase }}Importer(t *testing.T) {
 		// and fill the necessary details in the test template below.
 		"config file": {
 			Files: map[string]string{
-				// "~/path/to/config/file.yml": plugintest.LoadFixture(t, "your_local_fixture.yaml"),
+				// "~/path/to/config.yml": plugintest.LoadFixture(t, "config.yml"),
 			},
 			ExpectedCandidates: []sdk.ImportCandidate{
 			// 	{
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -467,7 +467,7 @@ func Test{{ .CredentialNameUpperCamelCase }}Importer(t *testing.T) {
 				},
 			},
 		},
-		// TODO: If you implemented a config file importer, add a test file example in {{ .Name }}/fixtures
+		// TODO: If you implemented a config file importer, add a test file example in {{ .Name }}/test-fixtures
 		// and fill the necessary details in the test template below.
 		"config file": {
 			Files: map[string]string{
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"html/template"
 	"log"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 	"unicode"
 
 	"github.com/1Password/shell-plugins/plugins"
@@ -156,6 +158,7 @@ func newPlugin() error {
 		IsNewCredentialName          bool
 		CredentialNameUpperCamelCase string
 		CredentialNameSnakeCase      string
+		TestCredentialExample        string
 	}{}
 
 	err := survey.Ask(questionnaire, &result)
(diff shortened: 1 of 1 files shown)


cmd/contrib/main.go
```diff
@@ -165,6 +168,17 @@ func newPlugin() error {
...
+	} else {
+		// If no example credential has been provided, a random 30-character alphanumeric
+		// value is generated to be used within the test suite.
+		lettersAndDigits := []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890")
+		rand.Seed(time.Now().UnixNano())
+		randomString := make([]rune, 30)
+		for i := range randomString {
+			randomString[i] = lettersAndDigits[rand.Intn(len(lettersAndDigits))]
+		}
+		result.TestCredentialExample = string(randomString)
```
hculea, line 181: If no value is provided as credential example, what do we want to use as a place-holder? I'm thinking any string that's randomised enough should be good enough (didn't use the `crypto` lib here - I don't see the need for cryptographic security for testing purposes).

cmd/contrib/main.go
```diff
@@ -414,6 +429,66 @@ func Try{{ .PlatformNameUpperCamelCase }}ConfigFile() sdk.Importer {
...
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{
+					Fields: map[sdk.FieldName]string{
+						fieldname.{{ .FieldName }}: "{{ .TestCredentialExample }}",
+					},
+				},
+			},
+		},
+		// TODO: If you implemented a config file importer, add a test file example in {{ .Name }}/fixtures
```
hculea, line 473: I think the best way to go here is to instruct the user to add their own fixture, rather than adding a default one that just never gets used. If the user doesn't add a file importer, it's another responsibility for them to delete it.

Any thoughts?

cmd/contrib/main.go
```diff
@@ -165,6 +168,17 @@ func newPlugin() error {
 
 	if result.ExampleCredential != "" {
 		result.ValueComposition = getValueComposition(result.ExampleCredential)
+		result.TestCredentialExample = result.ExampleCredential
```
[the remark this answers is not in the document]
florisvdg: The example secret should only be used to generate the value composition struct, which can then be used for generating test fixtures.

cmd/contrib/main.go
florisvdg: No need to care about it in this package, we already have a function in the `plugintest` package that handles this:
```go
plugintest.ExampleSecretFromComposition(schema.ValueComposition{
	Uppercase: true,
	Lowercase: true,
	Digits:    true,
})
```

cmd/contrib/main.go
florisvdg: Yeah, I think so too. We could consider to already create the `test-fixtures` dir, though.

cmd/contrib/main.go
florisvdg: Run `make aws/example-secrets` to see this in action.

cmd/contrib/main.go
hculea: I think `git` only indexes files, not directories. We could add a `.keep` file to it, but then we might as well add the fixture directly.

I propose we omit both the file and the directory.

AndyTitu (dismissed):
Left one brief suggestion which would have most likely been enforced by GitHub itself before unblocking the merge anyway.

cmd/contrib/main.go
```diff
@@ -414,6 +426,66 @@ func Try{{ .PlatformNameUpperCamelCase }}ConfigFile() sdk.Importer {
...
+						fieldname.{{ .FieldName }}: "{{ .TestCredentialExample }}",
+					},
+				},
+			},
+		},
+		// TODO: If you implemented a config file importer, add a test file example in {{ .Name }}/test-fixtures
+		// and fill the necessary details in the test template below.
+		"config file": {
+			Files: map[string]string{
+				// "~/path/to/config/file.yml": plugintest.LoadFixture(t, "your_local_fixture.yaml"),
```
florisvdg, line 474: In the commented `TryFile` section in the credential template, we're using a `Config` YAML struct as an example, so we could follow that pattern here as well:
```suggestion
				// "~/path/to/config/file.yml": plugintest.LoadFixture(t, "config.yml"),
```

volodymyrZotov (dismissed):
Looks good! 👍

volodymyrZotov (approved):

AndyTitu (approved):

PR #102, opened 2022-12-12
Revert "Revert using any instead of interface{}"

hculea:
Since we now use an up to date version of `go` consistently across all the closed-source code, using `interface{}` instead of `any` should not be a problem anymore.


mrjones2014 (approved):

AndyTitu:
lgtm since tests have passed

Issue #90, opened 2022-12-09
panic: unexpected EOF


shyim:
### op CLI version

2.9.1

### Goal or desired behavior

Sometimes there is a weird RPC client error see below

### Current behavior

```
panic: unexpected EOF

goroutine 1 [running]:
go.1password.io/op/op-cli/command/plugin/local.(*RPCClient).provisionCredential(0xe5?, {{0xc000386da8, 0x3}, {0xc000386db0, 0xa}, 0x0}, {{0xc00003a035, 0xb}, {0xc0001a63f0, 0x25}, ...}, ...)
	/builds/dev/b5/op/op-cli/command/plugin/local/rpc.go:57 +0x185
go.1password.io/op/op-cli/command/plugin/local.rpcProvisioner.Provision(...)
	/builds/dev/b5/op/op-cli/command/plugin/local/rpc.go:163
go.1password.io/op/op-cli/command/plugin.provision(0x3bd?, {{0xc000384c40, 0x1, 0x1}, {0xc00038f2c0, 0x1f}, {0xc000383860, 0x1, 0x1}, 0xc00039c510, ...}, ...)
	/builds/dev/b5/op/op-cli/command/plugin/provision_executable.go:164 +0xa76
go.1password.io/op/op-cli/command/plugin.Run(0x0?, {{0xc000384c40, 0x1, 0x1}, {0xc00038f2c0, 0x1f}, {0xc000383860, 0x1, 0x1}, 0xc00039c510, ...}, ...)
	/builds/dev/b5/op/op-cli/command/plugin/provision_executable.go:63 +0x338
go.1password.io/op/op-cli/command.(*pluginRunCommand).Run(0xc00042e870)
	/builds/dev/b5/op/op-cli/command/plugin_run.go:180 +0xccc
go.1password.io/op/op-cli/command.Bind.func3(0xc00048b340?, {0xc0004017c0?, 0x3?, 0x4?})
	/builds/dev/b5/op/op-cli/command/command.go:71 +0x4e
github.com/spf13/cobra.(*Command).execute(0xc00048b340, {0xc000401780, 0x4, 0x4})
	/builds/dev/b5/op/vendor/github.com/spf13/cobra/command.go:844 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0x15a8860)
	/builds/dev/b5/op/vendor/github.com/spf13/cobra/command.go:952 +0x39d
go.1password.io/op/op-cli/command.Execute()
[...]

PR #103, opened 2022-12-12
Fix local builds that use cache or additional command flags

AndyTitu:
Related issue: https://github.com/1Password/shell-plugins/issues/90

### Why

The `Cache` and `CommandLine` field were left nil in the provision output response. This caused panics for both plugins using `Cache` and plugins using additional command flags.

Actually, the desired response of the RPC call should be the `ProvisionOutput` exactly as it was passed down from OP (as @Marton6 noted), but for a certain reason the fields of the pointer argument (`output`) of the call were nullified. (were brought to their zero values).

### What

this MR:
- adds a `ProvisionOutput` field to the `ProvisionCredentialRequest`., and uses it as response of the RPC call
- adds functionality for catching a panic from a locally built plugin and outputting it before the OP panic is reached, so that the OP panic doesn't mask the true cause of the problem.

### DIFF
--- a/sdk/rpc/server/server.go
+++ b/sdk/rpc/server/server.go
@@ -3,10 +3,11 @@ package server
 import (
 	"context"
 	"fmt"
+	"runtime/debug"
+
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/rpc/proto"
 	"github.com/1Password/shell-plugins/sdk/schema"
-	"runtime/debug"
 )
 
 type errFunctionFieldNotSet struct {
--- a/sdk/rpc/server/server.go
+++ b/sdk/rpc/server/server.go
@@ -3,10 +3,10 @@ package server
 import (
 	"context"
 	"fmt"
-
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/rpc/proto"
 	"github.com/1Password/shell-plugins/sdk/schema"
+	"runtime/debug"
 )
 
 type errFunctionFieldNotSet struct {
@@ -144,17 +144,18 @@ func (t *RPCServer) CredentialProvisionerDescription(req proto.ProvisionerID, re
 // interface. The call is forwarded to the Provision() function of the Provisioner of the credential identified by
 // req.CredentialID.
 func (t *RPCServer) CredentialProvisionerProvision(req proto.ProvisionCredentialRequest, resp *sdk.ProvisionOutput) error {
+	defer func() {
+		if err := recover(); err != nil {
+			resp.AddError(fmt.Errorf("your locally built plugin failed with the following panic: %s; and with stack trace: %s", err, string(debug.Stack())))
+		}
+	}()
 	provisioner, err := t.getProvisioner(req.ProvisionerID)
 	if err != nil {
 		return err
 	}
-	*resp = sdk.ProvisionOutput{
-		Environment: make(map[string]string),
-		CommandLine: nil,
-		Files:       make(map[string]sdk.OutputFile),
-		Diagnostics: sdk.Diagnostics{},
-	}
+	*resp = req.ProvisionOutput
 	provisioner.Provision(context.Background(), req.ProvisionInput, resp)
+
 	return nil
 }
 
--- a/sdk/rpc/server/server.go
+++ b/sdk/rpc/server/server.go
@@ -151,7 +151,6 @@ func (t *RPCServer) CredentialProvisionerProvision(req proto.ProvisionCredential
 	}
 	*resp = req.ProvisionOutput
 	provisioner.Provision(context.Background(), req.ProvisionInput, resp)
-
 	return nil
 }
 
--- a/sdk/rpc/server/server.go
+++ b/sdk/rpc/server/server.go
@@ -3,7 +3,6 @@ package server
 import (
 	"context"
 	"fmt"
-	"runtime/debug"
 
 	"github.com/1Password/shell-plugins/sdk"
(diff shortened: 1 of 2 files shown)


sdk/rpc/server/server.go
```diff
@@ -136,17 +137,18 @@ func (t *RPCServer) CredentialProvisionerDescription(req proto.ProvisionerID, re
 // interface. The call is forwarded to the Provision() function of the Provisioner of the credential identified by
 // req.CredentialID.
 func (t *RPCServer) CredentialProvisionerProvision(req proto.ProvisionCredentialRequest, resp *sdk.ProvisionOutput) error {
+	defer func() {
+		if err := recover(); err != nil {
+			resp.AddError(fmt.Errorf("your locally built plugin failed with the following panic: %s; and with stack trace: %s", err, string(debug.Stack())))
```
Marton6, line 142:
```suggestion
			resp.AddError(fmt.Errorf("your locally-built plugin panicked with the following error: %s; and with stack trace: %s", err, string(debug.Stack())))
```

Marton6 (changes requested):
I have a small suggestion about the messaging around panics. Otherwise this MR looks good to me.

sdk/rpc/server/server.go
jpcoenen, line 142: I really like that we're improving what happens on a panic. But could we extract this into a separate PR? I'd like to have a discussion about the consistency between calls and whether we want to use `resp.AddError()` for this. I think there is something to say for having a special type in the `proto` package to encode panics.

sdk/rpc/server/server.go
```diff
@@ -136,17 +137,18 @@ func (t *RPCServer) CredentialProvisionerDescription(req proto.ProvisionerID, re
...
 	if err != nil {
 		return err
 	}
-	*resp = sdk.ProvisionOutput{
-		Environment: make(map[string]string),
-		CommandLine: nil,
-		Files:       make(map[string]sdk.OutputFile),
-		Diagnostics: sdk.Diagnostics{},
-	}
+	*resp = req.ProvisionOutput
```
jpcoenen, line 149: This is a breaking change as it sets a strict requirement on the `req.ProvisionOutput` field. 

For consideration: if we move the initialization to a shared function instead, we can achieve the same without having to break the API. Alternatively, we have to bump `proto.Version`.

jpcoenen: If we are changing the API, let's make sure to change it for all requests.

hculea:
Let's also tackle `DeprovisionOutput` within this MR, shall we?

sdk/rpc/server/server.go
hculea: Question here (just by looking at this): even if we move the initialisation to a shared function, we would still need to set a strict requirement for `req.CommandLine` field ([see Flo's adjustment](https://github.com/1Password/shell-plugins/commit/6bc2bd73ca63c50288c93c31ea9dfb2dda0ccd5e)) in order to pass the command line arguments, is this correct?

If this is the case, I think this prompts for bumping the `proto.Version` either way.

sdk/rpc/server/server.go
AndyTitu: Sure thing, I'll extract this in another PR.

sdk/rpc/server/server.go
jpcoenen: That's fair. I am not 100% happy with the fact that `req.CommandLine` (an output) is populated at the start of `Provision()`, but we can tackle that later and separately. 

The most important thing for now is that we make sure to merge this into the `protocol-v3` branch.

jpcoenen:
> Let's also tackle `DeprovisionOutput` within this MR, shall we?

Yes, same for `Import()`.

sdk/rpc/server/server.go
AndyTitu: This is indeed a protocol breaking change. I agree with Horia that there is no way around incrementing `port.Version`. The most elegant solution here (which would also avoid breaking protocol) would have been to use the `resp` parameter of the function directly as parameter for the `Provision` call here, but that pointer ends up pointing to a new struct when the rpc call is made. That is why I had to pass the output as a struct through the request structure.

AndyTitu:
> Let's also tackle `DeprovisionOutput` within this MR, shall we?

> Yes, same for ImportOutput.

Done!

sdk/rpc/server/server.go
AndyTitu: Thanks for the suggestion! I will apply this in the separate MR made especially for this panic catching.

jpcoenen:
Changed the base branch to `protocol-v3`.

jpcoenen:
I'm also going to rebase this branch onto `protocol-v3`. The current diff is a mess.

jpcoenen (approved):

sdk/rpc/server/server.go
```diff
@@ -148,13 +149,9 @@ func (t *RPCServer) CredentialProvisionerProvision(req proto.ProvisionCredential
...
 	provisioner.Provision(context.Background(), req.ProvisionInput, resp)
+
```
hculea, line 154: tinny tiny nit: Let's remove this newline for consistency 😄

hculea (approved):

sdk/rpc/server/server.go
hculea: If this is going to reset all the approvals, feel free to ignore this.

jpcoenen:
```suggestion
```

AndyTitu:
For reference, Horia and I functionally tested this for both `mysql` and `aws` plugins.

PR #105, opened 2022-12-12
Add helper to sanitize importer name hint

florisvdg:
Turns the common pattern of omitting "default" as the name hint into a helper function in the importer package.


SimonBarendse (approved):
Nice! Love that we're generalizing this. Makes it easier to add for every plugin that could use this.

PR #106, opened 2022-12-12
Add ReadMe plugin

### LINKED ISSUE #108: New plugin: ReadMe CLI (a.k.a. `rdme`)
### Platform or tool

ReadMe

### Relevant links

Homepage:
- https://readme.com/

Various links to `rdme` docs and package registries:
- https://docs.readme.com/main/docs/rdme
- https://github.com/readmeio/rdme
- https://www.npmjs.com/package/rdme
- https://github.com/marketplace/actions/rdme-sync-to-readme

### Goal or desired behavior

Run authenticated commands with a ReadMe API key.

### Contribution

### Additional information

_No response_

kanadgupta:
Hi folks, [big 1Password fans](https://blog.readme.com/two-factor-auth/) over here at ReadMe! Looking forward to being a part of this exciting new ecosystem 💙

~I still have a few small outstanding TODOs in this PR (that I'll address before officially marking this as ready for review), but the main outstanding issue is that I'm running into is that the env variable `RDME_API_KEY` isn't properly being provisioned in my testing 😕 see #107 for more information!~

**EDIT**: fixed, thanks for your help @SimonBarendse @florisvdg!

Closes #108

### DIFF
--- a/plugins/readme/api_key.go
+++ b/plugins/readme/api_key.go
@@ -13,8 +13,8 @@ import (
 
 func APIKey() schema.CredentialType {
 	return schema.CredentialType{
-		Name:          credname.APIKey,
-		DocsURL:       sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
+		Name:    credname.APIKey,
+		DocsURL: sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.Username,
--- a/plugins/readme/api_key.go
+++ b/plugins/readme/api_key.go
@@ -15,7 +15,6 @@ func APIKey() schema.CredentialType {
 	return schema.CredentialType{
 		Name:          credname.APIKey,
 		DocsURL:       sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
-		ManagementURL: sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.Username,
--- a/plugins/readme/api_key.go
+++ b/plugins/readme/api_key.go
@@ -28,7 +28,6 @@ func APIKey() schema.CredentialType {
 				MarkdownDescription: "API Key used to authenticate to ReadMe.",
 				Secret:              true,
 				Composition: &schema.ValueComposition{
-					Length: 75,
 					Prefix: "rdme_",
 					Charset: schema.Charset{
 						Lowercase: true,
--- a/plugins/readme/api_key.go
+++ b/plugins/readme/api_key.go
@@ -15,7 +15,7 @@ func APIKey() schema.CredentialType {
 	return schema.CredentialType{
 		Name:          credname.APIKey,
 		DocsURL:       sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
-		ManagementURL: sdk.URL("https://console.readme.com/user/security/tokens"), // TODO: Replace with actual URL
+		ManagementURL: sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.Username,
--- a/plugins/readme/api_key.go
+++ b/plugins/readme/api_key.go
@@ -35,6 +35,12 @@ func APIKey() schema.CredentialType {
 					},
 				},
 			},
+			{
+				Name:                fieldname.Website,
+				MarkdownDescription: "The ReadMe project URL.",
+				Secret:              false,
+				Optional:            true,
+			},
 		},
 		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
 		Importer: importer.TryAll(
@@ -43,11 +49,10 @@ func APIKey() schema.CredentialType {
 		)}
 }
 
-// TODO: figure out why RDME_API_KEY isn't being set as env var
 var defaultEnvVarMapping = map[string]sdk.FieldName{
 	"RDME_API_KEY": fieldname.APIKey,
 	"RDME_EMAIL":   fieldname.Username,
-	// "RDME_PROJECT": fieldname.Website,
+	"RDME_PROJECT": fieldname.Website,
 }
 
 func TryReadMeConfigFile() sdk.Importer {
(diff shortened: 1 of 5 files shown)


erunion (dismissed):
:nice-rgb:

plugins/readme/api_key.go
```diff
@@ -0,0 +1,87 @@
...
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
+		ManagementURL: sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Username,
+				MarkdownDescription: "Currently logged-in ReadMe user.",
```
SimonBarendse, line 22: We show this text _before_ the user is logged in. I think we should slightly reword to incorporate this. Similar to [how we do for MySQL username field](https://github.com/1Password/shell-plugins/blob/main/plugins/mysql/database_credentials.go#L32)

```suggestion
				MarkdownDescription: "ReadMe user to log-in as.",
```

plugins/readme/api_key.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
+		ManagementURL: sdk.URL("https://docs.readme.com/main/reference/intro/authentication"),
```
SimonBarendse, line 18: The `ManagementURL` field is meant to point to the dashboard where you can create a new key, rather than the docs that explain it (that's what the `DocsURL` entry you filled above is for). I believe https://dash.readme.com/ is where you create the API key for ReadMe, is that correct?

SimonBarendse: We'll use the management URL in future improvements. For example, we could offer users that have the 1Password extension installed, to save newly created API keys directly from within the browser when they visit the management page in the dashboard.

SimonBarendse (commented):
Love it! Thanks for contributing @kanadgupta! I've just left a few last nitpicks for you to consider, overall this looks great! I love the detail in your MR, including using the subdomain as naming suggestion for the 1Password item, the inclusion of the configuration file importer and adding these tests! :raised_hands:

Welcome to the ecosystem ReadMe! :heart_eyes:

plugins/readme/api_key.go
kanadgupta: So this was an intentional decision because our API key management dashboard has a few path parameters that are specific to the user's ReadMe project/version setup 😞 for example:

```
https://dash.readme.com/project/some-project/v1.0/api-key
https://dash.readme.com/project/another-project/v2.1/api-key
```

The reason I linked to the docs again is because this page contains [instructions](https://docs.readme.com/main/reference/intro/authentication#generating-new-api-keys-) on how to access the dashboard. Should I link to that section specifically?

I do love the management URL idea though, have y'all considered accepting wildcards for the management URL? Being able to pass in something like this would be super helpful for us:

```
https://dash.readme.com/project/**/v**/api-key
```

I do also have an internal ticket for setting up redirects so a user could pass in something like `https://dash.readme.com/my/api-key` and it would redirect them to the correct place in the dashboard, but don't have an ETA on that unfortunately.

plugins/readme/api_key.go
kanadgupta: Thanks for the clarification, made that change in https://github.com/1Password/shell-plugins/pull/106/commits/04ac3272820eb806de4b1beb50c946dae77b71ae (with a slight grammar tweak to your suggestion)

kanadgupta:
(oops, didn't mean to remove the review request for @DCKcode, sorry!)

plugins/readme/api_key.go
SimonBarendse: Okay, thank you for the clarification! :raised_hands:

Right now the only place we're using the management URL is in printing this message:

```Golang
helpText := fmt.Sprintf("To authenticate %s with 1Password, configure which item on 1Password to use as the %s.", pluginExecutable.Name, credentialType.Name)
helpText += fmt.Sprintf("\n\n→ If you already have your %s saved in 1Password, choose Search in 1Password.", credentialType.Name)
helpText += fmt.Sprintf("\n\n→ If you don't have your %s saved in 1Password yet, choose Import into 1Password.", credentialType.Name)

if credentialType.ManagementURL != nil {
	helpText += fmt.Sprintf("\n  You can create one here if you don't have one available:\n  %s", credentialType.ManagementURL)
}

if credentialType.DocsURL != nil {
		helpText += fmt.Sprintf("\n\n→ For more info about the %s, see the documentation:\n  %s", credentialType.Name, credentialType.DocsURL)
}
```

I propose we proceed as follows:
- I'll file an internal ticket for us to consider accepting wildcards for the management URL for when we'll use it in the browser extension.
- For now, let's omit the `ManagementURL` on the ReadMe plugin. It's an optional field. When either we have added wildcard support or when your redirect solution has shipped (whichever ships first), we can add the management URL to the ReadMe plugin. The reason I prefer to proceed with not setting the management URL for the ReadMe plugin yet is that this allows us to tweak the user experience slightly for plugins that don't have a management URL for any future uses we might have for the management URL. The fallback to the docs URL might look slightly different than using the dashboard URL.
- I'll file an internal ticket for us to slightly change the wording on that line about the DocsURL to also mention `create` so it's extra clear that that doc is the place to go for info about how to create the API Key for ReadMe (because the line above about creating a new credential is omitted when the management URL isn't set).

```suggestion
```

DCKcode:
This looks great - thanks for your contribution!

DCKcode (dismissed):

plugins/readme/api_key.go
kanadgupta: Ah okay, didn't realize this field was optional. Happy with this approach! Removed in https://github.com/1Password/shell-plugins/commit/57f5f20ce0dcbdd74770c2839afb6ae7ae7d2425

SimonBarendse (approved):

DCKcode (dismissed):

plugins/readme/api_key.go
```diff
@@ -0,0 +1,86 @@
...
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "API Key used to authenticate to ReadMe.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Prefix: "rdme_",
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
```
volodymyrZotov, line 35: Just wonder if the `APIKey` has a fixed length? It is worth adding it to the composition if so.

volodymyrZotov (approved):

kanadgupta:
(sorry folks, just fixing up a linting error!)

volodymyrZotov (approved):

DCKcode (approved):

kanadgupta:
So sorry y'all, I was playing around locally and realized that it doesn't make sense for us to include `whoami` in our list of commands. This is **_officially_** ready to be merged! 😅

DCKcode (approved):

volodymyrZotov (approved):

SimonBarendse (approved):

PR #109, opened 2022-12-13
DigitalOcean config file importer for MacOS

volodymyrZotov:
This PR adds Digitalocean config file importer for MacOS.

The config file is located in `${HOME}Library/Application Support/doctl/config.yaml` directory on OS X.

On Linux, doctl saves its configuration as `${XDG_CONFIG_HOME}/doctl/config.yaml` if the `${XDG_CONFIG_HOME}` environmental variable is set, or `~/.config/doctl/config.yaml` if it is not.  This will be added after this issue #153 is resolved.

The documentation is [here](https://github.com/digitalocean/doctl#configuring-default-values).


plugins/digitalocean/test-fixtures/config.yaml
```diff
@@ -0,0 +1,1188 @@
...
+  get:
+    format: ""
+    no-header: false
+  list:
+    format: ""
+    no-header: false
+  update:
+    default: false
+    description: ""
+    name: ""
```
florisvdg, line 1188: While I like it that we're using a realistic config file here, having all the zero-values in there does hurt readability, so I'm slightly leaning in favor of trimming it down to remove the zero-values:
```suggestion
access-token: dop_v1_tr33mpd5m8q9t3ncisqbceydi8dd2n60pl1yiycg97z25fkqffp8j6ycjexample
config: ${HOME}/Library/Application Support/doctl/config.yaml
context: default
output: text
```

plugins/digitalocean/personal_access_token.go
```diff
@@ -32,6 +34,33 @@ func PersonalAccessToken() schema.CredentialType {
...
+		out.AddCandidate(sdk.ImportCandidate{
+			Fields: map[sdk.FieldName]string{
+				fieldname.Token: config.Token,
+			},
+		})
+	})
+}
+
+type Config struct {
+	Token string `yaml:"access-token"`
```
florisvdg, line 65: Nit: would be nice to have struct/YAML naming parity:
```suggestion
		if config.AccessToken == "" {
			return
		}

		out.AddCandidate(sdk.ImportCandidate{
			Fields: map[sdk.FieldName]string{
				fieldname.Token: config.AccessToken,
			},
		})
	})
}

type Config struct {
	AccessToken string `yaml:"access-token"`
```

volodymyrZotov: Ok, makes sense.

florisvdg (approved):

plugins/digitalocean/test-fixtures/config.yaml
```diff
@@ -0,0 +1,4 @@
+access-token: dop_v1_tr33mpd5m8q9t3ncisqbceydi8dd2n60pl1yiycg97z25fkqffp8j6ycjexample
+config: ${HOME}/Library/Application Support/doctl/config.yaml
+context: default
```
florisvdg, line 3: Does the DigitalOcean config file support multiple contexts / profiles?

volodymyrZotov: The config file itself doesn't have a list of available contexts.
It can be found by using `doctl atuh list` command.

The `context` property in the config file represents the currently selected context.

florisvdg: So should this importer then import those other `auth list` profiles too? (Can be in a different PR)

volodymyrZotov: Yeah, I think that is something we can achieve.
When we switch the context it populates the config file with the proper token value. So, we would need to read the content of the config for each candidate.

In general, the process would look like that.
- execute `doctl auth list` and retrieve all profiles.
- go through each of them:
           - execute `doctl auth swith --context <name>` to set current context.
           - read the config file and extract the token and profile name to create a candidate.

florisvdg: But where does `doctl` get the token from when the user runs `doctl auth switch`?

plugins/digitalocean/personal_access_token.go
```diff
@@ -32,6 +34,33 @@ func PersonalAccessToken() schema.CredentialType {
...
-		Importer: importer.TryAllEnvVars(fieldname.Token, "DIGITALOCEAN_ACCESS_TOKEN"),
+		Importer: importer.TryAll(
+			importer.TryAllEnvVars(fieldname.Token, "DIGITALOCEAN_ACCESS_TOKEN"),
+			TryDigitalOceanConfigFile(),
+		),
 	}
 }
+
+func TryDigitalOceanConfigFile() sdk.Importer {
+	return importer.TryFile("~/Library/Application Support/doctl/config.yaml", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
```
hculea, line 45: Is a loading configuration from file only available for MacOS? Otherwise, we may want to add support for Linux as well (only these two OS's are supported on the CLI side)

hculea: I'd say let's wait for the merging of this: https://github.com/1Password/shell-plugins/pull/127

florisvdg: I'd also be okay with going ahead and merging this PR. Nothing will break in the mean time on Linux. We can add Linux support in a follow-up PR.

plugins/digitalocean/test-fixtures/config.yaml
volodymyrZotov: That's a good question. I suppose it calls some API under the hood to retrieve it.
At least `config.yaml` has a single `context` and `token` field that's replaced with different values when you do `switch`

florisvdg: In that case it probably requires a user login to complete, which would be out of scope for (how we currently see) importers.

plugins/digitalocean/personal_access_token.go
florisvdg: I'd opt for the following signature:
```go
importer.MacOnly(
	TryDigitalOceanConfigFile("~/Library/Application Support/doctl/config.yaml"),
),
```

plugins/digitalocean/personal_access_token_test.go
```diff
@@ -22,6 +22,32 @@ func TestPersonalAccessTokenImporter(t *testing.T) {
...
+					Fields: map[sdk.FieldName]string{
+						fieldname.Token: "dop_v1_tr33mpd5m8q9t3ncisqbceydi8dd2n60pl1yiycg97z25fkqffp8j6ycjexample",
+					},
+				},
+			},
+		},
+		"config file linux": {
+			OS: "linux",
+			Files: map[string]string{
+				digitalOceanConfigFileOnLinux(): plugintest.LoadFixture(t, "config.yaml"),
```
SimonBarendse, line 41: Because we're executing `digitalOceanConfigFileOnLinux` here in the testcase, this case tests a different scenario based on whether or not `$XDG_HOME` is set on the machine that's running these tests. Can we make the test statically defined instead?

```suggestion
				"~/.config/doctl/config.yaml": plugintest.LoadFixture(t, "config.yaml"),
```

hculea:
Looks like the pipeline image that runs the tests actually configures XDG_CONFIG_HOME. I'll mark this as draft for now, I'll take some tomorrow morning to wrap it up. 😄

plugins/digitalocean/personal_access_token.go
```diff
@@ -32,6 +36,42 @@ func PersonalAccessToken() schema.CredentialType {
...
+	})
+}
+
+func digitalOceanConfigFileOnLinux() string {
+	basePath := "~/.config"
+	if xdg := os.Getenv("XDG_CONFIG_HOME"); xdg != "" {
+		basePath = xdg
+	}
+	return path.Join(basePath, "doctl/config.yaml")
+}
```
florisvdg, line 73: This is not a DigitalOcean-only thing. This goes for (I think) every plugin we have that uses `~/.config`, which are quite a lot.

A while ago I [experimented](https://github.com/1Password/shell-plugins/commit/d44a749ba7d89e729a7a11366af6c3ac67cfd943) with a way to standardize this pattern that could be interesting to further explore.

hculea: You're right that we should seek to standardise this. I opened an issue for this, I propose we tackle it outside this MR.

https://github.com/1Password/shell-plugins/issues/153

hculea:
Based on Flo's thread above, we decided that it would be best to standardise importers that use `XDG_CONFIG_HOME`, since it is quite a common pattern. I'll therefore only tackle the MacOS importer in this PR (which has a static file path), as the plan was initially, and we can pick this up separately: https://github.com/1Password/shell-plugins/issues/153

accraw (dismissed):
Makes sense to me!

plugins/digitalocean/personal_access_token.go
```diff
@@ -32,6 +34,33 @@ func PersonalAccessToken() schema.CredentialType {
 		DefaultProvisioner: provision.EnvVars(map[string]sdk.FieldName{
 			"DIGITALOCEAN_ACCESS_TOKEN": fieldname.Token,
 		}),
-		Importer: importer.TryAllEnvVars(fieldname.Token, "DIGITALOCEAN_ACCESS_TOKEN"),
+		Importer: importer.TryAll(
+			importer.TryAllEnvVars(fieldname.Token, "DIGITALOCEAN_ACCESS_TOKEN"),
+			importer.MacOnly(TryDigitalOceanConfigFile("~/Library/Application Support/doctl/config.yaml")),
```
florisvdg, line 39: Should we use the following indentation pattern for these for readability purposes?
```suggestion
			importer.MacOnly(
				TryDigitalOceanConfigFile("~/Library/Application Support/doctl/config.yaml")
			),
```

florisvdg (approved):

AndyTitu (approved):
lgtm now

PR #110, opened 2022-12-13
Created plugin for homebrew.

markdorison:
## Description
Created a plugin for homebrew. Token is needed for connection to GitHub.

### DIFF
--- a/plugins/homebrew/personal_access_token.go
+++ b/plugins/homebrew/personal_access_token.go
@@ -1,0 +1,41 @@
+package homebrew
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func PersonalAccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.PersonalAccessToken,
+		DocsURL:       sdk.URL("https://docs.brew.sh/Manpage#environment"),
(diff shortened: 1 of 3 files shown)


florisvdg:
Thanks for another contribution! What's the use case for this plugin? More favorable rate limits for the `brew search` command?

markdorison:
Yes RE: rate limits! I have also run into it when [creating Homebrew pull requests](https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request) using `brew bump-cask-pr` and `brew bump-formula-pr`. There may be other usages that I am not aware of.

florisvdg:
Good to know. If it only applies to a few specific commands, I'd be in favor of only require authentication for those specific commands (even if we're not 100% sure that covers all commands). See [this PR](https://github.com/1Password/shell-plugins/pull/106) for an example of how to use the `needsauth` package for that.

markdorison:
@florisvdg Thanks for pointing me to that example! Updated!

plugins/homebrew/personal_access_token.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func PersonalAccessToken() schema.CredentialType {
```
SimonBarendse, line 12: I was wondering if we should re-use the definition of the GitHub Personal Access Token from the GitHub plugin, but considering GitHub CLI supports enterprise hosting while Brew only uses https://github.com I think creating a separate definition here like you did is the best approach. :+1:

SimonBarendse (approved):
Thank you Mark for another great contribution! :raised_hands:

DCKcode (approved):

Issue #91, opened 2022-12-09
Linode importer


florisvdg:
### Platform or tool

Linode

### Desired or expected behavior

Would be nice for the Linode plugin to support importing the token from the [`~/.config/linode-cli`](https://www.linode.com/docs/products/tools/cli/guides/install/#interactive-configuration).

### Current behavior

We're only looking at environment variables now.

SimonBarendse:
@alexclst do you want to give the importer for Linode a shot as well? I know this is your first experience with Golang, so come join us in the [Developer Slack workspace](https://join.slack.com/t/1password-devs/shared_invite/zt-1halo11ps-6o9pEv96xZ3LtX_VE0fJQA)! It's an amazing bunch of 1Password enthusiasts and we can help in case you have any questions.

alexclst:
@SimonBarendse I would say that I could give it a shot, except that now that I have the plugin configured I'd already deleted (actually long ago, as I'd been using secret references in `op` for a while for Linode) the config file, and it appears that the `linode-cli` will not let me remake it, at least when run through the 1Password shell plugin as it now is being done.

SimonBarendse:
To keep 1Password Shell Plugins installed on your machine, while also running Linode one more time without 1Password Shell Plugins, you could run it [through Docker](https://hub.docker.com/r/linode/cli).

You can use the steps for using an existing config file (`-v $HOME/.config/linode-cli:/home/cli/.config/linode-cli`) so that the configuration file you create in the docker container also ends up in your home dir.

alexclst:
@SimonBarendse I realized a simpler alternative was to just call `linode-cli` with its full path, to bypass the shell plugin. So now I have a config file, and can sometime look into the config file importer.

SimonBarendse:
If you run into any further issues, don't hesitate to reach out! We're here to help!

volodymyrZotov:
@alexclst Hey. I look at the file importer you added.
I'd suggest adding a test case that covers importing from a file.
Here is an example from sentry plugin. (make sure you created a folder `plugins/linode/test-fixtures` and put there the config file)
```
"config file": {
			Files: map[string]string{
				"~/.sentryclirc": plugintest.LoadFixture(t, ".sentryclirc"),
			},
			ExpectedCandidates: []sdk.ImportCandidate{
				{
					Fields: map[sdk.FieldName]string{
[...]

PR #113, opened 2022-12-14
Linode Config File Importer

alexclst:
This pull request adds an importer for `~/.config/linode-cli` to the Linode shell plugin. This would resolve issue #91.

### DIFF
--- a/plugins/linode/personal_access_token.go
+++ b/plugins/linode/personal_access_token.go
@@ -1,6 +1,8 @@
 package linode
 
 import (
+	"context"
+	"os"
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
 	"github.com/1Password/shell-plugins/sdk/provision"
@@ -31,9 +33,39 @@ func PersonalAccessToken() schema.CredentialType {
 		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
 		Importer: importer.TryAll(
 			importer.TryEnvVarPair(defaultEnvVarMapping),
+			TryConfigFile(),
 		)}
 }
 
 var defaultEnvVarMapping = map[string]sdk.FieldName{
 	"LINODE_CLI_TOKEN": fieldname.Token,
+}
+
+// TryConfigFile looks for the token in the ~/.config/linode-cli file.
+func TryConfigFile() sdk.Importer {
+	return importer.TryFile("~/.config/linode-cli", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		configFile, err := contents.ToINI()
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		if err != nil && !os.IsNotExist(err) {
+			out.AddError(err)
+		}
+
+		for _, section := range configFile.Sections() {
+			fields := make(map[sdk.FieldName]string)
+			if section.HasKey("token") && section.Key("token").Value() != "" {
+				fields[fieldname.Token] = section.Key("token").Value()
+			}
+
+			// add only candidates with required credential fields
+			if fields[fieldname.AccessKeyID] != "" && fields[fieldname.SecretAccessKey] != "" {
+				out.AddCandidate(sdk.ImportCandidate{
+					Fields:   fields,
+				})
+			}
+		}
+	})
 }
--- a/plugins/linode/personal_access_token.go
+++ b/plugins/linode/personal_access_token.go
@@ -2,6 +2,7 @@ package linode
 
 import (
 	"context"
+
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
 	"github.com/1Password/shell-plugins/sdk/provision"
--- a/plugins/linode/personal_access_token.go
+++ b/plugins/linode/personal_access_token.go
@@ -50,10 +50,8 @@ func TryConfigFile() sdk.Importer {
 			return
 		}
 
-		fields := make(map[sdk.FieldName]string)
 		for _, section := range configFile.Sections() {
 			if section.HasKey("token") && section.Key("token").Value() != "" {
-				fields[fieldname.Token] = section.Key("token").Value()
 				out.AddCandidate(sdk.ImportCandidate{
 					NameHint: importer.SanitizeNameHint(section.Name()),
 					Fields: map[sdk.FieldName]string{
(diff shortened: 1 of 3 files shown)


volodymyrZotov:
@alexclst Nice! Thank you for your contribution!
Please run `go fmt` to fix all the lint errors.

SimonBarendse:
@alexclst You can read more about what `gofmt` is and why it exists at https://go.dev/blog/gofmt

Most IDEs offer a way to automatically `gofmt` on save. I'd consider enabling that so that you don't have to worry about formatting for future changes.

plugins/linode/personal_access_token.go
```diff
@@ -31,9 +32,32 @@ func PersonalAccessToken() schema.CredentialType {
...
+		}
+
+		fields := make(map[sdk.FieldName]string)
+		for _, section := range configFile.Sections() {
+			if section.HasKey("token") && section.Key("token").Value() != "" {
+				fields[fieldname.Token] = section.Key("token").Value()
+			}
+		}
+
+		out.AddCandidate(sdk.ImportCandidate{
```
SimonBarendse, line 59: Consider adding a `NameHint` with the username (`acme` in your example). This will append the username to the name suggested to users to use for the 1Password item that will include the imported item. This will be useful for users that have more than one Linode access tokens.

SimonBarendse, line 59: If I understand the format of the configuration file correctly, it can contain multiple credentials. For each credential, we'll want to create a separate 1Password item. Therefore, we should move `out.AddCandidate` inside the loop through the sections, so that we're adding a new entry for each section that contains a token.

SimonBarendse: Consider adding another test fixture for an example with multiple sections (users) to ensure that works as expected.

SimonBarendse (commented):
Thanks for another great contribution @alexclst! :raised_hands:

I have a couple questions, which boil down to whether there's a use case in Linode to have multiple users/personal access tokens? I get this impression from the configuration file, but I'm not a Linode user myself, so I'll leave it to your judgement to consider whether there's a use case here we need to account for.

alexclst:
`go fmt` run.

As for multiple credentials, I personally think this is a lesser-used thing, since one creential can work across an entire account, and multiple services can therefore be using the same credential. At least, in my expeience, I use one account that has access to many different services.

plugins/linode/personal_access_token.go
alexclst: Regarding multiple credentials, I personally think this is a lesser-used thing, since one creential can work across an entire account, and multiple services can therefore be using the same credential. At least, in my expeience, I use one account that has access to many different services. I'm also not too clear on exactly where to pull the name hint from, as the account name is not quite the same kind of field in the config file as the token.

alexclst: If you help me see where to add in support for name hints and multiple credentials I'll also add a test fixture to test that.

volodymyrZotov: The loop will look something like that
```
for _, section := range configFile.Sections() {
	if section.HasKey("token") && section.Key("token").Value() != "" {
		fields[fieldname.Token] = section.Key("token").Value()
                 out.AddCandidate(sdk.ImportCandidate{
		        NameHint: importer.SanitizeNameHint(section.Name()),
		        Fields: map[sdk.FieldName]string{
			        fieldname.Token: section.Key("token").Value(),
		        },
	        })
	}
}
```
Note: you don't need to add candidates outside of the loop anymore.

Each section is considered a separate profile. If it has a `token` property we would create a candidate with the name the same as a section name.

plugins/linode/test-fixtures/linode-cli
```diff
@@ -0,0 +1,6 @@
+[DEFAULT]
+default-user = acme
+
+[acme]
+token = cn5z4umbimz0lxgzvps1bl979n8lpwnu6qmb4x19bddzx6siormnoxg2yexample
+
```
volodymyrZotov, line 6: Add another section that contains `token` below. For example
```
[prod]
token = abc123..........
```

plugins/linode/personal_access_token_test.go
```diff
@@ -0,0 +1,53 @@
...
+		"config file": {
+			Files: map[string]string{
+				"~/.config/linode-cli": plugintest.LoadFixture(t, "linode-cli"),
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{
+					Fields: map[sdk.FieldName]string{
+						fieldname.Token: "cn5z4umbimz0lxgzvps1bl979n8lpwnu6qmb4x19bddzx6siormnoxg2yexample",
+					},
+				},
```
volodymyrZotov, line 34: Add `NameHint: "acme"` for this candidate.

plugins/linode/personal_access_token_test.go
```diff
@@ -0,0 +1,53 @@
...
+			},
+		},
+		"config file": {
+			Files: map[string]string{
+				"~/.config/linode-cli": plugintest.LoadFixture(t, "linode-cli"),
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
```
volodymyrZotov, line 29: Add one more candidate with the `NameHint` and `Fields` from `linode-cli` filed.

For example, if you added the same section that I suggested
```
[prod]
token = abc123..........
```
`NameHint` will be `prod`
And `Fields` will contain `fieldname.Token: abc123..........`

plugins/linode/personal_access_token.go
volodymyrZotov: I added more comments that might help you with the unit tests.
Please let me know if you need help.

alexclst: That new version of the loop makes sense, I've now added it.

plugins/linode/personal_access_token.go
```diff
@@ -31,9 +33,34 @@ func PersonalAccessToken() schema.CredentialType {
...
+		configFile, err := contents.ToINI()
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		fields := make(map[sdk.FieldName]string)
+		for _, section := range configFile.Sections() {
+			if section.HasKey("token") && section.Key("token").Value() != "" {
+				fields[fieldname.Token] = section.Key("token").Value()
```
volodymyrZotov, line 56: Looks like you don't need `fields` slice anymore as you populate candidate fields with a literal map (line 60).

alexclst: Gotten rid of that slicing.

volodymyrZotov (approved):
Looks good! Thank you!

SimonBarendse (approved):
Thanks @alexclst for another great contribution! :raised_hands: 

Before you know it you're regularly writing Golang :smile:

DCKcode (approved):

PR #115, opened 2022-12-15
Survey credential name validation and transform improvments

volodymyrZotov:
This PR improves validation for `Credential Name` in a survey, when user runs `make new-plugin`. 
It also applies `Transform` function on user input. Therefore users are able to provide input in any case (upper, lower, camel etc.). 

`Validate` function checks that provided string is in TitleCase.
The user has to re-enter value/select from suggested if not passed. The error appears in a console in this case.

`Transfrom` function finds the corresponding value among defined credential names and returns it. Returns provided value otherwise.

### DIFF
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -118,33 +118,8 @@ func newPlugin() error {
 					return suggestions
 				},
 			},
-			Validate: func(ans any) error {
-				if str, ok := ans.(string); ok {
-					if len(str) == 0 {
-						return errors.New(`credential name must be titlecased, e.g. "Access Key" or "Personal Access Token"`)
-					}
-
-					words := strings.Split(str, " ")
-					for _, word := range words {
-						if unicode.IsLower(int32(word[0])) {
-							return errors.New(`credential name must be titlecased, e.g. "Access Key" or "Personal Access Token"`)
-						}
-					}
-				}
-
-				return nil
-			},
-			Transform: func(ans any) (newAns any) {
-				if str, ok := ans.(string); ok {
-					for _, name := range credname.ListAll() {
-						if strings.EqualFold(name.String(), str) {
-							return string(name)
-						}
-					}
-				}
-
-				return ans
-			},
+			Validate:  validateCredentialName,
+			Transform: transformCredentialName,
 		},
 		{
 			Name:   "ExampleCredential",
@@ -540,3 +515,32 @@ func init() {
 {{- end }}
 }
 `
+
+func validateCredentialName(ans any) error {
+	if str, ok := ans.(string); ok {
+		if len(str) == 0 {
+			return errors.New(`credential name must be titlecased, e.g. "Access Key" or "Personal Access Token"`)
+		}
+
+		words := strings.Split(str, " ")
+		for _, word := range words {
+			if unicode.IsLower(int32(word[0])) {
+				return errors.New(`credential name must be titlecased, e.g. "Access Key" or "Personal Access Token"`)
+			}
+		}
+	}
+
+	return nil
+}
+
+func transformCredentialName(ans any) (newAns any) {
+	if str, ok := ans.(string); ok {
+		for _, name := range credname.ListAll() {
+			if strings.EqualFold(name.String(), str) {
+				return string(name)
+			}
+		}
+	}
+
+	return ans
+}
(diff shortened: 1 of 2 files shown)


cmd/contrib/main.go
```diff
@@ -118,22 +118,36 @@ func newPlugin() error {
 					return suggestions
 				},
 			},
-			Validate: func(ans any) error {
+			Validate: func(ans interface{}) error {
```
SimonBarendse, line 121: Can you elaborate on why you changed from `any` to `interface{}`?

cmd/contrib/main.go
```diff
@@ -137,6 +137,17 @@ func newPlugin() error {
 
 				return nil
 			},
+			Transform: func(ans interface{}) (newAns interface{}) {
```
SimonBarendse, line 142: This change is perfect! Love that we're automatically changing this for the user (rather than making it part of the validation and return an error message), that way we're not burdening the user with doing more work than needed. :raised_hands:

cmd/contrib/main.go
```diff
@@ -118,17 +118,20 @@ func newPlugin() error {
...
+						if strings.ToLower(name.String()) == strings.ToLower(str) {
+							credNamesContainStr = true
+							break
 						}
 					}
-					if !hasUpper {
-						return errors.New(`credential name must be titlecased, e.g. "Access Key" or "Personal Access Token"`)
+
+					if !credNamesContainStr {
+						return errors.New(`credential name must be one of the suggested, e.g. "Access Key" or "Personal Access Token"`)
```
SimonBarendse, line 134: It's a feature that you currently can also use names that are not in the already defined list. We know this list is not exhaustive and there will be new plugins that require additional fields to be defined.

I believe the current behavior is that when you enter a name that's not yet defined, a TODO comment is added for the user to go and define it. That's great imo.

SimonBarendse: I think we can revert your change to the validation while keeping your change to the transform function to achieve the ideal result. What do you think?

cmd/contrib/main.go
volodymyrZotov: I remember[ there was a PR](https://github.com/1Password/shell-plugins/pull/73) to replace `any` with an `interface{}` in all the places in the code.
Looks like there are some compatibility issues with the closed-source codebase.

cmd/contrib/main.go
volodymyrZotov: Aha, I see.
I can offer another improvement for `Validate` function, as the initial one doesn't cover all the cases anyway. For example, such a string will pass `Access token` (`token` in lowercase).

What do you think about changing it to that
```
Validate: func(ans interface{}) error {
				if str, ok := ans.(string); ok {
					words := strings.Split(str, " ")
					for _, word := range words {
						if unicode.IsLower(int32(word[0])) {
							return errors.New(`credential name must be titlecased, e.g. "Access Key" or "Personal Access Token"`)
						}
					}
				}

				return nil
			},
```

With that implementation, we are sure that provided string is really in TitleCase.

SimonBarendse: Is `Validate` called before or after transformation?

I'd expect `Access token` to be automatically changed to `Access Token` because of the transformation function you wrote. If validation happens after transformation, I agree that it should check for casing as well. :+1:

cmd/contrib/main.go
SimonBarendse: In a [later MR](https://github.com/1Password/shell-plugins/pull/102), we've reverted the revert. `any` is now again what we use.

The reason it was initially reverted is because we still had a runner using a version of Go that does not yet support `any`. Since updating that runner to Go 1.19, we have reverted the revert again and switched back to using `any`.

cmd/contrib/main.go
volodymyrZotov: Yeah, `Transform` calls after `Validate` and only if validation passes.
I'll update `Validate` with the example I shared above.

SimonBarendse: Since there are some nuances here, I'd love if you could add some unit tests that cover the various combinations (`Access token` that's accepted and transformed to `Access Token`, `Foo bar` that's rejected, `Access Token` that's accepted etc.)

cmd/contrib/main.go
volodymyrZotov: Ok. Then I can use `any` there 😄

SimonBarendse (approved):

hculea (approved):

PR #116, opened 2022-12-16
RPC Protocol update to v3

AndyTitu:
This merges all protocol-v3 changes into main.


jpcoenen (approved):

hculea (approved):
Approving, since I reviewed individually all changes that went into `protocol-v3`. Looks good to me!

PR #121, opened 2022-12-19
RPC functions catch panics

AndyTitu:
Panics happening in the shell plugins code called over rpc are no longer ignored. This makes for more useful debug messages in errors.

### DIFF
--- a/sdk/rpc/server/server.go
+++ b/sdk/rpc/server/server.go
@@ -3,10 +3,11 @@ package server
 import (
 	"context"
 	"fmt"
+	"runtime/debug"
+
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/rpc/proto"
 	"github.com/1Password/shell-plugins/sdk/schema"
-	"runtime/debug"
 )
 
 type errFunctionFieldNotSet struct {
--- a/sdk/rpc/server/server.go
+++ b/sdk/rpc/server/server.go
@@ -118,6 +118,20 @@ func (t *RPCServer) ExecutableNeedsAuth(req proto.ExecutableNeedsAuthRequest, re
 // CredentialImport is a remote version of the Import() function in schema.CredentialType.
 // The call is forwarded to the Import() function of the credential identified by req.CredentialID.
 func (t *RPCServer) CredentialImport(req proto.ImportCredentialRequest, resp *sdk.ImportOutput) error {
+	defer func() {
+		if err := recover(); err != nil {
+			caughtPanic := fmt.Errorf("your locally built plugin's importing failed with the following panic: %s; and with stack trace: %s", err, string(debug.Stack()))
+			diagnostics := sdk.Diagnostics{Errors: []sdk.Error{{caughtPanic.Error()}}}
+			resp.Attempts = []*sdk.ImportAttempt{
+				{
+					Candidates:  nil,
+					Source:      sdk.ImportSource{},
+					Diagnostics: diagnostics,
+				},
+			}
+		}
+	}()
+
 	importer, ok := t.importers[req.CredentialID]
 	if !ok || importer == nil {
 		return &errFunctionFieldNotSet{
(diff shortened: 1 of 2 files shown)


sdk/rpc/server/server.go
```diff
@@ -145,6 +146,11 @@ func (t *RPCServer) CredentialProvisionerDescription(req proto.ProvisionerID, re
 // interface. The call is forwarded to the Provision() function of the Provisioner of the credential identified by
 // req.CredentialID.
 func (t *RPCServer) CredentialProvisionerProvision(req proto.ProvisionCredentialRequest, resp *sdk.ProvisionOutput) error {
+	defer func() {
+		if err := recover(); err != nil {
+			resp.AddError(fmt.Errorf("your locally built plugin failed with the following panic: %s; and with stack trace: %s", err, string(debug.Stack())))
+		}
+	}()
```
AndyTitu, line 153: before attempting to put "panic catchers" in the other functions as well, I figured a discussion about which technique we'd like to use here should be in place @jpcoenen

jpcoenen: I think the panic is part of the RPC layer and should therefore not "leak" into the rest of the plugin system. Also, an alternative method could provide some more flexibility for the CLI to deal with these panics.

What if instead of adding it to the `sdk.ProvisionOutput` type, we instead wrap the `sdk.ProvisionOutput` type in `proto.ProvisionCredentialResponse` type. For example:

```
type Panic struct {
       Error string
       Stack []byte 
}

type ProvisionCredentialResponse struct {
     Success sdk.ProvisionOutput
     Panic *Panic
}
```

Small downside of this is that we explicitly have to make sure to check if `Panic == nil` in the CLI, but I think that outweighs the fact that we can better tell if the plugin is panicking.

What do you think?

AndyTitu: Indeed the panic we are trying to catch does not fall into the same category as a plugin error that we can handle in the CLI but rather a RPC specific error that we need to handle at the RPC layer.

Would the proposed solution then entail changing the RPC functions' signature to 
`CredentialProvisionerProvision(req proto.ProvisionCredentialRequest, resp proto.ProvisionCredentialResponse)`, and therefore would require us to bump up the protocol version? @jpcoenen

jpcoenen: Yes, good call-out, you're right. I am not sure if it this is by itself is worth bumping the protocol version. 

We could also go for a dual approach: apply your suggestion to `main` now and also merge an improved method to the `protocol-v4` branch and leave it there until we have another change that also needs a protocol version bump. Then we immediately improve the status quo without breaking anything, but also have a nicer solution long-term. WDYT?

AndyTitu: The dual approach sounds good! I have reflected this plan in our internal issue: https://gitlab.1password.io/dev/b5/op/-/issues/3321. 

I will update all RPC functions with the simple panic catcher functionality in this PR and create another PR for merging the more complex/long term changes into `protocol-v4`.

AndyTitu: I have updated this PR accordingly and opened an another PR with the long term solution to be merged into newly created `protocol-v4` branch. https://github.com/1Password/shell-plugins/pull/149

sdk/rpc/server/server.go
```diff
@@ -117,6 +118,20 @@ func (t *RPCServer) ExecutableNeedsAuth(req proto.ExecutableNeedsAuthRequest, re
 // CredentialImport is a remote version of the Import() function in schema.CredentialType.
 // The call is forwarded to the Import() function of the credential identified by req.CredentialID.
 func (t *RPCServer) CredentialImport(req proto.ImportCredentialRequest, resp *sdk.ImportOutput) error {
+	defer func() {
+		if err := recover(); err != nil {
+			caughtPanic := fmt.Errorf("your locally built plugin's importing failed with the following panic: %s; and with stack trace: %s", err, string(debug.Stack()))
+			diagnostics := sdk.Diagnostics{Errors: []sdk.Error{{Message: caughtPanic.Error()}}}
```
jpcoenen, line 124: Can we pull this into a separate function? I think the CLI gives enough context to return a more generic error, like:

```
locally built plugin panicked: %s\nstack trace:\n%s
```

jpcoenen (commented):
One tiny remark. Apart from that, this looks good to me as a first iteration. It should provide a noticeable improvement to the development experience :+1:

jpcoenen (approved):
This looks pretty good for a temporary solution :+1: Ready to 🚢 imo!

accraw (approved):
Makes sense to me!

Issue #107, opened 2022-12-12
Issue provisioning auth using `needsauth.ForCommands`


kanadgupta:
### What does it concern?

The plugin SDK, design, or schema

### Goal or desired behavior

_(apologies in advance if this is a silly question, I'm a bit new to Go!)_

I'm attempting to create a new plugin for [`rdme`](https://www.npmjs.com/package/rdme) (see #106 #108) and I'm trying to use the `needsauth.ForCommands` helper function: https://github.com/1Password/shell-plugins/blob/33c03dd9cd7343e1fe68884d7053e242c8309312/sdk/needsauth/helpers.go#L20

When I create [a new array of commands](https://github.com/1Password/shell-plugins/blob/9fc032cd912952889fedad98536b3099334e109f/plugins/readme/rdme.go#L10-L28) and [pass it into this function](https://github.com/1Password/shell-plugins/blob/9fc032cd912952889fedad98536b3099334e109f/plugins/readme/rdme.go#L35), I would expect that each of the following commands would have the `NeedsAuthentication` rule set to `true` and therefore get the env variables provisioned properly:

```
rdme openapi
rdme docs
rdme docs:prune
rdme guides
rdme guides:prune
rdme changelogs
rdme custompages
rdme versions
rdme versions:create
rdme versions:delete
rdme versions:update
rdme categories
rdme categories:create
```

### Current behavior

When I attempt to create [the aforementioned array](https://github.com/1Password/shell-plugins/blob/9fc032cd912952889fedad98536b3099334e109f/plugins/readme/rdme.go#L10-L28) and [pass it into this function](https://github.com/1Password/shell-plugins/blob/9fc032cd912952889fedad98536b3099334e109f/plugins/readme/rdme.go#L35), env variables aren't being provisioned for the specified commands (e.g., `rdme openapi`).

I'm not sure how to debug this further but I believe it's because the plugin configuration doesn't think that those commands need auth for some reason. But when I comment out [these lines](https://github.com/1Password/shell-plugins/blob/9fc032cd912952889fedad98536b3099334e109f/plugins/readme/rdme.go#L13-L27) so the `commands` array only contains a single value, the env vars are provisioned as expected 🤔

### op CLI version

2.9.1

### shell-plugins local ref

9fc032cd912952889fedad98536b3099334e109f (full branch [here](https://github.com/kanadgupta/shell-plugins/commits/main))

florisvdg:
[...]

PR #126, opened 2022-12-23
Improve needsauth helper package

florisvdg:
This PR makes the following improvements to the `needsauth` package and related functionality:

- Remove the ability from `needsauth` helper functions to specify multiple patterns in a single rule. This removes confusion, as seen in #107.
- Make the `--version` detection more strict, to only opt out of authentication if the version arg is the only arg present, to avoid false positives such as: `deploy --version 1.0.0`.
- Also consider `-V` for the version check.
- Add a `needsauth` helper to check for invocations with empty args, which often brings up the help text, so should not require authentication.
- Add this new empty-args helper to a bunch of plugins that follow this pattern.
- Add a test helper to the `plugintest` package to easily add test cases for `NeedsAuth` entries.
- Add tests for the `needsauth` package using the new test helper.

### DIFF
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -518,7 +518,10 @@ func {{ .PlatformNameUpperCamelCase }}CLI() schema.Executable {
 		Name:      "{{ .PlatformName }} CLI", // TODO: Check if this is correct
 		Runs:      []string{"{{ .Executable }}"},
 		DocsURL:   sdk.URL("https://{{ .Name }}.com/docs/cli"), // TODO: Replace with actual URL
-		NeedsAuth: needsauth.NotForHelpOrVersion(),
+		NeedsAuth: needsauth.For(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
 		{{- if .CredentialName }}
 		Uses: []schema.CredentialUsage{
 			{
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -504,7 +504,7 @@ func {{ .PlatformNameUpperCamelCase }}CLI() schema.Executable {
 		Name:      "{{ .PlatformName }} CLI", // TODO: Check if this is correct
 		Runs:      []string{"{{ .Executable }}"},
 		DocsURL:   sdk.URL("https://{{ .Name }}.com/docs/cli"), // TODO: Replace with actual URL
-		NeedsAuth: needsauth.For(
+		NeedsAuth: needsauth.IfAll(
 			needsauth.NotForHelpOrVersion(),
 			needsauth.NotWithoutArgs(),
 		),
--- a/plugins/cargo/cargo.go
+++ b/plugins/cargo/cargo.go
@@ -7,18 +7,16 @@ import (
 	"github.com/1Password/shell-plugins/sdk/schema/credname"
 )
 
-var commands = [][]string{
-	{"publish"},
-	{"yank"},
-	{"owner"},
-}
-
 func CargoCLI() schema.Executable {
 	return schema.Executable{
-		Name:      "Cargo CLI",
-		Runs:      []string{"cargo"},
-		DocsURL:   sdk.URL("https://doc.rust-lang.org/cargo/index.html"),
-		NeedsAuth: needsauth.ForCommands(commands...),
+		Name:    "Cargo CLI",
+		Runs:    []string{"cargo"},
+		DocsURL: sdk.URL("https://doc.rust-lang.org/cargo/index.html"),
+		NeedsAuth: needsauth.OnlyFor(
+			needsauth.ForCommand("publish"),
+			needsauth.ForCommand("yank"),
+			needsauth.ForCommand("owner"),
+		),
 		Uses: []schema.CredentialUsage{
 			{
 				Name: credname.APIToken,
--- a/plugins/cargo/cargo.go
+++ b/plugins/cargo/cargo.go
@@ -12,8 +12,8 @@ func CargoCLI() schema.Executable {
 		Name:    "Cargo CLI",
 		Runs:    []string{"cargo"},
 		DocsURL: sdk.URL("https://doc.rust-lang.org/cargo/index.html"),
-		NeedsAuth: needsauth.For(
-			needsauth.OnlyFor(
+		NeedsAuth: needsauth.IfAll(
+			needsauth.IfAny(
 				needsauth.ForCommand("publish"),
 				needsauth.ForCommand("yank"),
 				needsauth.ForCommand("owner"),
--- a/plugins/cargo/cargo.go
+++ b/plugins/cargo/cargo.go
@@ -12,10 +12,13 @@ func CargoCLI() schema.Executable {
 		Name:    "Cargo CLI",
 		Runs:    []string{"cargo"},
 		DocsURL: sdk.URL("https://doc.rust-lang.org/cargo/index.html"),
-		NeedsAuth: needsauth.OnlyFor(
-			needsauth.ForCommand("publish"),
-			needsauth.ForCommand("yank"),
-			needsauth.ForCommand("owner"),
+		NeedsAuth: needsauth.For(
+			needsauth.OnlyFor(
+				needsauth.ForCommand("publish"),
+				needsauth.ForCommand("yank"),
+				needsauth.ForCommand("owner"),
+			),
+			needsauth.NotForHelp(),
 		),
 		Uses: []schema.CredentialUsage{
 			{
--- a/plugins/homebrew/brew.go
+++ b/plugins/homebrew/brew.go
@@ -12,9 +12,9 @@ func HomebrewCLI() schema.Executable {
 		Name:    "Homebrew CLI",
 		Runs:    []string{"brew"},
 		DocsURL: sdk.URL("https://brew.sh/"),
-		NeedsAuth: needsauth.For(
+		NeedsAuth: needsauth.IfAll(
 			needsauth.NotForHelpOrVersion(),
-			needsauth.OnlyFor(
+			needsauth.IfAny(
 				needsauth.ForCommand("search"),
 				needsauth.ForCommand("bump-cask-pr"),
 				needsauth.ForCommand("bump-formula-pr"),
--- a/plugins/homebrew/brew.go
+++ b/plugins/homebrew/brew.go
@@ -7,19 +7,20 @@ import (
 	"github.com/1Password/shell-plugins/sdk/schema/credname"
 )
 
-var commands = [][]string{
-	{"search"},
-
-	{"bump-cask-pr"},
-	{"bump-formula-pr"},
-}
-
 func HomebrewCLI() schema.Executable {
 	return schema.Executable{
-		Name:      "Homebrew CLI",
-		Runs:      []string{"brew"},
-		DocsURL:   sdk.URL("https://brew.sh/"),
-		NeedsAuth: needsauth.ForCommands(commands...),
+		Name:    "Homebrew CLI",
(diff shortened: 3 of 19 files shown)


cmd/contrib/main.go
```diff
@@ -518,7 +518,10 @@ func {{ .PlatformNameUpperCamelCase }}CLI() schema.Executable {
 		Name:      "{{ .PlatformName }} CLI", // TODO: Check if this is correct
 		Runs:      []string{"{{ .Executable }}"},
 		DocsURL:   sdk.URL("https://{{ .Name }}.com/docs/cli"), // TODO: Replace with actual URL
-		NeedsAuth: needsauth.NotForHelpOrVersion(),
+		NeedsAuth: needsauth.For(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
```
florisvdg, line 524: Any CLIs in our existing plugin catalog I'm missing that should also be using this new empty-args helper that opts out of authentication when no args are specified?

kanadgupta (approved):
I was literally about to file an issue requesting a test helper for the `needsauth` logic in each plugin, thank you so much! Lots of great stuff here, excited to see it go out 🥳

plugins/homebrew/brew.go
```diff
@@ -7,19 +7,20 @@ import (
...
-		Runs:      []string{"brew"},
-		DocsURL:   sdk.URL("https://brew.sh/"),
-		NeedsAuth: needsauth.ForCommands(commands...),
+		Name:    "Homebrew CLI",
+		Runs:    []string{"brew"},
+		DocsURL: sdk.URL("https://brew.sh/"),
+		NeedsAuth: needsauth.For(
+			needsauth.NotWithoutArgs(),
+			needsauth.NotForHelpOrVersion(),
+			needsauth.OnlyFor(
```
SimonBarendse, line 18: Does `OnlyFor` already imply `NotWithoutArgs`?

SimonBarendse (approved):
Nice! :raised_hands:

plugins/homebrew/brew.go
florisvdg: Yes, good catch! Fixed.

SimonBarendse (dismissed):

hculea:
I brought the MR up to date, to make it ready for merging with my last two commits.

I also went through the rest of the changes - looks good to me! However, to make sure I don't end up self-reviewing, I'll not be approving myself, but instead wait for two of my colleagues to get a chance to review too.

sdk/needsauth/helpers.go
```diff
@@ -15,22 +17,33 @@ func For(rules ...sdk.NeedsAuthentication) sdk.NeedsAuthentication {
 	}
 }
 
-// ForCommands returns a NeedsAuthentication rule to require authentication for
-// certain (sub)command, e.g. ["account"] or ["account", "list"], ["account", "delete"].
-func ForCommands(commands ...[]string) sdk.NeedsAuthentication {
+// OnlyFor returns a NeedsAuthentication rule that only opts in of the authentication requirement if there's
```
AndyTitu, line 20:
```suggestion
// OnlyFor returns a NeedsAuthentication rule that opts out of the authentication requirement, unless if there's at least one rule specified that opts in to the authentication requirement.
```

sdk/needsauth/helpers.go
```diff
@@ -1,9 +1,11 @@
...
 
-import "github.com/1Password/shell-plugins/sdk"
+import (
+	"github.com/1Password/shell-plugins/sdk"
+)
 
-// For returns a NeedsAuthentication rule that iterates over other NeedsAuthentication rules
-// until there's one that opts out of the authentication requirement.
+// For returns a NeedsAuthentication that opts in to the authentication requirement, unless there's
+// one rule specified that opts out to the authentication requirement.
```
AndyTitu, line 8:
```suggestion
// one rule specified that opts out of the authentication requirement.
```

sdk/needsauth/helpers.go
```diff
@@ -45,27 +58,79 @@ func Always() sdk.NeedsAuthentication {
...
+			}
+		}
+
+		return false
+	}
+}
+
+// NotWhenContainsArgs returns a NeedsAuthentication rule to not require authentication when
+// the exact sequence of argsToSkip is present somewhere in the command-line args.
+func NotWhenContainsArgs(argsToSkip ...string) sdk.NeedsAuthentication {
```
AndyTitu, line 81:
```suggestion
func NotWhenContainsArgsSequence(argsToSkip []string) sdk.NeedsAuthentication {
```
Should this be renamed by adding the `Sequence` keyword so to imply that a match is made for >2 terms only if they are found in the exact order they have been specified?

What would be the use case of calling `NotWhenContainsArgsSequence()`?

AndyTitu (commented):
The changes look good to me! I spent some time trying to understand the semantics of the `For` and `OnlyFor` functions and I realised the naming only makes sense when knowing that `For` is only meant to be used by `Not...` functions such as `NotWhenContainsArgs` (if someone were to use `For` with `ForCommand`, it would result in the plugin not needing authentication for that command, and needing authentication for all the rest of the commands, which is against what the naming suggests). The same goes for `OnlyFor` but mirrored. I think this should be mentioned in the comments, as I could not find better names for these functions that would universally make sense. I also made some other suggestions for the comments for clarity and a function name and parameter for clarity.

hculea:
> The changes look good to me! I spent some time trying to understand the semantics of the `For` and `OnlyFor` functions and I realised the naming only makes sense when knowing that `For` is only meant to be used by `Not...` functions such as `NotWhenContainsArgs` (if someone were to use `For` with `ForCommand`, it would result in the plugin not needing authentication for that command, and needing authentication for all the rest of the commands, which is against what the naming suggests). The same goes for `OnlyFor` but mirrored. I think this should be mentioned in the comments, as I could not find better names for these functions that would universally make sense. I also made some other suggestions for the comments for clarity and a function name and parameter for clarity.

Interesting perspective! What do you think if we named them according to the logical operators which they double? I think this would make it a lot clearer.

e.g. 

`OnlyFor` is a logical `Or`:
```
needsauth.OnlyFor(
			needsauth.ForCommand("publish"),
			needsauth.ForCommand("yank"),
			needsauth.ForCommand("owner"),
		)
```
This translates to "If the command is publish OR yank OR owner, ask for auth".

`For` is a logical `And`:
```
needsauth.For(
			needsauth.NotWithoutArgs(),
		),
```

This translates to "if the command isn't help or version AND isn't without args, ask for auth."

Using `For` with `ForCommand` would however still authenticate that command. Using it with multiple `ForCommand`s would mean that nothing gets authed anymore, since you can't have multiple same level commands at the same time. Is this your understanding as well, @AndyTitu ?

plugins/cargo/cargo.go
```diff
@@ -7,18 +7,16 @@ import (
...
-		DocsURL:   sdk.URL("https://doc.rust-lang.org/cargo/index.html"),
-		NeedsAuth: needsauth.ForCommands(commands...),
+		Name:    "Cargo CLI",
+		Runs:    []string{"cargo"},
+		DocsURL: sdk.URL("https://doc.rust-lang.org/cargo/index.html"),
+		NeedsAuth: needsauth.OnlyFor(
+			needsauth.ForCommand("publish"),
+			needsauth.ForCommand("yank"),
+			needsauth.ForCommand("owner"),
+		),
```
florisvdg, line 19: There should be an exception for `--help` when using these commands.

sdk/needsauth/helpers.go
florisvdg: Or how about:
```suggestion
func NotWhenContainsArgs(argsSequence ...string) sdk.NeedsAuthentication {
```

AndyTitu:
> Using it with multiple `ForCommand`s would mean that nothing gets authed anymore, since you can't have multiple same level commands at the same time. Is this your understanding as well, @AndyTitu ?

Yes, this is what I meant. And because the function is named `For` people could think it would work otherwise.

> `OnlyFor` is a logical `Or`; `For` is a logical `And`
Yes, that is true now that I think about it. But my impression is that `OnlyFor` was designed to only work for non negated arguments and `For` only with negated arguments(negated argument = `NotWithoutArgs` and similar functions, non negated argument = `ForCommand` function). I say this because, as you also noted, using `For` with multiple `ForCommands` is not working as expected (assuming the word `for` is interpreted as a for), because the statement will always be false. So, combining what you said with what I said, the use of `OnlyFor` is arg1 OR arg2, whereas `For` is not(arg1) AND not(arg2) (= not(arg1 OR arg2)). I would propose we document this in the comments of each function and rename the functions as such `OnlyFor` -> `IfEitherMatches`, `For`-> `IfAllMatch` or something similar.

plugins/cargo/cargo.go
AndyTitu: This would mean (`ForCommand(publish)` OR `ForCommand(yank)` OR `ForCommand(owner)`) AND `NotForHelp()` (as Horia [stated](https://github.com/1Password/shell-plugins/pull/126#issuecomment-1387166231) the `For` is a logical AND and `OnlyFor` is logical `OR`).

So we could do  
```
needsauth.For(
  needsauth.OnlyFor(
			  needsauth.ForCommand("publish"),
			  needsauth.ForCommand("yank"),
			  needsauth.ForCommand("owner"),
  ),
  needsauth.NotForHelp()
)
```
But I would say this is also a good argument for renaming these functions `OnlyFor` -> `IfEitherMatches `, `For`-> `IfAllMatch`

hculea:
I like it, I think we're approaching something more intuitive than the status quo. 😄 

I'll rename them slightly differently from your suggestion:
`OnlyFor` -> `IfAny` (using either suggests it would only take two arguments, which is not true)
`For` -> `IfAll`

Using `match` is a bit redundant, I hope these names capture the essence of the functions as succinctly as possible.

sdk/needsauth/helpers.go
AndyTitu: Yes, this makes sense.

hculea:
Done. I refrained from adding specific rules of thumb such as "Use IfAll only for negated statements" since there are also use-cases where these do not hold.

E.g. a plugin only requires auth for a single command, but only if `--auth` is passed. This would be:
```
needsauth.IfAll(
  ForCommand("mycmd"),
  ContainsArg("--auth"),
)
```

This is a toy example, but just to show that we shouldn't give erroneous indications (especially in the code comments) that don't universally hold.

AndyTitu (approved):
lgtm now thanks!

accraw (approved):
Makes sense to me!

SimonBarendse (approved):

SimonBarendse:
>  The changes look good to me! I spent some time trying to understand the semantics of the For and OnlyFor functions and I realised the naming only makes sense when knowing that For is only meant to be used by Not... functions such as NotWhenContainsArgs (if someone were to use For with ForCommand, it would result in the plugin not needing authentication for that command, and needing authentication for all the rest of the commands, which is against what the naming suggests).

We can let the compiler help here by having the Not functions have a different signature, and the `For` and `OnlyFor` (or `NotFor`) functions each accept only those functions with the signature that signals they were built to be used for them.

hculea:
I'll not implement this for the time being, given that there are use-cases where `For` and `OnlyFor` can be used for more than the finite set of negatory functions. Have a look at [this](https://github.com/1Password/shell-plugins/pull/126/files#diff-b81f6f744996a8ce8c540ea537fcc24fbde8908f861e549a4de005a8ffe9824fR110) test for an example of a more complex use-case. 

I propose we revisit this decision if the ease of use for this package will indeed prove to be a problem. Let me know if you disagree.

accraw (approved):

PR #127, opened 2022-12-23
Add importer helpers to support OS-specific importers

florisvdg:
### Usage

```go
Importer: importer.TryAll(
	importer.TryAllEnvVars(fieldname.Token, "EXAMPLE_ACCESS_TOKEN"),
	importer.MacOnly(
		TryExampleConfigFile("~/Library/Application Support/example/config.yml"),
	),
	importer.LinuxOnly(
		TryExampleConfigFile("$XDG_CONFIG_HOME/example/config.yml"),
	),
),
```

### DIFF
--- a/sdk/importer.go
+++ b/sdk/importer.go
@@ -40,6 +40,9 @@ type ImportSource struct {
 type ImportInput struct {
 	HomeDir string
 	RootDir string
+
+	// Supported values: "darwin", "linux"
+	OS string
 }
 
 type ImportOutput struct {
--- a/sdk/importer/helpers.go
+++ b/sdk/importer/helpers.go
@@ -14,6 +14,26 @@ func TryAll(importers ...sdk.Importer) sdk.Importer {
 	}
 }
 
+func MacOnly(importers ...sdk.Importer) sdk.Importer {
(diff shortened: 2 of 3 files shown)


shyim:
I am not sure if this is the right way to handle this.

It would make more sense to prefix the path with https://pkg.go.dev/os#UserConfigDir and be then platform independent. There is already an abstraction in Go we should not built it by own 😄 

Also you can go into weird edge cases like, you have Linux but not defined env `$XDG_CONFIG_HOME` for some reason

florisvdg:
The pattern used in Go's `UserConfigDir` is not widely spread enough to base this feature around it, unfortunately. There's many tools that use `~/.config` or `XDG_CONFIG_HOME` on Mac as well, instead of `Application Support`. That said, we can always still introduce a platform-independent shorthand later.

Also, the code snippet in this PR description was just one example use case for the `MacOnly` and `LinuxOnly` helpers introduced in this PR.

AndyTitu (approved):
current changes LGTM.

hculea (approved):

AndyTitu:
>  we can always still introduce a platform-independent shorthand later.

++

I will be merging this since it's treated as blocker for [DigitalOcean config file importer on macOS](https://github.com/1Password/shell-plugins/pull/109)

Issue #93, opened 2022-12-09
New plugin: Flyctl


shyim:
### Platform or tool

Fly.io

### Relevant links

- https://fly.io

### Goal or desired behavior

Just set a `FLY_ACCESS_TOKEN` it's easy

PR #141, opened 2023-01-12
Add support for flyctl

arunsathiya:
## Description

Add support for flyctl. Fixes https://github.com/1Password/shell-plugins/issues/93.

## Roadblocks with naming

`Flyctl` is the [actual product name](https://fly.io/docs/flyctl/), while they offer both `fly` and `flyctl` as valid commands. From [this comment](https://community.fly.io/t/timeline-for-flyctl-fly-rename/4453/5), it seems that `fly` will remain the primary command while also retaining `flyctl` as an alias.

For this reason, I named this shell plugin `flyctl` while specifying `fly` as the executable command. Is this fine?

## Other questions

It doesn't seem that running `op plugin init fly` automatically sets up `alias fly="op plugin run -- fly"` on `~/.config/op/plugins.sh` like it happens for `gh` or `aws`. I had to manually add that alias and run `source ~/.config/op/plugins.sh`. Is this normal?

I had to also add `alias flyctl="op plugin run -- fly"` so that the shell plugin can take over when the user runs `flyctl` commands.

### DIFF
--- a/plugins/flyctl/fly.go
+++ b/plugins/flyctl/fly.go
@@ -1,0 +1,22 @@
+package flyctl
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func FlyctlCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Flyctl",
+		Runs:      []string{"fly"},
+		DocsURL:   sdk.URL("https://fly.io/docs/flyctl/"),
+		NeedsAuth: needsauth.NotForHelpOrVersion(),
+		Uses: []schema.CredentialUsage{
+			{
+				Name: credname.AccessToken,
+			},
+		},
+	}
+}
--- a/plugins/flyctl/access_token.go
+++ b/plugins/flyctl/access_token.go
@@ -1,0 +1,68 @@
+package flyctl
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.AccessToken,
+		DocsURL:       sdk.URL("https://fly.io/docs/flyctl/auth-login/"),
+		ManagementURL: sdk.URL("https://fly.io/user/personal_access_tokens"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Token,
+				MarkdownDescription: "Token used to authenticate to Fly.io.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 43,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+						Symbols:   true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryEnvVarPair(defaultEnvVarMapping),
+			TryFlyctlConfigFile(),
+		)}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"FLY_ACCESS_TOKEN": fieldname.Token,
+}
+
+func TryFlyctlConfigFile() sdk.Importer {
+	return importer.TryFile("~/.fly/config.yml", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		var config Config
+		if err := contents.ToYAML(&config); err != nil {
+			out.AddError(err)
+			return
+		}
+
+		if config.Token == "" {
+			return
+		}
+
+		out.AddCandidate(sdk.ImportCandidate{
+			Fields: map[sdk.FieldName]string{
+				fieldname.Token: config.Token,
+			},
+		})
+	})
+}
+
+type Config struct {
+	Token string `yaml:"access_token"`
+}
--- a/plugins/flyctl/access_token_test.go
+++ b/plugins/flyctl/access_token_test.go
@@ -1,0 +1,53 @@
+package flyctl
+
+import (
+	"testing"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func TestAccessTokenProvisioner(t *testing.T) {
+	plugintest.TestProvisioner(t, AccessToken().DefaultProvisioner, map[string]plugintest.ProvisionCase{
+		"default": {
+			ItemFields: map[sdk.FieldName]string{
+				fieldname.Token: "DtP7HoOPOBHMP6bE5tx3nguB5r2zPpSbg9hlEXAMPLE",
+			},
+			ExpectedOutput: sdk.ProvisionOutput{
+				Environment: map[string]string{
+					"FLY_ACCESS_TOKEN": "DtP7HoOPOBHMP6bE5tx3nguB5r2zPpSbg9hlEXAMPLE",
+				},
+			},
+		},
+	})
+}
+
+func TestAccessTokenImporter(t *testing.T) {
+	plugintest.TestImporter(t, AccessToken().Importer, map[string]plugintest.ImportCase{
+		"environment": {
+			Environment: map[string]string{
+				"FLY_ACCESS_TOKEN": "DtP7HoOPOBHMP6bE5tx3nguB5r2zPpSbg9hlEXAMPLE",
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{
+					Fields: map[sdk.FieldName]string{
+						fieldname.Token: "DtP7HoOPOBHMP6bE5tx3nguB5r2zPpSbg9hlEXAMPLE",
+					},
+				},
+			},
+		},
+		"config file": {
+			Files: map[string]string{
+				"~/.fly/config.yml": plugintest.LoadFixture(t, "config.yml"),
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{
+					Fields: map[sdk.FieldName]string{
+						fieldname.Token: "DtP7HoOPOBHMP6bE5tx3nguB5r2zPpSbg9hlEXAMPLE",
+					},
+				},
+			},
+		},
+	})
+}
--- a/plugins/flyctl/plugin.go
+++ b/plugins/flyctl/plugin.go
@@ -1,0 +1,22 @@
+package flyctl
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "flyctl",
+		Platform: schema.PlatformInfo{
+			Name:     "Fly.io",
+			Homepage: sdk.URL("https://fly.io/docs/flyctl/"),
+		},
+		Credentials: []schema.CredentialType{
+			AccessToken(),
+		},
+		Executables: []schema.Executable{
+			FlyctlCLI(),
+		},
+	}
+}
--- a/plugins/flyctl/test-fixtures/config.yml
+++ b/plugins/flyctl/test-fixtures/config.yml
@@ -1,0 +1,2 @@
+access_token: DtP7HoOPOBHMP6bE5tx3nguB5r2zPpSbg9hlEXAMPLE
+wire_guard_state: {}


arunsathiya:
>It doesn't seem that running op plugin init fly automatically sets up alias fly="op plugin run -- fly" on ~/.config/op/plugins.sh like it happens for gh or aws. I had to manually add that alias and run source ~/.config/op/plugins.sh. Is this normal?

To elaborate on this a bit, here's how the flow looks when testing my local build:

```
➜  shell-plugins git:(add/flyctl) op plugin list
SOURCE                USAGE         NAME                  REQUIRED FIELDS
Local Build           fly           Fly.io                Token
1Password Registry    aws           AWS                   Access Key ID, Secret Access Key
.
.
.
➜  shell-plugins git:(add/flyctl) op plugin init fly
#######################################################################
# WARNING: 'flyctl' is not from the official registry.                #
# Only proceed if you are the developer of 'flyctl'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/flyctl. #
#######################################################################

Flyctl [test build]
Authenticate with Fly.io Access Token.

? Locate your Fly.io Access Token: Search in 1Password...

? Locate your Fly.io Access Token: Fly.io Access Token (Personal)

? Configure when "Fly.io Access Token" will be used to authenticate: Prompt me for each new terminal session

Successfully set up authentication for Flyctl[31;1m [test build][0m.
Run any fly command to see it action!

➜  shell-plugins git:(add/flyctl) fly auth whoami
Error No access token available. Please login with 'flyctl auth login'


```

The command above fails because I don't have `~/.fly/config.yml` on my computer. My expectation is that the token is sourced from 1Password at this stage.

Since that doesn't work, I manually edit `~/.config/op/plugins.sh` and it works:

```
➜  shell-plugins git:(add/flyctl) code ~/.config/op/plugins.sh
➜  shell-plugins git:(add/flyctl) source ~/.config/op/plugins.sh
➜  shell-plugins git:(add/flyctl) fly auth whoami
#######################################################################
# WARNING: 'flyctl' is not from the official registry.                #
# Only proceed if you are the developer of 'flyctl'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/flyctl. #
#######################################################################
redacted
```

plugins/flyctl/fly.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func FlyctlCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Flyctl",
+		Runs:      []string{"fly"},
```
arunsathiya, line 13: Does `Executable.Runs` not support more than one command at this time? 

I tried `Runs:      []string{"fly", "flyctl"},`  but after building the local binary, setting up the plugin fails:

```
➜  shell-plugins git:(add/flyctl) ✗ make flyctl/build
# BETA NOTICE: The plugin ecosystem is in beta and is subject to change.
# You may have to update or recompile your local builds every now and then to keep them
# compatible with the 1Password CLI updates.

go build -o ~/.op/plugins/local/flyctl -ldflags="-X 'main.PluginName=flyctl'" ./cmd/contrib/build/
➜  shell-plugins git:(add/flyctl) ✗ op plugin list
SOURCE                USAGE         NAME                  REQUIRED FIELDS
Local Build           fly flyctl    Fly.io                Token
1Password Registry    aws           AWS                   Access Key ID, Secret Access Key
.
.
.
.
.
.
➜  shell-plugins git:(add/flyctl) ✗ op plugin init fly
[ERROR] 2023/01/12 02:26:54 unknown plugin: fly
➜  shell-plugins git:(add/flyctl) ✗ op plugin init flyctl
[ERROR] 2023/01/12 02:26:56 unknown plugin: flyctl
```

@SimonBarendse, it's at this point, I scheduled a Calendly pairing session but managed to resolve this by completely removing `flyctl`.

SimonBarendse:
> For this reason, I named this shell plugin flyctl while specifying fly as the executable command. Is this fine?

Given `flyctl` is the binary name, and currently `fly` is aliased to `flyctl`, I believe using `flyctl` for the plugin is the most flexible implementation, because it will work both when you invoke using `fly` as a user and when you use `flyctl`:

Alias resolution when the plugin aliases `flyctl`:
- `fly` -> `flyctl` -> `op plugin run -- flyctl`
- `flyctl` -> `op plugin run -- flyctl`

Alias resolution when the plugin aliases `fly`:
- `fly` -> `op plugin run -- fly` -> `flyctl` (during execution of `op plugin run`)
- `flyctl` -> `fly` (:warning: note no `op plugin run` here)

Furthermore, the plugin will be more explicit, because it will alias the binary name rather than another alias. This is important to us, because it helps in the transparency of what the plugin executes.

Therefore, please use `flyctl` as the command for the plugin.

SimonBarendse:
> It doesn't seem that running op plugin init fly automatically sets up alias fly="op plugin run -- fly" on ~/.config/op/plugins.sh like it happens for gh or aws. I had to manually add that alias and run source ~/.config/op/plugins.sh. Is this normal?

> I had to also add alias flyctl="op plugin run -- fly" so that the shell plugin can take over when the user runs flyctl commands.

No, this is not expected. I'm hoping my reply above will solve both. Please let us know if my above suggestion does not solve this, as we'd have to dig deeper to solve this in that case.

arunsathiya:
Thanks for reviewing, @SimonBarendse! I have been at this for the last two days but I am stuck. Here's my progress:

- Renamed plugin command to `flyctl`
- Shell Plugins execute for `flyctl` commands.
- But for `fly` commands, the CLI seems to look for the auth token locally on my computer and it fails.

Visualization of the command flow:

```
➜  ~ flyctl auth whoami
#######################################################################
# WARNING: 'flyctl' is not from the official registry.                #
# Only proceed if you are the developer of 'flyctl'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/flyctl. #
#######################################################################
email-address-redacted
➜  ~ fly auth whoami
Error No access token available. Please login with 'flyctl auth login'

```

When I add a new alias for `fly` on `~/.op/plugins.sh` with value of `op plugin run -- flyctl`, it works as expected:

```
➜  ~ cat ~/.op/plugins.sh
export OP_PLUGIN_ALIASES_SOURCED=1
alias gh="op plugin run -- gh"
alias src="op plugin run -- src"
alias railway="op plugin run -- railway"
alias aws="op plugin run -- aws"
alias flyctl="op plugin run -- flyctl"
alias fly="op plugin run -- flyctl"
➜  ~ source ~/.op/plugins.sh
➜  ~ fly auth whoami
#######################################################################
# WARNING: 'flyctl' is not from the official registry.                #
# Only proceed if you are the developer of 'flyctl'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/flyctl. #
#######################################################################
email-address-redacted
```

But based on [your comment here](https://github.com/1Password/shell-plugins/pull/141#issuecomment-1383788095), it sounds like the second alias shouldn't be needed? 🤔 At this point, I am stuck and wanted to share my progress to move forward our discussion.

arunsathiya:
>No, this is not expected.

Regarding this part, I noticed that I was looking at `~/.config/op/plugins.sh` earlier but I see that the correct/actual path is `~/.op/plugins.sh`. Is there a conversation/Slack thread/GitHub link where I can read more about the differences between the two files?

arunsathiya:
There's no issue with the `op plugin init plugin-name` command by the way. It works as expected! ✨

SimonBarendse:
> I noticed that I was looking at ~/.config/op/plugins.sh earlier but I see that the correct/actual path is ~/.op/plugins.sh. Is there a conversation/Slack thread/GitHub link where I can read more about the differences between the two files?

Does https://developer.1password.com/docs/cli/config-directories help?

SimonBarendse:
Based on what you're sharing I have a hunch that _maybe_  `fly` -> `flyctl` may not be implemented with an alias. Could you share the output of `which fly` when you don't have `alias fly="op plugin run -- flyctl"` manually added to the alias file?

arunsathiya:
It looks like the two commands are two different binaries:

```
➜  ~ which fly
/opt/homebrew/bin/fly
➜  ~ which flyctl
/opt/homebrew/bin/flyctl

➜  ~ type fly
fly is /opt/homebrew/bin/fly
➜  ~ type flyctl
flyctl is /opt/homebrew/bin/flyctl
```

I Googled a bit and came across [this blog post](https://fly.io/blog/the-september-fly-changelog-new-names-and-easier-updates/):

>We've modified the installer for the Fly CLI so that while it installs itself as flyctl it also sets up a symbolic link to the fly command name. Once you are up to date with Flyctl, you'll be able to use fly as your preferred command. We're taking this carefully as this is the start of a migration which will end with fly being the actual name of the CLI app.

At this stage, my understanding stands that `fly` is not an alias to `flyctl`, rather the Fly.io platform does something special to sync the data between the two binaries. 🤔 Perhaps someone from Fly.io (tagging two folks that I know of, @DAlperin and @michaeldwan: hi!) could shed some light?

DAlperin:
Hey @arunsathiya :). The `fly` binary is a symlink to `flyctl`. You can see this by running `ls -l /opt/homebrew/Cellar/flyctl/$(flyctl version | cut -d' ' -f 2 | cut -c2-)/bin`

hculea:
Hey @arunsathiya, is there anything that we can help with, to go forward with this MR?

Would love to help get this to the finish line! Let us know! 😄

arunsathiya:
Thank you, @DAlperin, and thank you for your patience, @hculea!

A quick note that I didn't get around taking another look at this PR yet. With the start of my new semester classes, things have been a bit packed at my end but this is on my radar for the weekend. I'll aim to have an update by Monday.

AndyTitu (approved):
The current version of this PR lgtm!

arunsathiya:
Just wanted to make a quick note that I've started exploring an option today to support aliases in shell plugin creation flow, and then generate the shell plugin alias in `~/.op/config/plugins.sh` for all plugin aliases. That way the main binary name and all aliases will work. 

I suggest we hold off on merging the current version because it doesn't actually work for "fly" commands. It does only for "flyctl".

hculea:
Sounds good, please do let us know if we can help in any way! In the meantime, feel free to mark this PR as 'Draft', to make sure it doesn't get any reviews, since the version is not final 😄 

And thank you for doing such a thorough job in trying to come up with the best solution for this rather uncommon situation.

arunsathiya:
Hi @hculea, sorry about the rather long silence on this -- been a busy-ish month for me. 

I had a detailed look at what we could do in this case, and here's my understanding: it is possible to introduce a new survey prompt on the [newPlugin()](https://github.com/1Password/shell-plugins/blob/72f2370edfc8428cf11a4e471fca3535f5732e69/cmd/contrib/main.go#L90-L249) setup flow, where the alternative CLI aliases/symlinks (in this case, `fly`) can be collected, but the final step of introducing `alias fly="op plugin run -- fly"` on the `~/.op/plugins.sh` file needs to be done manually. It's only `alias flyctl="op plugin run -- flyctl"` that's added automatically at the moment.

In order to also introduce `alias fly="op plugin run -- fly"`, I think the changes need to be introduced in the OP CLI code itself. I am not seeing any way to do that within the Shell Plugins project.

Please advise if my understanding is correct here, and if there are other ways to achieve the overall goal of supporting both `flyctl` and `fly`, let me know too.

arunsathiya:
For now, I have rebased with the main branch.

hculea:
Hey Arun, terribly sorry for the late response, looks like I've missed your reply!

In order to request changes needing to be made in `1Password CLI`, feel free to open up a new issue using the `1Password CLI overall shell plugins functionality` template. We'll make sure to pick it up internally 😄 

As for possible workarounds for this, the plugin ecosystem, in its inception, offers support for multi-executable plugins. The limitations around this is imposed by 1Password CLI, which handles only one exec for plugin.

Would adding support for this feature also help in this case, in order to circumvent adding an additional prompt?

arunsathiya:
>As for possible workarounds for this, the plugin ecosystem, in its inception, offers support for multi-executable plugins. The limitations around this is imposed by 1Password CLI, which handles only one exec for plugin.

That is interesting, thanks for sharing that @hculea! 

I have started a conversation here:

- https://github.com/1Password/shell-plugins/issues/230

Until we have a way to move forward, I'll mark this PR as a draft.

hculea:
Keep an eye on https://github.com/1Password/shell-plugins/pull/232, I think the merging of this will get you unblocked 😄

hculea:
@arunsathiya looks like this may be good to go too! Let's add both flyctl and fly as executables.

smorimoto:
@hculea If @arunsathiya allows the maintainers to edit, I think you all can move this PR forward 🙂

plugins/flyctl/fly.go
arunsathiya: This should be addressed by https://github.com/1Password/shell-plugins/pull/141/commits/01c6f85fc3113c02bbcefdde6e01d1d53e6576e8.

arunsathiya:
Sorry folks, been away. 

It's great that multiple executables can be defined within a single plugin! I have defined both `flyctl` and `fly` as executables within the `flyctl` shell plugin. 

I have tested with `op plugin init flyctl` and `op plugin init fly` and both aliases get configured as expected:

```
alias fly="op plugin run -- fly"
alias flyctl="op plugin run -- flyctl"
```

Other commands work well too: `op plugin list`, `op plugin clear flyctl`, `op plugin clear fly`, `op plugin inspect flyctl`, `op plugin inspect fly`

@hculea I think this PR is good to go now, but let me know if you have any questions/comments.

AndyTitu (approved):
lgtm! happy that multi executable support was able to unblock this!

SimonBarendse (approved):
Thanks for the contribution @arunsathiya! Happy we found a solution here 😄

PR #143, opened 2023-01-13
add support for Databricks

bsamseth:
This adds support for the [Databricks CLI](https://docs.databricks.com/dev-tools/cli/index.html). 

This CLI needs to know a "host" URL to the instance of Databricks you're working with, as well as your credentials that are specific to that instance. Personal access tokens are they primary way to authenticate, while username+password is possible but discouraged.

I've tested this on my own config, and both import and usage seems to work just fine 💯  Please let me know if you'd like to see more test cases or anything else before accepting. 

import

A couple notes: 
- I've opted for the "Person Access Token" credentials type, despite username+password being an option. The code here handles both cases, but the label of "Personal Access Token" isn't quite right in both cases. I still think it should be as I've proposed here to encourage the recommended flow, but an alternative would be to use a more generic "Configuration" label. A second opinion here would be nice 🙂 
- When doing `op plugin list`, this plugin shows up with `REQUIRED FIELDS: host`. This is because while you do need to provide _either_ a token _or_ a username and password, neither of them can be marked as required on their own. Probably not a big deal, but resolving it can't really be done without making some changes to `schema.CredentialField` and related code.
- The config files at the moment typically also includes a `jobs-api-version = 2.x` line. This is ignored by the plugin with the following reasoning:
  1. Unlike the other config options, the CLI does not accept setting the API version with an environment variable. In order to support this the plugin would have to make use of the file-based provisioning variant, as well as add a highly specific field definition to the SDK. While possible, I don't believe it is justified given the next two points.
  2. The version _can_ be set with a command line flag, and _is only relevant for the `databricks jobs` subcommand_.
  3. Perhaps most importantly, **this version flag is temporary**, and future versions of the CLI will default to the newer API version.

### DIFF
--- a/plugins/databricks/personal_access_token.go
+++ b/plugins/databricks/personal_access_token.go
@@ -15,9 +15,6 @@ func PersonalAccessToken() schema.CredentialType {
 	return schema.CredentialType{
 		Name:    credname.PersonalAccessToken,
 		DocsURL: sdk.URL("https://docs.databricks.com/dev-tools/auth.html"),
-		// The management URL is spesific to each deployment of Databricks, but this
-		// link describes how to manage tokens provided you are logged in.
-		ManagementURL: sdk.URL("https://docs.databricks.com/dev-tools/auth.html"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.Host,
--- a/plugins/databricks/personal_access_token.go
+++ b/plugins/databricks/personal_access_token.go
@@ -1,0 +1,118 @@
+package databricks
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func PersonalAccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.PersonalAccessToken,
+		DocsURL: sdk.URL("https://docs.databricks.com/dev-tools/auth.html"),
+		// The management URL is spesific to each deployment of Databricks, but this
+		// link describes how to manage tokens provided you are logged in.
+		ManagementURL: sdk.URL("https://docs.databricks.com/dev-tools/auth.html"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Host,
+				MarkdownDescription: "The host URL of your Databricks instance. Should start with https://",
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Uppercase: true,
+						Digits:    true,
+						Symbols:   true,
+					},
+					Prefix: "https://",
+				},
+			},
+			{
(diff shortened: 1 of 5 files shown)


AndyTitu:
@DCKcode @SimonBarendse for sec review.

plugins/databricks/personal_access_token.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func PersonalAccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.PersonalAccessToken,
+		DocsURL: sdk.URL("https://docs.databricks.com/dev-tools/auth.html"),
+		// The management URL is spesific to each deployment of Databricks, but this
+		// link describes how to manage tokens provided you are logged in.
+		ManagementURL: sdk.URL("https://docs.databricks.com/dev-tools/auth.html"),
```
SimonBarendse, line 20: ReadMe is in a simular situation where the management URL are specific/different for each user. We discussed how to handle this situation for the `ManagementURL` in this thread: https://github.com/1Password/shell-plugins/pull/106#discussion_r1049357649. I propose we do the same here:
- Don't set the optional `ManagementURL` field, the product will fall back to using the DocsURL with slightly different wording.
- Possible future extensions might support wildcards for the ManagementURL so it can support ReadMe, Databricks and others that have dynamic management URLs.

To help us improve & smoothen the contribution process, @bsamseth was it clear for you the ManagementURL field is optional?

SimonBarendse:
> I've opted for the "Person Access Token" credentials type, despite username+password being an option. The code here handles both cases, but the label of "Personal Access Token" isn't quite right in both cases. I still think it should be as I've proposed here to encourage the recommended flow, but an alternative would be to use a more generic "Configuration" label. A second opinion here would be nice :slightly_smiling_face:

I agree that the product can be opinionated towards the desired option. This also provides the best experience for the most common case of following Databricks recommended method of authentication (more closely matching wording).

I'd like to consider if we can even go one step further and choose to not support username+password at all. This will allow us to set the personal access token as required, which has the following advantages:
- `op plugin init` currently only prompts the user to set required fields. For this plugin, we'll definitely want to prompt for the personal access token during initialization.
- Simpler documentation, as we can omit username+password specifics. So the user does not need to consider which to use. They just have the single option.
- We help the user towards Databricks recommended form of authentication.

SimonBarendse (commented):
Thank you for an excellent contribution @bsamseth! Really appreciate your notes, makes it easier for us to think along and discuss. :raised_hands: 

I've also commented a small suggestion based on code review.

florisvdg:
I'd indeed remove the username and password fields and make the token field required. If we want to support both credential types, then we should create separate CredentialType structs for them.

plugins/databricks/personal_access_token.go
bsamseth: Got it, removed in 9d5f612b6127a9f1664ba054b86f1947d4fa917d. 

It was not immediately clear that it was optional from the template it self. However, a quick goto-definition on the field shows me a clear comment, so that's equally on me. It didn't occur to me to check though, so perhaps the template wording should/could be changed to make this more clear.

bsamseth:
I've removed username+password and made the token field required. Let me know if there's anything else you would want changed 👍🏻

DCKcode (approved):

SimonBarendse (approved):
Thanks again! :green_heart:

florisvdg (approved):

Issue #144, opened 2023-01-14
New plugin: Sourcegraph


arunsathiya:
### Platform or tool

Sourcegraph

### Relevant links

- https://docs.sourcegraph.com/cli/
- https://docs.sourcegraph.com/cli/quickstart
- https://docs.sourcegraph.com/cli/explanations/env

### Goal or desired behavior

Authenticate Sourcegraph CLI to a Sourcegraph instance.

### Contribution

- [X] I'd be up for building the plugin myself.

PR #146, opened 2023-01-17
Add support for Sourcegraph

arunsathiya:
This PR adds support for the Sourcegraph CLI. Fixes https://github.com/1Password/shell-plugins/issues/144.

Couple of notes:

- The CLI does not support config file imports. The CLI always expects the Sourcegraph instance host address and the access token to be set as environment variables.
- The `ManagementURL` isn't set for now because it's dynamic depending on the Sourcegraph instance host address **and** the username. In my case, it is https://sourcegraph.com/users/arun/settings/tokens. This case is similar to what happened with ReadMe, so based on [this advise](https://github.com/1Password/shell-plugins/pull/106#discussion_r1049357649) from @SimonBarendse, I haven't included `ManagementURL` field. When Shell Plugins support setting of dynamic `ManagementURL`s, we can update this plugin.

Here's a quick visualization of the plugin flow:

```
➜  shell-plugins git:(add/sourcegraph) src login

❌ Problem: No access token is configured.

🛠  To fix: Create an access token at https://sourcegraph.com/user/settings/tokens, then set the following environment variables:

   SRC_ENDPOINT=https://sourcegraph.com
   SRC_ACCESS_TOKEN=(the access token you just created)

   To verify that it's working, run this command again.

➜  shell-plugins git:(add/sourcegraph) op plugin init src
############################################################################
# WARNING: 'sourcegraph' is not from the official registry.                #
# Only proceed if you are the developer of 'sourcegraph'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/sourcegraph. #
############################################################################

Sourcegraph CLI [test build]
Authenticate with Sourcegraph Access Token.

? Locate your Sourcegraph Access Token: Sourcegraph Access Token (Personal)

? Configure when "Sourcegraph Access Token" will be used to authenticate: Prompt me for each new terminal session

The last step is to set up an alias for src.
You can do so by running the following command:

  source /Users/arun/.op/plugins.sh

Afterwards, run any src command to see it in action!
➜  shell-plugins git:(add/sourcegraph) source ~/.op/plugins.sh
➜  shell-plugins git:(add/sourcegraph) src login
############################################################################
# WARNING: 'sourcegraph' is not from the official registry.                #
# Only proceed if you are the developer of 'sourcegraph'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/sourcegraph. #
############################################################################

✔️  Authenticated as arun on https://sourcegraph.com
```

Happy to address any feedback, thanks!

### DIFF
--- a/plugins/sourcegraph/access_token.go
+++ b/plugins/sourcegraph/access_token.go
@@ -28,7 +28,7 @@ func AccessToken() schema.CredentialType {
 				Optional: true,
 			},
 			{
-				Name:                fieldname.HostAddress,
+				Name:                fieldname.Endpoint,
 				MarkdownDescription: "Base URL for your Sourcegraph instance.",
 				Secret:              false,
 				Composition: &schema.ValueComposition{
@@ -59,6 +59,6 @@ func AccessToken() schema.CredentialType {
 }
 
 var defaultEnvVarMapping = map[string]sdk.FieldName{
-	"SRC_ENDPOINT":     fieldname.HostAddress,
+	"SRC_ENDPOINT":     fieldname.Endpoint,
 	"SRC_ACCESS_TOKEN": fieldname.Token,
 }
--- a/plugins/sourcegraph/access_token.go
+++ b/plugins/sourcegraph/access_token.go
@@ -16,13 +16,13 @@ func AccessToken() schema.CredentialType {
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.Username,
-				MarkdownDescription: "Your Sourcegraph username.",
+				MarkdownDescription: "Username on your Sourcegraph instance.",
 				Secret:              false,
 				Composition: &schema.ValueComposition{
 					Charset: schema.Charset{
 						Lowercase: true,
 						Digits:    true,
-						Symbols:   true,
+						Specific:  []rune{'.', '-', '_'},
 					},
 				},
 				Optional: true,
--- a/plugins/sourcegraph/access_token.go
+++ b/plugins/sourcegraph/access_token.go
@@ -1,0 +1,64 @@
+package sourcegraph
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.AccessToken,
+		DocsURL: sdk.URL("https://docs.sourcegraph.com/cli"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Username,
+				MarkdownDescription: "Your Sourcegraph username.",
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+						Symbols:   true,
+					},
+				},
+				Optional: true,
+			},
+			{
+				Name:                fieldname.HostAddress,
+				MarkdownDescription: "Base URL for your Sourcegraph instance.",
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+						Symbols:   true,
+					},
+				},
+			},
+			{
+				Name:                fieldname.Token,
+				MarkdownDescription: "Token used to authenticate to Sourcegraph.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 40,
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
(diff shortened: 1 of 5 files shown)


plugins/sourcegraph/access_token.go
```diff
@@ -0,0 +1,64 @@
...
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Username,
+				MarkdownDescription: "Your Sourcegraph username.",
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+						Symbols:   true,
```
arunsathiya, line 25: Quick note that Sourcegraph definitely supports symbols in the username:

image

plugins/sourcegraph/access_token.go
```diff
@@ -0,0 +1,64 @@
...
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+						Specific:  []rune{'.', '-', '_'},
+					},
+				},
+				Optional: true,
+			},
```
hculea, line 29: Is there a reason for including the username here (even as optional), given the credential provisioner and importer do not provision/import this field at all?

hculea (commented):
Thank you for your contribution! I only left one question.

plugins/sourcegraph/access_token.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryEnvVarPair(defaultEnvVarMapping),
+		)}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"SRC_ENDPOINT":     fieldname.HostAddress,
```
SimonBarendse, line 62: Is there a reason to deviate the field name from the vocabulary Sourcegraph is using in the environment variable? i.e. a reason not to call the field `Endpoint`?

plugins/sourcegraph/access_token.go
```diff
@@ -0,0 +1,64 @@
...
+	"SRC_ACCESS_TOKEN": fieldname.Token,
```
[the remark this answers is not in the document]
florisvdg: We've been using `Token` as the field name elsewhere, as a shorthand to avoid too much duplication with the credential type name.

arunsathiya:
Thanks for reviewing, all of you! I have addressed your comments and welcome any other feedback.

hculea:
> Thanks for reviewing, all of you! I have addressed your comments and welcome any other feedback.

@arunsathiya I don't see any changes since last reviewed, is it possible that you pushed them on a different branch?

arunsathiya:
Sorry about the lack of clarity, @hculea. I meant that I have addressed your feedback verbally. The only pending change is renaming `HostAddress` field to `Endpoint` and I am waiting for a clarification on how to proceed with `AccessToken` vs `Token`. I wanted to address both suggestions in the code at the same time.

Anyway, I have updated to `Endpoint` for now.

hculea:
Thanks for clarifying. Please, have a look over the thread I started as well: https://github.com/1Password/shell-plugins/pull/146#discussion_r1071961934

plugins/sourcegraph/access_token.go
```diff
@@ -0,0 +1,63 @@
...
+				MarkdownDescription: "Username on your Sourcegraph instance.",
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+						Specific:  []rune{'.', '-', '_'},
+					},
+				},
+			},
```
hculea, line 28: Thanks for changing this, Arun 😄 However, I still think this needs some work.

Specifically, I would like to completely exclude the `username` field from this credential's definition.

This is because nor the provisioner neither the importer use this field, so functionally it serves no purpose.

That being said, there would be no problem if the user decides to add the username of their account within the item where the AccessToken and Endpoint are stored, but this should not be the responsibility of the Shell Plugin.

Let me know what you think.

plugins/sourcegraph/access_token.go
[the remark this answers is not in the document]
SimonBarendse: Okay, in that case this shouldn't be a blocker for this PR, and we can re-assess for all plugins together as an independent effort in case we want to change this.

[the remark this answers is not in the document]
SimonBarendse: i.e. we can leave as is: `Token` for now.

plugins/sourcegraph/src.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func SourcegraphCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Sourcegraph CLI",
+		Runs:      []string{"src"},
+		DocsURL:   sdk.URL("https://docs.sourcegraph.com/cli"),
+		NeedsAuth: needsauth.NotForHelpOrVersion(),
```
hculea, line 15: Should we opt out of authentication in the case where no arguments are provided, as well?

hculea (approved):

DCKcode (approved):

PR #149, opened 2023-01-17
Create RPC function response structs which are wrapping sdk outputs

AndyTitu:
RPC call responses are wrapping Provision, Deprovision and Import outputs, and are adding a Panic field meant to be set when panics arise in the RPC level. 

This is a better solution than https://github.com/1Password/shell-plugins/pull/121. It also requires protocol update to v4 hence it will get merged in `protocol-v4` to get shipped with more future similar changes.

This increases flexibility of handling these types of characteristic panics in the CLI.

See this dicussion for further details: https://github.com/1Password/shell-plugins/pull/121/files#r1055291768

### DIFF
--- a/sdk/rpc/proto/protocol.go
+++ b/sdk/rpc/proto/protocol.go
@@ -76,11 +76,21 @@ type ImportCredentialRequest struct {
 	sdk.ImportOutput
 }
 
+type ImportCredentialResponse struct {
+	*sdk.ImportOutput
+	Panic *Panic
+}
+
 // ProvisionCredentialRequest augments sdk.ProvisionInput with a CredentialID so Provision() can be called over RPC.
 type ProvisionCredentialRequest struct {
 	ProvisionerID
 	sdk.ProvisionInput
 	sdk.ProvisionOutput
+}
+
+type ProvisionCredentialResponse struct {
+	*sdk.ProvisionOutput
+	Panic *Panic
 }
 
 // DeprovisionCredentialRequest augments sdk.DeprovisionInput with a CredentialID so Deprovision() can be called over RPC.
@@ -90,9 +100,19 @@ type DeprovisionCredentialRequest struct {
 	sdk.DeprovisionOutput
 }
 
+type DeprovisionCredentialResponse struct {
+	*sdk.DeprovisionOutput
+	Panic *Panic
+}
+
 // ExecutableNeedsAuthRequest augments sdk.NeedsAuthenticationInput with the ID of an executable so NeedsAuth() can be
 // called over RPC. ExecutableID resembles the slice index of the executable in schema.Plugin.
 type ExecutableNeedsAuthRequest struct {
 	ExecutableID
 	sdk.NeedsAuthenticationInput
+}
+
+type Panic struct {
+	Error string
+	Stack []byte
 }
--- a/sdk/rpc/proto/protocol.go
+++ b/sdk/rpc/proto/protocol.go
@@ -113,6 +113,6 @@ type ExecutableNeedsAuthRequest struct {
 }
 
 type Panic struct {
-	Error string
-	Stack []byte
+	RecoveredString string
+	Stack           []byte
 }
--- a/sdk/rpc/proto/protocol.go
+++ b/sdk/rpc/proto/protocol.go
@@ -77,7 +77,7 @@ type ImportCredentialRequest struct {
 }
 
 type ImportCredentialResponse struct {
-	*sdk.ImportOutput
+	sdk.ImportOutput
 	Panic *Panic
 }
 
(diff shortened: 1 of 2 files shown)


sdk/rpc/server/server.go
```diff
@@ -116,16 +117,25 @@ func (t *RPCServer) ExecutableNeedsAuth(req proto.ExecutableNeedsAuthRequest, re
 
 // CredentialImport is a remote version of the Import() function in schema.CredentialType.
 // The call is forwarded to the Import() function of the credential identified by req.CredentialID.
-func (t *RPCServer) CredentialImport(req proto.ImportCredentialRequest, resp *sdk.ImportOutput) error {
+func (t *RPCServer) CredentialImport(req proto.ImportCredentialRequest, resp proto.ImportCredentialResponse) error {
+	defer func() {
+		if err := recover(); err != nil {
+			resp.Panic = &proto.Panic{
+				Error: fmt.Sprintf("your locally built plugin's importing failed with the following panic: %s", err),
```
jpcoenen, line 124: The fact that we're dealing with an ImportRequest should be clear from the response type. Similarly, the fact that this is set in a struct called `Panic` should indicate that this is about a panic.

So if we just set `Error: err.Error()` we provide the most flexibility for the consumer (CLI) of this struct. The CLI can add the extra context if it sees fit or leave it out if it is not needed.

sdk/rpc/server/server.go
```diff
@@ -116,16 +117,25 @@ func (t *RPCServer) ExecutableNeedsAuth(req proto.ExecutableNeedsAuthRequest, re
 
 // CredentialImport is a remote version of the Import() function in schema.CredentialType.
 // The call is forwarded to the Import() function of the credential identified by req.CredentialID.
-func (t *RPCServer) CredentialImport(req proto.ImportCredentialRequest, resp *sdk.ImportOutput) error {
+func (t *RPCServer) CredentialImport(req proto.ImportCredentialRequest, resp proto.ImportCredentialResponse) error {
```
jpcoenen, line 120: The `net/rpc` package [defines](https://pkg.go.dev/net/rpc) that the second argument (`resp`) has to be a pointer. The reason is that the second argument is used as a return value. If it's not a pointer, any changes to it will be lost after this function returns.

sdk/rpc/proto/protocol.go
```diff
@@ -76,23 +76,43 @@ type ImportCredentialRequest struct {
 	sdk.ImportOutput
 }
 
+type ImportCredentialResponse struct {
+	*sdk.ImportOutput
```
jpcoenen, line 80: The code might become a little easier if we drop this pointer. It's no longer needed if the methods of `RPCServer` start accepting a pointer value as their second argument (see other comment).

sdk/rpc/server/server.go
AndyTitu: Alright, makes sense. The only problem I still see here is that the `err`  we get from recover has type any, so it does not have any `.Error()` method, and casting to an error could sometimes fail because `recover()` does not always return an error. It was designed to also return the data-structure with which the panic has been initialized (`panic("oh- oh")`). Should we assume that in our case, we'll only catch panics from which we can recover errors, and cast?

sdk/rpc/server/server.go
AndyTitu: Done!

sdk/rpc/proto/protocol.go
AndyTitu: Done!

sdk/rpc/server/server.go
jpcoenen: Oh, yeah, you're right! I keep on forgetting that `panic()` accepts any value 😅 

Let's not cast the value; that could trigger another panic. What we could do instead, is rename the `Error` field. If it is instead called something like `Recovered` or `RecoveredString`, it would make more sense to just set it to `fmt.Sprintf("%s", r)` (where `r` is `recover()`'s return value).

sdk/rpc/proto/protocol.go
jpcoenen: It does not strictly _have_ to be a pointer, but it does more easily allow us to check if there was a panic (by checking if `Panic` is `nil`).

Alternatively, we can also add an extra field that indicates whether the call was successful or not, but a pointer should work just fine.

accraw (approved):

sdk/rpc/proto/protocol.go
AndyTitu: This makes sense, thanks!

sdk/rpc/server/server.go
AndyTitu: Makes sense, thanks,done!

jpcoenen (approved):

PR #152, opened 2023-01-18
Add plugin for OpenAI CLI

jodyheavener:
This PR adds a new plugin for the CLI in [OpenAI](https://openai.com/)'s Python [libary](https://pypi.org/project/openai/).

This CLI uses a single API Key environment variable, so the plugin simply provisions that.

## Testing steps

- [Install the OpenAI CLI](https://pypi.org/project/openai/)
- [Create an OpenAI account](https://beta.openai.com/signup)
- From your account, [create a new API Key](https://beta.openai.com/account/api-keys)
  - Either create a new vault item with this key, or manually create one during plugin init
- Checkout this branch: `git pull && git checkout jh/add-openai`
- Build the plugin `make openai/build`
- Init the plugin: `op plugin init openai`, source plugins
  - If you haven't created a vault item, this is where you'll manually create one with you API Key
- Run a command, for example: `openai api completions.create -e ada -p "Hello world"`

The command should successfully execute. Using the above example, the result would be:

CleanShot 2023-01-17 at 19 31 32@2x

## Example secrets:

```
API Key:
  API Key: sk-yEyY18xzH5IiiORdCDzstp1h2xrxCydfh9tjFveUyEXAMPLE
```

### DIFF
--- a/plugins/openai/api_key.go
+++ b/plugins/openai/api_key.go
@@ -1,0 +1,39 @@
+package openai
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://beta.openai.com/docs"),
+		ManagementURL: sdk.URL("https://beta.openai.com/account/api-keys"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "API Key used to authenticate to OpenAI.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 51,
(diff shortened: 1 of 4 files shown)


plugins/openai/api_key.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://beta.openai.com/docs"),
+		ManagementURL: sdk.URL("https://beta.openai.com/account/api-keys"),
```
hculea, line 16: Given these URLs start with `beta`, I'm assuming this product is still in beta, although I was not able to find a beta notice on the site.

I just wanted to see how exactly would a potential protocol change, on the side of a third party CLI, play in with our backwards compatibility promise. I'd reckon our promise does not extend to the third party tools we support through our plugins, but wanted to raise this just so we can make a conscious  decision on it. cc @SimonBarendse

hculea (approved):
Thank you for this addition! Looks good, code-wise, just a more conceptual question from my part.

plugins/openai/api_key.go
SimonBarendse: Great question!

We'll want to implement versioning to be able to handle backwards incompatible changes from the third-party CLIs. The plugin could behave differently depending on the version of the third-party CLI you have installed.

Issue #158, opened 2023-01-23
New plugin: Vultr


arunsathiya:
### Platform or tool

Vultr

### Relevant links

- Platform homepage: https://vultr.com/
- API docs: https://www.vultr.com/api/#section/Authentication
- API Key management: https://my.vultr.com/settings/#settingsapi
- CLI docs on GitHub: https://github.com/vultr/vultr-cli

### Goal or desired behavior

Authenticate to the Vultr platform.

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

This is fairly straightforward:

- Authentication is handled using environment variable `VULTR_API_KEY`
- If a `~/.vultr-cli.yaml` file exists, `api-key` value from that file is used. The config file can live anywhere, in which case config file is supplied to the command using `--config /path/to/file` option.
- If the config file is missing a `api-key` value, the CLI program defaults to using the value of the environment variable.

For the shell plugin:

- We can build support for importing the API Key from the config file.
- We can handle authentication/provision using environment variable.

PR #159, opened 2023-01-23
Add support for Vultr

arunsathiya:
Add support for Vultr. Fixes https://github.com/1Password/shell-plugins/issues/158.

- The shell plugin supports importing of API Key from `~/.vultr-cli.yaml` file.
- Provisions the credential as an environment variable `VULTR_API_KEY`

## Testing steps

- Sign up for a Vultr account. 
- Get an API Token from [this page](https://my.vultr.com/settings/#settingsapi). You'll need to set up billing to get a token though.
- Set up Vultr CLI using `vultr-cli account`, and create a file at `~/.vultr-cli.yaml` with the content: `api-key: key-goes-here`
- Checkout this branch and run `make vultr/build`
- Run `op plugin init vultr-cli`, and source shell plugin alias as well by running `source ~/.op/plugins.sh`
- Follow on-screen instructions to import your token from the config file.
- Restart shell and run `vultr-cli` and notice that the credential is now fetched from the 1Password item and provisioned as an environment variable.

## Example secret:

```
➜  shell-plugins git:(add/vultr) make vultr/example-secrets
go run cmd/contrib/main.go vultr/example-secrets
API Key:
  API Key: MH0SBC3BVWH27A00Q9CCCVKJKDUWCEXAMPLE
```

Thanks and feedback welcome!


plugins/vultr/vultr-cli.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func VultrCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Vultr CLI",
+		Runs:      []string{"vultr-cli"},
+		DocsURL:   sdk.URL("https://github.com/vultr/vultr-cli"),
+		NeedsAuth: needsauth.NotForHelpOrVersion(),
```
hculea, line 15: Let's opt out of authentication when `--config` is specified, since this indicated a custom credential being provisioned.

plugins/vultr/api_key.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://www.vultr.com/api/#section/Authentication"),
```
hculea, line 17: This link relates to the `vultr` API. From what I could see, the only documentation related to the CLI is within their repo's README (https://github.com/vultr/vultr-cli), and [this](https://www.vultr.com/news/How-to-Easily-Manage-Instances-with-Vultr-CLI/) article. Should we rather link to a section from one of these?

hculea (commented):
Looks great, thank you so much for this addition! ❤️ I've only left a couple of nitpicks.

plugins/vultr/api_key.go
arunsathiya: I arrived at that link from the link as follows: CLI documentation on GitHub > Vultr API Key management page > Vultr API docs:

Screenshot 2023-01-23 at 7 31 24 PM

Screenshot 2023-01-23 at 7 31 50 PM

Screenshot 2023-01-23 at 7 31 35 PM

Since `DocsURL` here refers about the credential itself and not about CLI docs, I am inclined to leave the current link as is. 🤔

plugins/vultr/vultr-cli.go
arunsathiya: That's definitely something I considered but for some reason assumed that users would use the same API Key both on environment variable and via config file. I have now updated the plugin to not use 1Password-stored API Key for `--config` commands.

arunsathiya:
Thanks for raising some great points, @hculea! I have addressed your feedback above.

plugins/vultr/vultr-cli.go
hculea: Thank you for addressing this. Could you add both these conditions to the authentication criteria? You can do this using `IfAll`. Something along the lines of:
```
NeedsAuth: needsauth.IfAll(
  needsauth.NotWhenContainsArgs("--config"),
)
```

Additionally, should this plugin prompt for authorization when no arguments are provided? i.e. only calling `vultr-cli`, with no other subcommands/flags/args.

arunsathiya: I feel silly removing the NotForHelpOrVersion() check. 😅 All changes should be addressed now: https://github.com/1Password/shell-plugins/pull/159/commits/acecc5d8ed77b3a456f4891e1ab945b8c7567086. 

Thanks for all the guidance, @hculea!

hculea:
Awesome! It now looks good to me, many thanks for this addition!

Code review ✅

hculea (approved):

SimonBarendse (approved):
Thank your for another great addition, Arun! :green_heart:

plugins/vultr/api_key.go
hculea: Thank you for expanding, Arun! 😄 That makes sense, I agree we should leave it like this.

Issue #154, opened 2023-01-19
New plugin: SnowSQL


williamhpark:
### Platform or tool

SnowSQL

### Relevant links

* General: https://docs.snowflake.com/en/user-guide/snowsql.html
* Getting Started with SnowSQL: https://quickstarts.snowflake.com/guide/getting_started_with_snowsql/index.html
* Installing SnowSQL: https://docs.snowflake.com/en/user-guide/snowsql-install-config.html
* Connecting through SnowSQL: https://docs.snowflake.com/en/user-guide/snowsql-start.html

### Goal or desired behavior

Authenticate the SnowSQL (Snowflake) CLI using credentials so Snowflake users can quickly authenticate into and control all aspects of their Snowflake Data Cloud.

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

After installing the SnowSQL CLI (I installed on macOS using [Homebrew](https://docs.snowflake.com/en/user-guide/snowsql-install-config.html#installing-snowsql-on-macos-using-homebrew-cask)), you sign-in to your Snowflake account using `snowsql` (you may include `-a` and `-u` flags for account and username). You then get prompted for your account, username, and password in the terminal.

It seems that there are [three ways](https://docs.snowflake.com/en/user-guide/snowsql-start.html#specifying-passwords-when-connecting) to specify credentials to authenticate:
1. Entered via [interactive prompt](https://docs.snowflake.com/en/user-guide/snowsql-start.html#label-snowsql-additional-connection-parameters-prompt) in SnowSQL (applies to passwords only).
2. Defined in the SnowSQL configuration file using the password option. For details, see [Configuring Default Connection Settings](https://docs.snowflake.com/en/user-guide/snowsql-start.html#configuring-default-connection-settings).
3. Specified using the `SNOWSQL_PWD` environment variables. For details, see [Using Environment Variables](https://docs.snowflake.com/en/user-guide/snowsql-start.html#using-environment-variables).

For development of this plugin, it seems that taking advantage of option 2 and the file provisioner may be the best option. The credentials can be stored in `~/.snowsql/config` then used to authenticate.

williamhpark:
**Update:** I ran into issues when trying to implement the plugin using file provisioning, and I realized that there are actually environment variables that can be used for the account and username too (`SNOWSQL_ACCOUNT` and `SNOWSQL_USER`, not just the password. I decided that using environment variables for the provisioner would be more straight-forward.

PR #161, opened 2023-01-23
New plugin: SnowSQL

williamhpark:
## Summary

Add support for a new plugin for the Snowflake CLI, SnowSQL.

Addresses: #154 

### Setup

1. Create a Snowflake account following the guide [here](https://quickstarts.snowflake.com/guide/getting_started_with_snowsql/index.html).
2. Build and test the SnowSQL plugin locally according to [these instructions](https://developer.1password.com/docs/cli/shell-plugins/contribute/). E.g. `make snowflake/build`

### Importer

1. After installing and setting up SnowSQL, a config file should have been created at `~/.snowsql/config`. Add your Snowflake credentials to this file under `[connections]`. Initialize the SnowSQL plugin and test the importer using this config file. The credentials should be saved in 1Password successfully. **Note:** By default, the `~/.snowsql/config` file has a `[connections]` section where the actual credentials are stored, as well as an uncommented `[connections.example]` section where example credentials are displayed. No matter where `[connections.example]` is located and even if the example credentials are uncommented, only the true credentials under `[connections]` should ever be saved by the importer.
2. Initialize the SnowSQL plugin again but this time, import credentials by entering the credentials manually in the terminal. The credentials should, again, be saved in 1Password successfully.

### Provisioner

1. After initializing the SnowSQL plugin and having your Snowflake credentials saved in 1Password, run a `snowsql` command (just `snowsql` should suffice, but if you want to add connection parameters the available ones are listed [here](https://docs.snowflake.com/en/user-guide/snowsql-start.html#connection-syntax)). You should be taken through the authentication flow with the SnowFlake Credentials that were saved in 1Password, and you shouldn't have to enter the credentials manually. Once authenticated, you should enter the SnowSQL shell, from which you can interact with your Snowflake warehouses/databases.

image

2. Try running a `snowsql` command with each of the following arguments (descriptions can be found [here](https://docs.snowflake.com/en/user-guide/snowsql-start.html#connection-syntax)):
```
-a
--accountname
-u
--username
--authenticator
--config
-P
--prompt
-v
--version
--help
```
You should not be prompted to authenticate through the 1Password CLI.

**Note**: While testing the provisioner, ensure that your credentials are not currently stored in `~/.snowsql/config` nor environment variables.

### Additional Information

- Be aware that the plugin name is `snowflake` and the executable name is `snowsql`.


plugins/snowflake/credentials.go
```diff
@@ -0,0 +1,108 @@
...
+		credentialsFile, err := contents.ToINI()
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		// The generated ~/.snowsql/config file, by default, includes an uncommented [connections.example] credential section below the actual
+		// [connections] section. We save the accountname, username, and password listed under [connections] then stop parsing the file
+		fields := make(map[sdk.FieldName]string)
+		for _, section := range credentialsFile.Sections() {
```
arunsathiya, line 86: I think we should rework this part because, for initial setup (that is, a config file without anything under `[connections]`) results in `[connections.example]`'s values getting added to 1Password:

image

We should prompt the user to enter the credentials manually instead.

Something like this should do the trick:

```suggestion
		for _, section := range credentialsFile.Sections() {
			if !section.HasKey("accountname") && !section.HasKey("username") && !section.HasKey("password") {
				return
			}
```

arunsathiya: Likewise, the parts below could be wrapped with a `section.Name() == "[connections]"`, I think. I haven't really tested that though but can give it a shot tomorrow morning:

```
if section.HasKey("accountname") && section.Key("accountname").Value() != "" {
	fields[fieldname.Account] = section.Key("accountname").Value()
}

if section.HasKey("username") && section.Key("username").Value() != "" {
	fields[fieldname.Username] = section.Key("username").Value()
}

if section.HasKey("password") && section.Key("password").Value() != "" {
	fields[fieldname.Password] = section.Key("password").Value()
}
```

arunsathiya (commented):
This is a great start! I left some comments to improve the importing flow for your consideration.

plugins/snowflake/credentials.go
```diff
@@ -0,0 +1,108 @@
...
+func TrySnowflakeConfigFile() sdk.Importer {
+	return importer.TryFile("~/.snowsql/config", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		credentialsFile, err := contents.ToINI()
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		// The generated ~/.snowsql/config file, by default, includes an uncommented [connections.example] credential section below the actual
+		// [connections] section. We save the accountname, username, and password listed under [connections] then stop parsing the file
```
SimonBarendse, line 84: Could you add a test case (with fixture) for this scenario (connections.example) as well?

plugins/snowflake/credentials.go
SimonBarendse: Great find, Arun! Thank you

SimonBarendse (dismissed):
Excellent! Snowflake will make for a great addition to the ecosystem!

Arun made some great suggestions in his review. I don't have much to add to that.

plugins/snowflake/credentials.go
williamhpark: Thanks for the suggestion Arun! Wrapping the conditionals with `if section.Name() == "connections"` as you suggested worked great and the importer behaviour works as expected. The `if !section.HasKey("accountname") && !section.HasKey("username") && !section.HasKey("password")` checks seem to be already covered by the conditionals I wrote so I don't think it's necessary to cover those again.

plugins/snowflake/snowsql.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func SnowflakeCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Snowflake",
+		Runs:      []string{"snowsql"},
+		DocsURL:   sdk.URL("https://docs.snowflake.com/en/user-guide/snowsql.html"),
+		NeedsAuth: needsauth.NotForHelpOrVersion(),
```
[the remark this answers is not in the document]
williamhpark: Good point! I modified the rules so that it opts out of 1Password authentication if any of the following arguments are passed:
```
-a
--accountname
-u
--username
--authenticator
--config
-P
--prompt
```

Descriptions for each of these arguments can be found here: https://docs.snowflake.com/en/user-guide/snowsql-start.html#connection-syntax. Let me know if there seem to be any arguments I missed, or if we shouldn't opt-out of authentication for any of the arguments I included.

plugins/snowflake/credentials.go
williamhpark: Just added two more tests! Please let me know if anything seems off.

williamhpark:
Updated the testing instructions based on recent feedback on the PR.

plugins/snowflake/snowsql.go
```diff
@@ -0,0 +1,32 @@
...
+	return schema.Executable{
+		Name:    "Snowflake",
+		Runs:    []string{"snowsql"},
+		DocsURL: sdk.URL("https://docs.snowflake.com/en/user-guide/snowsql.html"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWhenContainsArgs("-a"),
+			needsauth.NotWhenContainsArgs("--accountname"),
+			needsauth.NotWhenContainsArgs("-u"),
+			needsauth.NotWhenContainsArgs("--username"),
```
hculea, line 20: Is it possible that, if the command contains only one of these flags, the rest of the credentials could be retrieved from 1Password? 

In that sense, I'd like to only opt out of authentication when the command contains _all_ the necessary flags for specifying credentials. What do you think?

hculea: What would be really awesome, perhaps we can open a ticket about it: When the user specifies `--accountname`, the CLI automatically provisions the credential that has the `account` field set to that value. Likewise for the other two credentials.

hculea (commented):
Looks good! Only one comment, otherwise this is good to go.

plugins/snowflake/snowsql.go
SimonBarendse: I like that idea @hculea! Let's indeed open a follow-on ticket for that. While it's a great addition, there's already much value in including the Snowflake plugin before the credential selection through the `--account` flag is implemented.

plugins/snowflake/credentials.go
SimonBarendse: This looks great!

plugins/snowflake/credentials.go
SimonBarendse: Jup, with your updated changes, this looks right to me!

SimonBarendse (approved):

plugins/snowflake/snowsql.go
williamhpark: @hculea I like that suggestion as well! One thing I do want to mention is that there's no way to pass the password as an argument for the SnowSQL CLI, only account or username.

williamhpark: @hculea Would you be able to create the new follow-up issue? Or if you would prefer, I would be happy to write it up too.
For now, I'm marking this thread as resolved!

hculea: Thanks for expanding, William. In this case, what is the reason that we'd like to opt out of authentication when these flags are provided? The password would still need to be retrieved from 1Password.

hculea: Simon - absolutely, I wouldn't like to hold this back, my proposal is something we can do as an iterative improvement if anything. I think there are many plugins that could benefit from such a callback system 😄

DCKcode (dismissed):

plugins/snowflake/snowsql.go
AndyTitu: From my understanding, this situation falls under the question: "Should we allow users to bypass the shell plugin, if they want to specify some credentials manually?" @arunsathiya proposed this approach in this [comment](https://github.com/1Password/shell-plugins/pull/161#discussion_r1084899549).

The way I see it, here are the use cases:
1. User does not have all their accounts' credentials stored in 1P, and they would like to authenticate with another set of credentials, or they just want to quickly use some demo/test creds for some testing, where security is not a concern. 

> In this case, what is the reason that we'd like to opt out of authentication when these flags are provided? 

When one of `--accountname` or `--username` is provided, if we choose to bypass shell plugin auth, the default behaviour of `snowsql` would be to interactively prompt the user for specifying all the missing credentials (such as `password`). 

If we were not to bypass shell plugin auth, when someone would specify `--accountname`, for instance, the arg value will take precedence over our provisioned  `SNOWSQL_ACCOUNT` env var. However, the other secrets (`username` and `password`) will come from our `SNOWSQL_USER` and `SNOWSQL_PWD` and this will result in a credentials mismatch most likely. In this case, providing args in the command would prove breaking.

For the `--config` flag things are simple because this would point `snowsql` to take the creds straight from a config file, and regardless whether we enforce shell plugin auth or not, the config file passed through the flag will take priority over other credentials.

Therefore, I would say in this case we have 2 options between we should decide:
- we let users bypass shell plugin auth if they want to use other credentials, and we choose to remove shell plugin auth if any of the flags `--accountname`,`--username` or `--config` are present (as @williamhpark is currently doing)
- we direct users to import all their credential sets into 1P, and then we fallback on use case 2.

2. User has all their credentials in 1Password but would like to dynamically choose which one to use. 

Users could currently achieve this by re-doing the step of `op plugin init snowsql` and choose the other item where they stored their credentials. 

@hculea proposed a cooler alternative: matching the flags `--accountname`,`--username` in the command with the credentials configured to be working with the plugin, and failing if nothing matches.

AndyTitu: > Therefore, I would say in this case we have 2 options between which we should decide:

What do you think @SimonBarendse @hculea @williamhpark  and @arunsathiya ?

hculea: Hmm ok makes sense to let this be a way to opt out of the authentication.

We can afterwards iteratively add the dynamic matching of flags with 1password fields, and still support this behaviour, in case no creds are found. I'm okay to leave this as-is.

williamhpark: Thank you for laying all of that out @AndyTitu! Makes perfect sense to me.

I agree that matching the passed flags with 1Password fields would be really cool, and would probably be super convenient for certain developers who cycle between several accounts often.

plugins/snowflake/credentials.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func Credentials() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.Credentials,
+		DocsURL: sdk.URL("https://docs.snowflake.com/en/user-guide/snowsql.html"),
```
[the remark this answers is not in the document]
williamhpark: I can definitely make the docs URL more specific. However, since the URL you linked doesn't mention much about the account (e.g. where to find the account name), which is an important component of the credentials, I feel like this URL might be more appropriate: https://quickstarts.snowflake.com/guide/getting_started_with_snowsql/index.html.

Just want to make sure, are you suggesting that we add one of the suggested (e.g. `UserCredentials`, `UserPassword`, `UsernamePassword`, `UserLogin`) as an addition to the `credname` package and then use that new credname in this file? If this is the case, I feel like this may be better addressed in a follow-up issue, please let me know what you think.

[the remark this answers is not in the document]
hculea: Sounds good. With this in mind, I'll be resolving this thread. 

I'll make a note to open an internal ticket based on our discussion above.

plugins/snowflake/snowsql.go
hculea: Sounds good. With this in mind, I'll be resolving this thread.

I'll make a note to open an internal ticket based on our discussion above.

hculea (dismissed):

DCKcode (approved):

plugins/snowflake/credentials.go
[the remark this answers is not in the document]
florisvdg: I'm not sure from the top of my head if changing the credential name later will be a breaking change. I personally think that calling this credential type `credname.UsernamePassword` would make the most sense. I believe we don't have a constant for that yet in the `credname` package, but that's an easy change. As its string value, I'd go for `"Username/Password"`.

But open to other suggestions as well.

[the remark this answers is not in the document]
williamhpark: Since the credentials for this plugin actually contain three fields (account, username, password), I feel like just `"Username/Password"` may be a bit misleading. For this case, I feel like just `Credentials` or maybe `UserCredentials` would be best, since something like `AccountUsernamePassword` would be messy. However, I'm not sure what difference `Credentials` and `UserCredentials` would make, so I feel like keeping it as `Credentials` for now is probably fine? Let me know if you disagree.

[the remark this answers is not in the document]
florisvdg: "Username/Password" is the means of authentication here. Does not necessarily mean that the item only contains those fields.

You can think of the credential type name as an answer to "How would you like to authenticate?" or "Which kind of credential should be used to authenticate?". Answering that with just "Credentials" seems off there, because Snowflake supports multiple different credential types (key file, token, password).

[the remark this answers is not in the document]
AndyTitu: I am proposing "Username and Password" instead of the slash naming because it makes the change straightforward to introduce without changing the logic for creating files from the credentials name during the `make new-plugin` command (slashes are not allowed in file names). @williamhpark encountered some issue with this.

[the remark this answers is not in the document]
florisvdg: Probably a good idea anyway to filter out special chars when determining the file name (so in this case it could be `username_password.go`). I wouldn't let that be the reason to change it.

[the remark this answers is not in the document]
williamhpark: I decided that "User Credentials" may actually be a better choice than "Username and Password".

Although this is quite nitpicky, using "Username and Password" would result in the Snowflake credentials being imported as "Snowflake Username and Password" which isn't completely accurate and didn't look great to me.

Screenshot 2023-02-09 at 10 29 02 AM

"User Credentials" seems general enough to encompass credentials that require just a username and password, as well as credentials that require other fields to authenticate as well, like `account` in the case of this SnowSQL CLI. 

Screenshot 2023-02-09 at 10 33 21 AM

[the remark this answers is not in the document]
williamhpark: By the way I noticed that the credential types get published to the Go package here: https://pkg.go.dev/github.com/1Password/shell-plugins@v0.0.0-20230126100958-d4822d5ebb3d/sdk/schema/credname. If we make changes to the credential types, like I did in this PR, is there any additional action we need to take so that it gets reflected on the page I linked or would the changes get mirrored automatically once the PR is merged? cc: @SimonBarendse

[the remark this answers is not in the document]
AndyTitu: > is there any additional action we need to take so that it gets reflected on the page I linked or would the changes get mirrored automatically once the PR is merged?

I managed to reflect the [latest version of this repo on pkg.go.dev](https://pkg.go.dev/github.com/1Password/shell-plugins?tab=versions) by running:

` GOPROXY=proxy.golang.org go list -m github.com/1Password/shell-plugins@5c8ee89` where 5c8ee89 is the latest commit short hash, to obtain the canonical version: `github.com/1Password/shell-plugins v0.0.0-20230210102903-5c8ee89b020f`

Then I ran `curl https://sum.golang.org/lookup/github.com/1password/shell-plugins@v0.0.0-20230210102903-5c8ee89b020f` to force pkg.go.dev to obtain the hash for that particular version. If the version doesn't exist, it will fetch it - and then hash it.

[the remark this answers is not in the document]
AndyTitu: > `User Credentials` seems general enough to encompass credentials that require just a username and password, as well as credentials that require other fields to authenticate as well, like account in the case of this SnowSQL CLI. 

In this case, I would advise on `User Login Details` as it is more specific. It is a synonym to username +password but can also encompass additional credentials. I am agreeing with the fact that here we do not only deal with username and password but also the `account` field is required for auth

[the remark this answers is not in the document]
williamhpark: @AndyTitu Changed to `Login Details` per your suggestion.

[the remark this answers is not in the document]
AndyTitu: This is good with me!

sdk/schema/credname/names.go
```diff
@@ -13,8 +13,8 @@ const (
 	AuthToken           = sdk.CredentialName("Auth Token")
 	CLIToken            = sdk.CredentialName("CLI Token")
 	Credential          = sdk.CredentialName("Credential")
-	Credentials         = sdk.CredentialName("Credentials")
```
SimonBarendse, line 16: Was this previously unused? Because it looks like the `Credentials` type was there before this PR

SimonBarendse (approved):

sdk/schema/credname/names.go
williamhpark: Only references/uses of the constant seem to be in the file where it's defined, `sdk/schema/credname/names.go`. To double-check, a global search for `credname.Credentials` also gives no results.

image

SimonBarendse: Perfect!

Issue #156, opened 2023-01-22
New plugin: ngrok


arunsathiya:
### Platform or tool

ngrok

### Relevant links

- https://ngrok.com/ (Platform homepage)
- https://ngrok.com/docs/ngrok-agent/ngrok (CLI docs)
- https://dashboard.ngrok.com/get-started/your-authtoken (Authtoken -- it is called `Authtoken` instead of `Auth Token` for some reason -- to connect to the ngrok platform)
- https://dashboard.ngrok.com/api/ (API Key to connect to the ngrok API platform)

### Goal or desired behavior

Authenticate to the ngrok platform to serve two purposes:

- Securely tunnel traffic to the internet.
- Interact with the ngrok API.

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

I would like some advise on how to proceed there though as the `ngrok` CLI (called `ngrok Agent` as per [the docs here](https://ngrok.com/docs/ngrok-agent/)) serves two purposes: tunnel traffic to the internet **and** interact with the ngrok API.

- The first is done using their `Authtoken`: https://dashboard.ngrok.com/get-started/your-authtoken
- The second is done using the `API Key`: https://dashboard.ngrok.com/api/

Should I define two "credential names", or just one for `Authtoken` and define the other, `API Key`, as just a "field name"?

arunsathiya:
Just noticed that multiple credentials are not supported yet:

```
✘ Has no more than one credential type defined. Plugins with multiple credential types are not supported yet
✘ Has no more than one executable defined. Plugins with multiple executables are not supported yet
```

PR #165, opened 2023-01-25
Add support for ngrok

arunsathiya:
This PR adds support for [ngrok](https://ngrok.com/). Fixes https://github.com/1Password/shell-plugins/issues/156

## Testing instructions

- Sign up for ngrok.
- Get Auth Token from https://dashboard.ngrok.com/get-started/your-authtoken and an API token from https://dashboard.ngrok.com/api
- Checkout this git branch and run `make ngrok/build`
- Run `op plugin init ngrok`
- Since you do not have a ngrok config file at this point in time, you will be able to manually type your auth token and your API key.
- Ensure that the two are stored to a 1Password item.
- Also test environment variable-based import and config file imports.
- - For env var import test, first run `export NGROK_AUTHTOKEN=token` and `export NGROK_API_KEY=token` and then run `op plugin init ngrok`
- - For config file import test, save the below config file at `~/Library/Application Support/ngrok/ngrok.yml` if you are on Mac, or at `~/.config/ngrok/ngrok.yml` if you are on Linux, and then run `op plugin init ngrok`
- Make sure that `ngrok` without any commands or args does not prompt for 1Password auth.
- Make sure that `ngrok api tunnels list` and `ngrok http 80` work (first is API key test and the second is Auth Token test; for the second, you can see your account name on the generated tunnel)

## Sample config

```yaml
version: "2"
authtoken: uSuQ7LUOJLs4xRbIySZ15F4v5KxfTnMknMdFEXAMPLE
api_key: L4STpMP3K8FNaQjBo5EAsXA2SThzq0J7BKD3jUZgtEXAMPLE
```

## Notes

- The CLI supports two environment variables `NGROK_AUTHTOKEN` and `NGROK_API_KEY` but `ngrok api` commands fail even with the second env var set. I have asked about it here: https://github.com/inconshreveable/ngrok/issues/895
- For now, the shell plugin provisions creds using a temp config file. When the env var issue is resolved on ngrok side, we can provision using environment variables if that's preferred.
- "version" is a mandatory field in the config file. Provisioning fails without it. For some reason, that requirement is not mentioned here: https://ngrok.com/docs/ngrok-agent/ngrok

### DIFF
--- a/plugins/ngrok/credentials.go
+++ b/plugins/ngrok/credentials.go
@@ -45,17 +45,6 @@ func Credentials() schema.CredentialType {
 					},
 				},
 			},
-			{
-				Name:                fieldname.Version,
-				MarkdownDescription: "[Version of the ngrok config file](https://ngrok.com/docs/ngrok-agent/config#version). Required field for configuration file-based authentication.",
-				Optional:            false,
-				Composition: &schema.ValueComposition{
-					Length: 1,
-					Charset: schema.Charset{
-						Digits: true,
-					},
-				},
-			},
 		},
 		DefaultProvisioner: provision.TempFile(ngrokConfig, provision.Filename("config.yml"), provision.AddArgs("--config", "{{ .Path }}")),
 		Importer: importer.TryAll(
@@ -73,7 +62,7 @@ func ngrokConfig(in sdk.ProvisionInput) ([]byte, error) {
 	config := Config{
 		AuthToken: in.ItemFields[fieldname.AuthToken],
 		APIKey:    in.ItemFields[fieldname.APIKey],
-		Version:   in.ItemFields[fieldname.Version],
+		Version:   "2", // required field for ngrok CLI to work when file-based configuration is used; automatically configured by the CLI program and is not configurable by the user
 	}
 	contents, err := yaml.Marshal(&config)
 	if err != nil {
@@ -103,7 +92,6 @@ func TryngrokConfigFile(path string) sdk.Importer {
 			Fields: map[sdk.FieldName]string{
 				fieldname.AuthToken: config.AuthToken,
 				fieldname.APIKey:    config.APIKey,
-				fieldname.Version:   config.Version,
 			},
 		})
 	})
@@ -112,5 +100,5 @@ func TryngrokConfigFile(path string) sdk.Importer {
 type Config struct {
 	AuthToken string `yaml:"authtoken"`
 	APIKey    string `yaml:"api_key"`
-	Version   string `yaml:"version"`
+	Version   string
 }
--- a/plugins/ngrok/credentials_test.go
+++ b/plugins/ngrok/credentials_test.go
@@ -14,7 +14,6 @@ func TestCredentialsProvisioner(t *testing.T) {
 			ItemFields: map[sdk.FieldName]string{
 				fieldname.AuthToken: "uSuQ7LUOJLs4xRbIySZ15F4v5KxfTnMknMdFEXAMPLE",
 				fieldname.APIKey:    "L4STpMP3K8FNaQjBo5EAsXA2SThzq0J7BKD3jUZgtEXAMPLE",
-				fieldname.Version:   "2",
 			},
 			CommandLine: []string{"ngrok"},
 			ExpectedOutput: sdk.ProvisionOutput{
@@ -55,7 +54,6 @@ func TestCredentialsImporter(t *testing.T) {
 					Fields: map[sdk.FieldName]string{
 						fieldname.AuthToken: "uSuQ7LUOJLs4xRbIySZ15F4v5KxfTnMknMdFEXAMPLE",
 						fieldname.APIKey:    "L4STpMP3K8FNaQjBo5EAsXA2SThzq0J7BKD3jUZgtEXAMPLE",
-						fieldname.Version:   "2",
 					},
 				},
 			},
@@ -70,7 +68,6 @@ func TestCredentialsImporter(t *testing.T) {
 					Fields: map[sdk.FieldName]string{
 						fieldname.AuthToken: "uSuQ7LUOJLs4xRbIySZ15F4v5KxfTnMknMdFEXAMPLE",
 						fieldname.APIKey:    "L4STpMP3K8FNaQjBo5EAsXA2SThzq0J7BKD3jUZgtEXAMPLE",
-						fieldname.Version:   "2",
 					},
 				},
 			},
--- a/sdk/schema/fieldname/names.go
+++ b/sdk/schema/fieldname/names.go
@@ -43,7 +43,6 @@ const (
 	URL             = sdk.FieldName("URL")
 	User            = sdk.FieldName("User")
 	Username        = sdk.FieldName("Username")
-	Version         = sdk.FieldName("Version")
 	Website         = sdk.FieldName("Website")
 )
 
@@ -88,7 +87,6 @@ func ListAll() []sdk.FieldName {
 		URL,
 		User,
 		Username,
-		Version,
 		Website,
 	}
 }


arunsathiya:
This PR is now ready for review. ✅

plugins/ngrok/credentials.go
```diff
@@ -0,0 +1,102 @@
...
+
+func TryngrokConfigFile(path string) sdk.Importer {
+	return importer.TryFile(path, func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		var config Config
+		if err := contents.ToYAML(&config); err != nil {
+			out.AddError(err)
+			return
+		}
+
+		if config.AuthToken == "" {
```
hculea, line 85: Any reason to not include a similar check for `config.APIKey`?

plugins/ngrok/credentials.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return []byte(contents), nil
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"NGROK_AUTHTOKEN": fieldname.AuthToken,
+	"NGROK_API_KEY":   fieldname.APIKey,
+}
```
hculea, line 75: If this is a valid way to specify credentials, why not provision them as environment variables as well? This would get rid of having to somehow determine the config file version, and the workaround of needing to hardcode it in the Config passed to the process. 

Also, file provisioning comes with some design limitations related to the possibility of repeated reading, which is not a concern when using environment variables. Whenever possible, we should provision envvars.

plugins/ngrok/credentials.go
```diff
@@ -0,0 +1,102 @@
...
+				TryngrokConfigFile("~/.config/ngrok/ngrok.yml"),
+			),
+		)}
+}
+
+func ngrokConfig(in sdk.ProvisionInput) ([]byte, error) {
+	config := Config{
+		AuthToken: in.ItemFields[fieldname.AuthToken],
+		APIKey:    in.ItemFields[fieldname.APIKey],
+		Version:   "2", // required field for ngrok CLI to work
```
hculea, line 63: Have a look at my other comment, I would like to use an environment variable provisioner to get rid of needing to hardcode this. If we can make a good argument for keeping it here, I would much rather import it from the config file, whenever possible, given it is one of the fields there:
https://ngrok.com/docs/ngrok-agent/config#global-options

hculea (commented):
Thank you for this addition, @arunsathiya ! You're on fire with all these new plugins, absolutely love it ❤️ I've left a few comments below.

plugins/ngrok/credentials.go
arunsathiya: Thanks for the fast review, @hculea! I made a note on my PR but just to reiterate, the API Key environment variable isn’t working as expected. Seems to be an issue with ngrok CLI and I’ve asked about it here:

https://github.com/inconshreveable/ngrok/issues/895

That’s the only reason I chose file based provisioning.

plugins/ngrok/credentials.go
arunsathiya: Makes sense to import from the config file instead of hardcoding. I’ll update today morning the first thing.

plugins/ngrok/credentials.go
hculea: My apologies, seems like I did not read through with your MR description. Thanks for clarifying! I'm looking forward to see the response of folks from `ngrok`.

arunsathiya: No problem at all! I’m not sure if we’d hear back soon because some issues seem to indicate that issue tracker isn’t maintained. I’ll try asking on their slack but in the event we don’t hear soon, are we in agreement to go ahead with file provisioning approach?

plugins/ngrok/credentials.go
arunsathiya: I think I checked only for auth token because it’s the minimum required for most features while API key is only for API subcommand. 

I’ll take a closer look tomorrow morning as well.

plugins/ngrok/credentials.go
hculea: Sounds good! 

One more suggestion - I am not sure the repo where you have opened the issue is the currently maintained version of ngrok? (although it has 22.5k stars)

However, by their [changelogs](https://ngrok.com/docs/ngrok-agent/changelog#ngrok-agent-311---2023-01-13), it seems that the last released version was on 13th of January.

You might have more luck opening an issue on one of the repos within this org: https://github.com/ngrok

Hope this helps!

arunsathiya: I asked on one of `ngrok` organization's repository but no response:

- https://github.com/ngrok/ngrok-api-go/issues/20

I have given one last attempt [here on their Slack](https://ngrokcommunity.slack.com/archives/C033D4W1R6E/p1675671666308269).

I also had a look at `[ngrok-api-go](https://github.com/ngrok/ngrok-api-go/)` code but couldn't find any references to `NGROK_API_TOKEN` environment variable, which suggests that the [documentation here](https://ngrok.com/docs/ngrok-agent#supported-env-variables) may be outdated. I suggest we wait for one of ngrok's staff to respond to my Slack ping.

plugins/ngrok/credentials.go
arunsathiya: We now import the ngrok version from the config file instead of hardcoding it: https://github.com/1Password/shell-plugins/pull/165/commits/09b80ad8b86c3bdea3d220b21ebfe30acf42a3f4.

plugins/ngrok/credentials.go
arunsathiya: For the reasons that `ngrok api` commands need a valid API Key and version value to be present in the provisioning file, we now skip file import and prompt for manual input if API Key or Version are missing: 69ad00b327a6bf3489783999317a1f0442ec8d08.

arunsathiya:
Thanks for all the comments, @hculea. I have addressed your feedback above, but as a last step, [we are waiting for ngrok staff to respond to my Slack question](https://ngrokcommunity.slack.com/archives/C033D4W1R6E/p1675671666308269). I'll followup as soon as I hear from them.

plugins/ngrok/credentials.go
arunsathiya: Hi @hculea, I have some updates:

image

- ngrok staff confirmed ([on their Slack](https://ngrokcommunity.slack.com/archives/C033D4W1R6E/p1675725230866699?thread_ts=1675671666.308269&cid=C033D4W1R6E)) that environment variable isn't supported for API part of the ngrok agent for now. Auth Token (which is used for the rest of the ngrok agent features) is though.
- They removed it from the docs: https://github.com/ngrok/ngrok-docs/pull/188

At this point, my suggestion is to move forward with file-based provisioning, because at least that way, API key can be used. Once they announce enviroment variable-based authentication for API as well, we can our provisioning.

What do you think? Happy to take a different approach if you feel so. 

This is the last blocker for this PR to be merged.

arunsathiya:
Just so it isn't missed, [I shared the latest update here](https://github.com/1Password/shell-plugins/pull/165#discussion_r1099631826).

AndyTitu (commented):
From my understanding after reading `ngrok`'s documentation, it authenticates with either the auth token or the api key (they are not both required at the same time right?). 

It could be the case that they are both required, as I am not too familiar with `ngrok` documentation. In this case, the current solution is definitely the way to go for now, since the env vars are out of the question because the api key is not covered.

In the case where they are not both required at the same time, I'm understanding that your reason for going for the config file provisioner is to allow users to choose between the 2 different authentication formulas, since the config file supports both while env vars do not.

- However, the structure of the current plugin is in such a way that users cannot really choose between them, but they should rather have both, because these fields are both denoted with `Optional: false`. 

- If you want users to be able to choose between both of them you would have to specify `Optional:true` for both. Then this would also not be right because at least 1 is required. 

To avoid these complications I would suggest just going ahead with the env var approach, and only support the auth token. In this way we also avoid storing the non-secret version in 1Password.

arunsathiya:
Thanks for your detailed review, @AndyTitu! Regarding the Authtoken vs API Key confusion, all commands require the Authtoken with the exception of `ngrok api` and its subcommands. It's not clearly mention on the CLI docs, but this gives me that impression:

https://ngrok.com/docs/ngrok-agent/ngrok#ngrok-api

image

I have also tested running `ngrok api` subcommands with just the Authtoken, and they fail. It's only after provisioning a API key as well, the commands work.

Having said that, the ngrok team fixed envvar support for API key as of yesterday. It's just a matter of waiting for the update to be available to the general public, which I don't have a timeline on:

image

At this point, we could wait for that updated version to be available and then provision credentials as envvars, or if we want to get this out right now, we can:

- retain the current version of supporting both Authtoken and API key, 
- have both as required fields, because without the API key stored in 1Password, all `ngrok api` subcommands will fail,
- and, switch to envvar-based approach when the new ngrok client update is out. 

>In this way we also avoid storing the non-secret version in 1Password.

Could you elaborate on this part, please? Did you mean to say that API Token is not considered a secret credential?

plugins/ngrok/credentials.go
```diff
@@ -0,0 +1,113 @@
...
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+			{
+				Name:                fieldname.Version,
+				MarkdownDescription: "API Key used to authenticate to ngrok API.",
```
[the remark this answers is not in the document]
arunsathiya: Good spot, I'll update this shortly but I have just realized that there's a larger blocker in terms of this field. If the user is importing the configuration file from the computer, the version gets imported fine, but if the user chooses to manually type the field values, it is possible for them to type a value different from "2". 

ngrok agent/CLI expects the config file version to be exactly "2".

So, I am starting to think that it makes more sense to just wait for the next ngrok CLI/agent update to be available and then provision as just envvars. There wouldn't be a need to touch the version field name at all in that case.

What do you think, @AndyTitu?

AndyTitu:
Thanks for the additional information @arunsathiya ! In my opinion, the best way ahead given our current situation would be to go ahead with the current state of this PR, with only a small change added. To clarify, this is what I'm looking for (mostly we should already be there):
- use the config file provisioner where both auth token and api key are provisioned together, alongside the version, as we are doing now
- keep the version as it is now (not hardcoded, and imported from a config file/manually added into 1Password) - even though a user mistake would cause the plugin to crash, my back-of-a-napkin estimate is that in 90% of the cases users will know what the version should be
- make the api token `optional`. Most users will have an auth token, and most of the cases could be handled by only provisioning an auth token. For these users, the API key is therefore optional. For users who do use an api, they will need to take the extra step of creating and adding the API key into 1Password. 

This will avoid having to wait on 3rd party releases. When their release is out we could file a change for moving over to env vars to get rid of the need to specify the version.

plugins/ngrok/credentials.go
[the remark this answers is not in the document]
AndyTitu: Thanks! My answer is in here: https://github.com/1Password/shell-plugins/pull/165#issuecomment-1427661811

plugins/ngrok/credentials.go
AndyTitu: I responded to this on a wider note here: https://github.com/1Password/shell-plugins/pull/165#issuecomment-1427661811

SimonBarendse:
I have opened https://github.com/1Password/shell-plugins/issues/182 for us to look into supporting different credential types for different commands. In the meantime, we can proceed with either having support for just the auth token and no support for the `ngrok api` command yet, or we can proceed with accepting the API key as optional field. I'm okay with either approach.

arunsathiya:
Thanks @AndyTitu and @SimonBarendse, I am on board with this approach as well. This PR is ready for one final look. 🤞🏼 But, the test is failing with undefined credname.Credentials. I see it okay here though:

https://github.com/1Password/shell-plugins/blob/b957d2877a71a3d299e3e09271cf75d328e6f576/sdk/schema/credname/names.go

Is there a way to re-run tests without having to push a new commit?

SimonBarendse:
I think we need to re-add the credential type, we removed it assuming it wasn't used here: https://github.com/1Password/shell-plugins/pull/161#discussion_r1104723327

SimonBarendse:
@arunsathiya I've re-opened an old discussion here: https://github.com/1Password/shell-plugins/pull/165#discussion_r1105651772

arunsathiya:
>We need to re-add the credential type, we removed it assuming it wasn't used here: https://github.com/1Password/shell-plugins/pull/161#discussion_r1104723327
Could you pull in main to your branch, and then make sure it's added back in?

Thanks for the pointer, Simon. Rebased and tests now pass. ✅ 

Agree on hardcoding the version number; precisely why I had set that up as such in the first place.

plugins/ngrok/credentials.go
arunsathiya: No problem about the back and forth. Always helps to evaluate all options for users' best experience. :) With this comment we should be good to go: https://github.com/1Password/shell-plugins/pull/165#issuecomment-1430525243.

hculea: This will likely not be an issue anymore, if we decide to wait and go with envvars, which is my preferred approach 😄 

However, I do not agree that we should have this hard-coded. This does not take into consideration future possible version bumps in the config file format that `ngrok` may perform. 

That being said, we might reach the situation where the importer imports credentials from a config file with version "3", but provisions it to be "2", which is not guaranteed to work.

I don't think that we should be imposing more restrictions than the underlying CLI does, for its own configuration. In the case of this config's file version, this is something that the user has access to as well. Seems unnecessary to me to guard that, when the call to the 3rd party CLI is done through a plugin.

I'd love to hear what you both think!

hculea: I guess one could argue that if the config file version is bumped, we should make a conscious choice to bump it plugin-side as well, to make sure what we provision is indeed compliant with the new format.

If this is indeed how we want to ration this, I think it would also be wise to only import from a config file if the format version is "2" as well, since otherwise the current parsing logic may not produce reliable results.

hculea: Once again, I'd be in favour for waiting and using an env-var provisioner. But I think this idea could work as well, if we're really eager to push this out 😄 Thanks for the back and forth!

SimonBarendse: > I guess one could argue that if the config file version is bumped, we should make a conscious choice to bump it plugin-side as well, to make sure what we provision is indeed compliant with the new format.

:100: this. The more complexity we can absorb in the plugin itself, the better experience we all have using the plugin. If something changes in the ngrok configfile format/version, only a single user of the plugin has to make the change in the plugin, and all of us will get the new and improved behavior. If we push setting/maitaining the version number to every single user by moving it into the 1Password item of all, everyone needs to maintain and update it separately in the case of a change.

If the way to pass a credential would differ between two major versions of a tool (e.g. ngrok), we can even add version detection in the plugin ecosystem, so that the behavior of the plugin is dependent on the version of the tool you have installed, so that it continues to work and you can switch between the versions seamlessly. We can tackle that implementation if and when needed, that should be relatively straightforward to add.

SimonBarendse: Switching to an envvar provisioner is a non-breaking change, so we can always consider that in a next iteration of this plugin. The current implementation is already super valuable for users, so let's get it out there :rocket:

plugins/ngrok/credentials.go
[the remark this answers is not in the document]
SimonBarendse: Resolving this since there's no version field on the 1Password item anymore

SimonBarendse (approved):
Love your work on this @arunsathiya! Thank you for figuring out the differences between Auth Token and API Key and a way to move forward with this. The plugin is in a great spot! Looking forward to release this one soon :rocket:

Once we've tackled https://github.com/1Password/shell-plugins/issues/182, we can re-visit this plugin to polish it further with first-class support for the `ngrok api` commands.

Thanks again! :raised_hands:

hculea (approved):

plugins/ngrok/credentials.go
arunsathiya: Since the ngrok team already has a fix for API Token's envvar, we can push that update fairly soon. It's just a matter of few more days/weeks for the ngrok update to be available on the package managers they support. 🤞🏼

arunsathiya:
@SimonBarendse @hculea A quick note that I had previously incorrectly understood that `NotForExactArgs()` check meant both the command and its subcommands, which resulted in this breakage:

- https://github.com/1Password/shell-plugins/issues/185

I was testing only `ngrok update` which had given me that impression. 

I now have a fix for it here:

- https://github.com/1Password/shell-plugins/pull/187

Would we be able to release a patch build? My apologies for the hassle here.

PR #168, opened 2023-01-28
New plugin: npm install / publish

russell-accurx:
## Summary

A plugin that adds support for `npm install` and `npm publish` from and to private `npm` repositories. These are the only commands (to my knowledge) that need authentication, so I've also added a new `needsauth` helper to support this case. 

This works for my needs but I think I might need some advice on how best to make it into a general purpose plugin (see *Help needed* below). Also NB. I have not written a single line of Go before! The docs + generators made it v easy to get something working - really great :-D.

## Testing

Create a npm package repo:

```
git init test-project
cd test-project
npm init
```

Set the package `name` to `@YOUR_GITHUB_USERNAME/test-project` and the `repository` to `https://github.com/YOUR_GITHUB_USERNAME/test-project` and make sure both your github username and the project name are all in lower case - this is important for github packages to work!

Add an `.npmrc` file to your package repository containing the following:

```
@YOUR_GITHUB_USERNAME:registry=https://npm.pkg.github.com/YOUR_GITHUB_USERNAME
//npm.pkg.github.com/:_authToken=${NPM_TOKEN}
```

Create a repo on github for your repository and push your code to it.

Build and initialize your plugin as described in [these docs](https://developer.1password.com/docs/cli/shell-plugins/contribute/). When prompted for a personal access token, create one from your Gthib account as [described here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token). It needs the `write:packages` permission only.

Back in the `test-project` directory, run `npm publish`. This command should succeed and you will see your package appear in the right-hand column of your Github repo / in Github packages:

image

To test `npm install`, create the same `.npmrc` file in any other npm project and run `npm install @YOUR_GITHUB_USERNAME/test-project` - it should succeed.

You can see the test project I used to write these test instructions [here](https://github.com/russell-accurx/test-project).

## Help needed

As mentioned above, this works for my use case: installing and publishing packages to a single private repo using Github packages. There are some more general cases however that it might be good to support, but I'm not sure how:

1. Other private npm repositories are available - however there's only space for one set of docs links / instructions links for e.g. creating the personal access token. Examples of other private repositories are npmjs.org itself and azure artifacts.
2. It's possible to have multiple private repositories configured in your `.npmrc`, one for each `@scope` (the prefix of the package name). Not sure how or if it's possible to support this.
3. This plugin injects the auth token as the environment variable `NPM_TOKEN`, but that name isn't fixed - you can use any environment variable name you like, as long as it matches what you put in `.npmrc`.

PS. Having just read the contributor guidelines, I'm going to set up a GPG token now so I can sign my commit. BRB!


plugins/npm/npm.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func NPMCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "NPM install/publish",
+		Runs:      []string{"npm"},
+		DocsURL:   sdk.URL("https://docs.npmjs.com/cli"),
+		NeedsAuth: needsauth.IfAny(
+			needsauth.WhenContainsArgs("i"),
+			needsauth.WhenContainsArgs("install"),
```
azu, line 17: I don't know if it should be supported, but npm install has a lot of different alias.

> npm install [<package-spec> ...]
> aliases: add, i, in, ins, inst, insta, instal, isnt, isnta, isntal, isntall
> https://docs.npmjs.com/cli/v9/commands/npm-install

plugins/npm/npm.go
```diff
@@ -0,0 +1,29 @@
...
+			needsauth.WhenContainsArgs("ci"),
```
azu, line 18: Same here

> npm ci
> aliases: clean-install, ic, install-clean, isntall-clean
> https://docs.npmjs.com/cli/v9/commands/npm-ci

plugins/npm/npm.go
AndyTitu: Yes, I would say all aliases should be included here, otherwise the plugin's security guarantees could simply be bypassed by using an alias.

AndyTitu:
> Other private npm repositories are available - however there's only space for one set of docs links / instructions links for e.g. creating the personal access token. Examples of other private repositories are npmjs.org itself and azure artifacts.

This could be solutioned in the future by creating multiple `CredentialType`s and appending them to 
```
Credentials: []schema.CredentialType{
 			PersonalAccessToken(),
 		},
```
Unfortunately we do not support multiple `CredentialType`s for now. But this is definitely something we are working towards as a solution to this type of use cases.

> It's possible to have multiple private repositories configured in your .npmrc, one for each @scope (the prefix of the package name). Not sure how or if it's possible to support this.

This seems similar to `aws` assumeRole use case. https://github.com/1Password/shell-plugins/pull/92. We are looking into user friendly solution for this, and your input about the UX would be greatly appreciated. How would you like to be able to specify the scope of a `npm publish` command for example? 

For now this specific use case could be handled by setting up all your scoped in the `.npmrc`:
```
@SCOPE1: ...... NPM_TOKEN_SCOPE1
@SCOPE2: ..... NPM_TOKEN_SCOPE2
``` 
file and setting up all secrets in 1Password as well as setting up the proper env var mapping in the plugin:
```
"NPM_TOKEN_SCOPE1": fieldname.Token + "1"
"NPM_TOKEN_SCOPE2": fieldname.Token + "2"
```

Then when you execute `npm publish --scope=@scope2`, all secrets get provisioned the right one gets selected. This is totally not the most elegant and a bit of too much setup to be done so we are definitely looking into building some better support for this type of use case.

> This plugin injects the auth token as the environment variable NPM_TOKEN, but that name isn't fixed - you can use any environment variable name you like, as long as it matches what you put in .npmrc.

The name of the env var is pretty static as it gets specified only once in the definition of the plugin and cannot be changed unless the plugin itself changes. So this will require some user-side effort to make sure the env vars names are matching.

AndyTitu:
Btw just wanted to mention that this plugin rocks! Thanks for contributing @russell-accurx ! 

I just realised that the approach we usually encourage for package managers is to first allow the official registry, so it would be a bit more natural to support the NPM registry with this plugin first before we move on to other registries, including the Github packages registry, when multiple credentials will be supported per plugin. Do you think we could change this PR to reflect that?

russell-accurx:
thanks @AndyTitu ! yes that sounds sensible. i'll update it soon when i have time - hopefully in the next week or so.

fwiw i solved my immediate problem by creating a simple bash (well, fish) wrapper script for `npm` that pulls the token from the mac keychain and sets it as an env var. i initially tried using `op run` for this, but it ended up being too slow to be usable - perhaps because it checks all environment variables to see if any need to be replaced? i couldn't tell

plugins/npm/npm.go
hculea: >  the plugin's security guarantees could simply be bypassed by using an alias

Could you expand on that? I'm not sure I understand why that would be the case.

hculea:
Hey @russell-accurx!!

I'm just now reading through the threads. An NPM plugin would be an awesome addition to the ecosystem, thank you for you contribution!! ❤️ 

I only have a few comments: first thing that pops is that, in its current shape, this plugin would still require the user to perform an extra manual step, in order to replace their NPM token with a reference to the NPM_TOKEN envvar in their .npmrc. With Shell Plugins, we're trying to come up with a setup solution that doesn't require any additional steps, and is similar across different plugins, so I'd love to explore getting rid of that. We can definitely work together on that 😄 

Secondly, as Andi mentioned, it would be more natural to start by adding support for the official NPM registry first. Now, I realise that this may be straying away from solving your problem and addressing your use-case. I've opened an issue about supporting multiple backends https://github.com/1Password/shell-plugins/issues/192 to ensure that we're still moving in a direction that will ultimately be helpful for you. With this in mind, if you want to continue working on this MR, despite the lack of immediate applicability for your use-case, that would be awesome, and we'd be delighted to help you move this forward. Otherwise, that's completely okay, I'm confident that someone else from the community would pick this up 🚀

ghengeveld:
Would be amazing if this would handle the OTP (2nd factor) as well 🙌

Issue #133, opened 2023-01-02
Fastly cli


shyim:
### Platform or tool

Fastly

### Relevant links

* https://fastly.com

### Goal or desired behavior

Personal Access Token

### Contribution

- [X] I'd be up for building the plugin myself.

arunsathiya:
I started a PR here:

- https://github.com/1Password/shell-plugins/pull/169

Hope that's okay with you, @shyim, and I welcome any feedback you may have!

Issue #130, opened 2022-12-25
Easily switch between multiple credentials for a single CLI tool.


skyzyx:
### op CLI version

2.12.0

### Goal or desired behavior

There are several cases where I manage different credential sets for the same CLI command.

For example, I manage over 180 AWS accounts (I use AWS Vault for this and am not planning on switching, but it makes for a good example.)

I have read-only + admin tokens for both GitHub.com as well as multiple GitHub Enterprise users. If I'm using the `gh` command, switching between them should be easy.

### Current behavior

Unknown. I can't tell if/how this is supported.

### Relevant log output

```shell
N/A
```

arunsathiya:
Just wanted to share my interest for this ability as well. Right now, the only way to switch to another credential is to first run `op plugin clear plugin-name` followed by any command for that plugin, thus being prompted to choose the credential again.

I ran into the need for this ability when building the Fastly shell plugin:

- https://github.com/1Password/shell-plugins/pull/169

A variant of the `[needsauth.Always()](https://github.com/1Password/shell-plugins/blob/d4822d5ebb3ddf2d633316e1a2f06852bebccaf5/sdk/needsauth/helpers.go#L55-L59)` helper function would be nice. For example, in the Fastly CLI case, any command starting with `fastly profile` could prompt the user to choose the credential before executing the rest of the command. Example code:

```golang
NeedsAuth: needsauth.IfAll(
			needsauth.Always(
				needsauth.ForCommand("profile"),
			),
		),
```

Lugoues:
The ability to specify a secret name directly on the cli when executing would be amazing.
`op plugin run --secret "AWS\my-account" -- aws`

chrislambe:
I would love to have this for picking a database environment (e.g. `local`, `dev`, `staging`, etc.) when using `pgcli`. Maybe if the plugin has been `init`'d with multiple credentials it can prompt to choose one instead of just using the most recently selected credential?

scottisloud:
For those still watching this issue: 
Out of curiosity, when you set up a new plugin, you're asked if you want:
* To be asked to pick a credential every terminal session
* To use the selected credential only in this directory
* To be used globally

Do the first two options provide any help in these situation? For example, different repos could have different credentials for the same CLI using the "only in this directory".

scottisloud:
[...]

Issue #153, opened 2023-01-19
Add builtin support for XDG_CONFIG_HOME


hculea:
### What does it concern?

The plugin SDK, design, or schema

### Goal or desired behavior

Many plugins use `$XDG_CONFIG_HOME` as default location for their configuration file, defaulting to `$HOME/.config` if this is not defined.

The plugin system should have support for this, such that plugins can only define one path in their importers.

This has already been explored here: 
https://github.com/1Password/shell-plugins/commit/d44a749ba7d89e729a7a11366af6c3ac67cfd943

### Current behavior

Plugins need to define importers for two paths: `$XDG_CONFIG_HOME/...` and `$HOME/.config/...`.

PR #169, opened 2023-01-31
Add support for Fastly with envvar provisioning and config file imports

arunsathiya:
This PR adds support for [Fastly CLI](https://github.com/fastly/cli/). Fixes https://github.com/1Password/shell-plugins/issues/133.

## Key things to know:

- Fastly uses `API Token`s to authenticate.
- `FASTLY_API_TOKEN` envvar (environment variable) is the easiest way to authenticate. The CLI also supports `-t`, `--token` as command flags to login, but environment variables are preferred.

### Issue with profile subcommands

`fastly profile` command has a bunch of sub-commands `create`, `delete`, `list`, `switch`, `token` and `update`. Of these commands, except for `list` subcommand, all subcommands can be used in the form of the same environment variable (_I think_), but that's not possible just yet because Shell Plugins does not offer a simple way to switch to another credential:

- https://github.com/1Password/shell-plugins/issues/130

Until that feature is available, I am suggesting that we opt out of 1Password authentication for `fastly profile` commands.

The other command, `fastly profile list` basically looks for the list of available profiles in the config file and operate on the data available there. With Shell Plugins though, the config file will basically be void of these profiles, which in turn means that that command will just fail.

Overall, to me, it makes sense to avoid 1Password authentication for `fastly profile` commands for now. 

### Issue with config file imports

For some reason, I am unable to get MacOnly and LinuxOnly OS-specific config file imports working. It's probably something simple that I am missing. cc @hculea as we previously scheduled a call to discuss this. 🙏🏼 Thank you for any insights!

### DIFF
--- a/plugins/fastly/api_token.go
+++ b/plugins/fastly/api_token.go
@@ -1,0 +1,77 @@
+package fastly
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIToken,
+		DocsURL:       sdk.URL("https://docs.fastly.com/en/guides/using-api-tokens"),
+		ManagementURL: sdk.URL("https://manage.fastly.com/account/personal/tokens"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Token,
+				MarkdownDescription: "API Token used to authenticate to Fastly.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 32,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryAllEnvVars(fieldname.Token, "FASTLY_API_TOKEN"),
+			importer.MacOnly(
+				TryFastlyConfigFile("~/Library/Application Support/fastly/config.toml"),
+			),
+			importer.LinuxOnly(
+				TryFastlyConfigFile("~/.config/fastly/config.toml"),
+			),
+		),
+	}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"FASTLY_API_TOKEN": fieldname.Token,
+}
+
+func TryFastlyConfigFile(path string) sdk.Importer {
+	return importer.TryFile(path, func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		var config Config
+		if err := contents.ToTOML(&config); err != nil {
+			out.AddError(err)
+			return
+		}
+
+		for profileName, configProfile := range config.Profile {
+			out.AddCandidate(sdk.ImportCandidate{
+				Fields: map[sdk.FieldName]string{
+					fieldname.Token: configProfile.Token,
+				},
+				NameHint: importer.SanitizeNameHint(profileName),
+			})
+		}
+	})
+}
+
+type ConfigProfile struct {
+	Email string `toml:"email"`
+	Token string `toml:"token"`
+}
+
+type Config struct {
+	Profile map[string]ConfigProfile
+}
--- a/plugins/fastly/api_token_test.go
+++ b/plugins/fastly/api_token_test.go
@@ -1,0 +1,93 @@
+package fastly
+
+import (
+	"testing"
+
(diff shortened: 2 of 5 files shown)


shyim (approved):
wow thanks ❤️ . Looks really good!

Integralist:
👋🏻 @arunsathiya 

If there are any issues with the Fastly CLI that you think might make integration with 1Password easier, then do please open an issue at https://github.com/fastly/cli

arunsathiya:
Converting to draft for now because I need help with OS-specific importers.

hculea:
Thank you for this addition, Arun! ❤️ I've looked over the importers and both look good to me. The paths also look to be correct - user config dir suffixed by `fastly/config.toml`.

Off the top of my head, I could see a possible problem on UNIX systems (aside from MacOS), in the situation where the user would have the `XDG_CONFIG_HOME` envvar pointing to a custom location.

Is it possible that this is the case, in your situation? If not, could you maybe describe what is the unexpected behaviour that you're encountering, when trying to run the file importers? Thanks!

arunsathiya:
Thanks for taking a look, @hculea. `XDG_CONFIG_HOME` is empty for me:

```
➜  ~ echo $XDG_CONFIG_HOME

```

However, I have a positive news. Mac OS config file import is working as expected now. I am not sure what I was doing incorrectly the last week. Perhaps I didn't have any configuration profiles on the `/Users/arun/Library/Application\ Support/fastly/config.toml` file. 🤷🏼 

I tried testing it on Linux (Ubuntu) on Docker, but it seems that Shell Plugins aren't supported on it:

```
root@c4fc0e1feba7:~/.config/op/plugins/local# ls
fastly
root@c4fc0e1feba7:~/.config/op/plugins/local# cd ~
root@c4fc0e1feba7:~# op plugin init fastly
WARNING:  is not a shell officially supported by 1Password Shell Plugins!
[ERROR] 2023/02/08 04:34:49 unknown plugin: fastly
```

I am not too sure how to move forward in terms of Linux testing, but given that the path is correct, we could assume that it works fine on Linux.

At this point in time, this PR is ready for a review! 😄

arunsathiya:
Side note: 

I generated the fastly shell plugin build on my Mac and copied to the Docker container using `docker cp ~/.op/plugins/local/fastly c4fc0e1feba7:/root/.config/op/plugins/local`, but I don't know if that's a sufficient way to do it or if the build for Linux will completely vary.

AndyTitu:
I think you should copy the config to the `.op` file location. This is globally used by locally built shell plugins: `docker cp ~/.op/plugins/local/fastly c4fc0e1feba7:/root/.op/plugins/local` @arunsathiya This should solve the above issue.

plugins/fastly/api_token.go
```diff
@@ -0,0 +1,77 @@
...
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryAllEnvVars(fieldname.Token, "FASTLY_API_TOKEN"),
+			importer.MacOnly(
+				TryFastlyConfigFile("~/Library/Application Support/fastly/config.toml"),
```
AndyTitu, line 38: if fastly's config file location depends on `XDG_CONFIG_HOME` I would first check here if that env var is set to anything, and if yes, use that location, otherwise use the already existing logic.

hculea: I would suggest not to do that. We're planning to add `XDG_CONFIG_FILE` support for all plugins at once: https://github.com/1Password/shell-plugins/issues/153

Easiest would be to use the hardcoded location for now, as it is currently in place. This would ensure that different plugins don't use diverging standards for handling this variable, which would make the migration harder, later on. 😄

Let me know if you disagree, @AndyTitu

AndyTitu: Resolving

arunsathiya:
@AndyTitu, I am unable to copy to that location because that location doesn't exist:

```
➜  ~ docker cp ~/.op/plugins/local/fastly a1ac133d0753:/root/.op/plugins/local
Error: No such container:path: a1ac133d0753:/root/.op/plugins
```

Following [this documentation](https://developer.1password.com/docs/cli/config-directories/), the envvar XDG_CONFIG_HOME isn't set in my Docker-based Ubuntu container, which is why I am copying to `/root/.config/op/plugins/local`

```
root@a1ac133d0753:/# echo $XDG_CONFIG_HOME

```

Overall though, I believe we don't have any blockers, except that I have not tested the Linux importer.

hculea (approved):
Looks good to me! Thank you for this contribution! ❤️

hculea:
I think the issue in your tests above @arunsathiya is that the path actually doesn't exist, since those directories never got created (that happens automatically, when building a plugin). Running a `mkdir -p /root/.config/op/plugins/local` inside the container should fix the issue.

One more thing to have in mind - the binaries are OS dependent, so you may want to set `GOOS` and `GOARCH` to those of your container, before running the `make build` command on the plugin.

That being said, I don't see not having tested the Linux importer as a blocker. Good to go IMO! 🚢

SimonBarendse (approved):
Regarding the `fastly profile` commands, I think it makes total sense for the 1Password Shell Plugin not to operate on those. When you're using the 1Password shell plugin for Fastly, you no longer need to use these commands, since you're managing the profiles/credentials through 1Password. So this is perfect as is :+1: :rocket: We'll include it in the next 1Password CLI release.

Appreciate yet another amazing contribution, thank you @arunsathiya ! :heart:

PR #172, opened 2023-02-02
Add StackStorm shell plugin

mamercad:
Add [StackStorm](https://stackstorm.com) `st2` CLI shell plugin.


hculea:
Hey @mamercad , thank you for your contribution! I noticed that your PR has been in draft for a while.

Is there a specific reason it is not yet ready for review? Feel free to reach out to us, if you think we could help!
You can either:
* Book a free [pairing session](https://calendly.com/d/grs-x2h-pmb/1password-shell-plugins-pairing-session) with one of our developers
* Join the [Developer Slack workspace](https://join.slack.com/t/1password-devs/shared_invite/zt-1halo11ps-6o9pEv96xZ3LtX_VE0fJQA), and ask us any questions there

Let us know! We'd love to get this rolling 😄

mamercad:
> Hey @mamercad , thank you for your contribution! I noticed that your PR has been in draft for a while.
> 
> Is there a specific reason it is not yet ready for review? Feel free to reach out to us, if you think we could help! You can either:
> 
> * Book a free [pairing session](https://calendly.com/d/grs-x2h-pmb/1password-shell-plugins-pairing-session) with one of our developers
> * Join the [Developer Slack workspace](https://join.slack.com/t/1password-devs/shared_invite/zt-1halo11ps-6o9pEv96xZ3LtX_VE0fJQA), and ask us any questions there
> 
> Let us know! We'd love to get this rolling 😄

Ah, I can't quite remember, I've been busy with `$work` 😦 I'll set a reminder to get back to this.

mamercad:
> > Hey @mamercad , thank you for your contribution! I noticed that your PR has been in draft for a while.
> > Is there a specific reason it is not yet ready for review? Feel free to reach out to us, if you think we could help! You can either:
> > 
> > * Book a free [pairing session](https://calendly.com/d/grs-x2h-pmb/1password-shell-plugins-pairing-session) with one of our developers
> > * Join the [Developer Slack workspace](https://join.slack.com/t/1password-devs/shared_invite/zt-1halo11ps-6o9pEv96xZ3LtX_VE0fJQA), and ask us any questions there
> > 
> > Let us know! We'd love to get this rolling 😄
> 
> ~Ah, I can't quite remember, I've been busy with `$work` 😦 I'll set a reminder to get back to this.~

I've marked it ready for review.

plugins/st2/auth_token.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AuthToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.AuthToken,
+		DocsURL:       sdk.URL("https://docs.stackstorm.com/authentication.html"),
+		ManagementURL: sdk.URL("https://docs.stackstorm.com/authentication.html"),
```
hculea, line 18: The management URL should be the address where users can create new tokens/see existing ones.

As an example, for GitHub, that is https://github.com/settings/tokens.

plugins/st2/auth_token.go
```diff
@@ -0,0 +1,66 @@
...
+				Composition: &schema.ValueComposition{
+					Length: 32,
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
```
hculea, line 33: Screenshot 2023-03-30 at 10 26 29

I found this in the documentation of `st2`. We should take this into account, as, if I understand it correctly, it otherwise it would basically cache the first token it receives. 

Is this something that would have to be done manually by the user, as a step of setting up the plugin?

hculea: Link: https://docs.stackstorm.com/reference/cli.html

plugins/st2/st2.go
```diff
@@ -0,0 +1,25 @@
...
+
+func StackStormCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "StackStorm st2 CLI",
+		Runs:      []string{"st2"},
+		DocsURL:   sdk.URL("https://docs.stackstorm.com/reference/cli.html"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
```
hculea, line 18: I think we should also circumvent auth in case `--config-file` is passed.
```
st2 --config-file=~/.st2/prod-config action list
```
is the example given in the docs.

hculea (commented):
Thank you for your contribution! ❤️

plugins/st2/auth_token.go
SimonBarendse: If this doesn't exist for the service, you can leave it `nil`

plugins/st2/auth_token.go
hculea: For context: Shell Plugins aim to get rid of the need to store secrets in plain-text on disk. My concern is that with the cache enabled, this would not be the case.

mamercad: > Is this something that would have to be done manually by the user, as a step of setting up the plugin?

To the best of my knowledge, setting this would be up to the user.

plugins/st2/auth_token.go
mamercad: > If this doesn't exist for the service, you can leave it `nil`

I don't think there's a way to see these in the UI, but, there's the API endpoint [`POST /auth/v1/tokens`](https://api.stackstorm.com/auth/v1/tokens/#/token_controller.post); not sure if/how to represent this "partial" in `ManagementURL` (I'll push with `nil`).

plugins/st2/auth_token.go
mamercad: > For context: Shell Plugins aim to get rid of the need to store secrets in plain-text on disk. My concern is that with the cache enabled, this would not be the case.

Yes, without `cache_token: False`, after logging in, `~/.st2/token-<username>` gets created and looks like this:

```shell
❯ cat ~/.st2/token-mark
{"token": "thisisthetokenhunter2", "expire_timestamp": 1680391552}
```

mamercad: Perhaps relying on the config file `~/.st2/config` is the best path forward?

```
[credentials]
# Credentials used to authenticate against the auth API and retrieve the auth
# token
username = test1
password = testpassword
```

AndyTitu:
This PR seems pretty close to the finish line actually! 

What I would propose we do is:
1. make sure we always provision `cache_token: false` to the tool so we make sure the retrieved token does not get spilled to disk. However, I am wondering whether the toke, which gets written to disk and which seems to be actually used for performing the authentication, is the actual token that you have in 1Password, or is a temporary one as hinted by the `Expiry` field that follows it. 
 - if that's a temporary token we need to find out how to retrieve that, maybe using an already existing SDK.
 - if that temporary token cannot be retrieved we should stick to the username/password approach with the file provisioner.
2. If we go for the token, given that we are not caching it anymore, let's use the shell-plugin built in cache to store that.

Let me know if this makes sense @mamercad !

plugins/st2/auth_token.go
AndyTitu: I've continued this thread [here](https://github.com/1Password/shell-plugins/pull/172#issuecomment-1523207921).

mamercad:
> This PR seems pretty close to the finish line actually!
> 
> What I would propose we do is:
> 
> 1. make sure we always provision `cache_token: false` to the tool so we make sure the retrieved token does not get spilled to disk. However, I am wondering whether the toke, which gets written to disk and which seems to be actually used for performing the authentication, is the actual token that you have in 1Password, or is a temporary one as hinted by the `Expiry` field that follows it.
> 
> * if that's a temporary token we need to find out how to retrieve that, maybe using an already existing SDK.
> * if that temporary token cannot be retrieved we should stick to the username/password approach with the file provisioner. However, if possible, I'd prefer to sticking with the env var approach since that's less prone to changes/errors in the future.
> 
> 2. If we go for the token, given that we are not caching it anymore, let's use the shell-plugin built in cache to store that.
> 
> Let me know if this makes sense @mamercad !

It does, though, I think I'd like to try switching to username and password (pushed). But, I'm not having much luck:

```shell
❯ op plugin init st2
####################################################################
# WARNING: 'st2' is not from the official registry.                #
# Only proceed if you are the developer of 'st2'.                  #
# Otherwise, delete the file at /Users/mark/.op/plugins/local/st2. #
####################################################################

StackStorm CLI [test build]
Authenticate with StackStorm Username and Password.

? Locate your StackStorm Username and Password: StackStorm (Homelab)

op: The "StackStorm (Homelab)" item does not follow required item structure for credential type Username and Password:
    username    Username used to authenticate to StackStorm.
    password    Password used to authenticate to StackStorm.
    website     StackStorm base URL.

To continue, you'll have to update the field names. This can be done interactively. Proceed? [Y/n] y
[ERROR] 2023/04/29 16:40:07 cannot edit item: "StackStorm" has fewer fields than the Username and Password credential type requires. Edit your item to include at least 3 fields with non-empty labels or select another item
```

Feels like I have the right labels:

```shell
❯ op item get "StackStorm" --format=json | grep label
      "label": "website",
      "label": "username",
      "label": "password",
      "label": "notesPlain",
      "label": "base_url",
```

Any ideas?

AndyTitu:
> ```shell
> username
> ```

@mamercad I understand the confusion. I think what the CLI complains about is the item not having the proper amount of fields to be representative for such a credential type. The label here refers to the name of the field rather than the `label` type. This should be fixed by creating an item with fields (they can just be text fields) matching the names of the secrets/configs username, password, website.

mamercad:
> creating an item

What kind of item?

AndyTitu:
@mamercad Oh, my bad, I got confused. I think I found the problem. Are these fields inside sections or do they have a specific designation (is the item a login type and the username and password fields are an actual main `username` and `password` of the item)? This will cause the error you are seeing. `username` and `password` should be flat mapped inside the item (they shouldn't be part of any section) and they should not have any designation for them to be valid candidates for renaming. I understand this is a bit weird. I'll look into creating an issue for this.

mamercad:
> @mamercad Oh, my bad, I got confused. I think I found the problem. Are these fields inside sections or do they have a specific designation (is the item a login type and the username and password fields are an actual main `username` and `password` of the item)? This will cause the error you are seeing. `username` and `password` should be flat mapped inside the item (they shouldn't be part of any section) and they should not have any designation for them to be valid candidates for renaming. I understand this is a bit weird. I'll look into creating an issue for this.

It looks like this:

image

AndyTitu:
> > @mamercad Oh, my bad, I got confused. I think I found the problem. Are these fields inside sections or do they have a specific designation (is the item a login type and the username and password fields are an actual main `username` and `password` of the item)? This will cause the error you are seeing. `username` and `password` should be flat mapped inside the item (they shouldn't be part of any section) and they should not have any designation for them to be valid candidates for renaming. I understand this is a bit weird. I'll look into creating an issue for this.
> 
> It looks like this:
> 
> image

The error could appear because of the duplicate field names in the item as the CLI doesn't know which one to choose. Could you try removing the designated fields (the ones on top) and only keep the regular ones? That said, I checked and we are tracking internally the improvement of how we are looking for matching fields, as this was a recurrent point of feedback.

The MR seems ready imo, so if we could resolve this last issue, I'd say it would be close to merging

PR #176, opened 2023-02-11
Add plugin for Treasure Data Toolbelt

Lewuathe:
# Summary
Add support for a new plugin for the [TreasureData Toolbelt](https://docs.treasuredata.com/display/public/PD/TD+Toolbelt)

# Testing
## Setup
1. Create a TreasureData account through the site [here](https://www.treasuredata.com/).
2. Get the API key for CLI tools according to the [guide](https://docs.treasuredata.com/display/public/PD/Getting+Your+API+Keys).
3. Following the [guide](https://docs.treasuredata.com/display/public/PD/Installing+TD+Toolbelt+and+Treasure+Agent), setup the Treasure Data Toolbelt in your environment 
3. Build and test the TreasureData plugin locally according to [these instructions](https://developer.1password.com/docs/cli/shell-plugins/contribute/). 

```
$ make treasuredata/validate
$ make treasuredata/build
```

# Importer

After installing and setting up Treasure Data Toolbelt, a config file should have been created at `~/.td/td.conf`. 
Initialize the Treasure Data plugin and test the importer using this config file. The credentials should be saved in 1Password successfully by selecting this config file. The entry should have `user` and `api key` fields.

# Provisioner
After initializing the Treasure Data plugin and having your Treasure Data credentials saved in 1Password, run a `td` command. The provisioner should inject `TD_API_KEY` environment so that the tool can refer to the necessary credential at runtime.
Try running the following command.

```
$ td db:list
+---------------------+----------+
| Name                | Count    |
+---------------------+----------+
| sample_datasets     | 0        |
+---------------------+----------+
```

Note: While testing the provisioner, ensure that your credentials are not currently stored in `~/.td/td.conf` nor environment variables.

### DIFF
--- a/plugins/treasuredata/access_key.go
+++ b/plugins/treasuredata/access_key.go
@@ -1,0 +1,68 @@
+package treasuredata
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AccessKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.AccessKey,
+		DocsURL:       sdk.URL("https://docs.treasuredata.com/display/public/PD/TD+Toolbelt"),
+		ManagementURL: sdk.URL("https://console.treasuredata.com/app/mp/ak"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.User,
+				MarkdownDescription: "User name specified by email",
+			},			
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "APIKey used to authenticate to Treasure Data.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			TryTreasureDataConfigFile("~/.td/td.conf"),
+		)}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"TD_API_KEY": fieldname.APIKey,
+}
+
+func TryTreasureDataConfigFile(path string) sdk.Importer {
+	return importer.TryFile(path, func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		credentialsFile, err := contents.ToINI()
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		fields := make(map[sdk.FieldName]string)
+		for _, section := range credentialsFile.Sections() {
+			if section.HasKey("user") && section.Key("user").Value() != "" {
+				fields[fieldname.User] = section.Key("user").Value()
+			}			
+			if section.HasKey("apikey") && section.Key("apikey").Value() != "" {
+				fields[fieldname.APIKey] = section.Key("apikey").Value()
+			}
+		}
+
+		out.AddCandidate(sdk.ImportCandidate{
+			Fields: fields,
+		})
+	})
+}
--- a/plugins/treasuredata/td.go
+++ b/plugins/treasuredata/td.go
@@ -1,0 +1,25 @@
+package treasuredata
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func TreasureDataCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Treasure Data Toolbelkt",
+		Runs:      []string{"td"},
+		DocsURL:   sdk.URL("https://docs.treasuredata.com/display/public/PD/TD+Toolbelt"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
+		Uses: []schema.CredentialUsage{
+			{
+				Name: credname.AccessKey,
+			},
+		},
+	}
+}
--- a/plugins/treasuredata/test-fixtures/td.conf
+++ b/plugins/treasuredata/test-fixtures/td.conf
@@ -1,0 +1,3 @@
+[account]
+  user = user@example.com
+  apikey = 1/xxx
--- a/plugins/treasuredata/access_key_test.go
+++ b/plugins/treasuredata/access_key_test.go
@@ -1,0 +1,42 @@
+package treasuredata
+
+import (
+	"testing"
+	
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+	
+func TestAccessKeyProvisioner(t *testing.T) {
+	plugintest.TestProvisioner(t, AccessKey().DefaultProvisioner, map[string]plugintest.ProvisionCase{
+		"default": {
+			ItemFields: map[sdk.FieldName]string{
+				fieldname.APIKey: "1/xxx",
+			},
+			ExpectedOutput: sdk.ProvisionOutput{
+				Environment: map[string]string{
+					"TD_API_KEY": "1/xxx",
+				},
+			},
+		},
+	})
+}
+
+func TestAccessKeyImporter(t *testing.T) {
+	plugintest.TestImporter(t, AccessKey().Importer, map[string]plugintest.ImportCase{
+		"TD config file": {
+			Files: map[string]string{
+				"~/.td/td.conf": plugintest.LoadFixture(t, "td.conf"),
+			},
+			ExpectedCandidates: []sdk.ImportCandidate{
+				{
+					Fields: map[sdk.FieldName]string{
+						fieldname.User: "user@example.com",
+						fieldname.APIKey: "1/xxx",
+					},
+				},
+			},
+		},
+	})
+}
--- a/plugins/treasuredata/plugin.go
+++ b/plugins/treasuredata/plugin.go
@@ -1,0 +1,22 @@
+package treasuredata
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "treasuredata",
+		Platform: schema.PlatformInfo{
+			Name:     "Treasure Data",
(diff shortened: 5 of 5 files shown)


AndyTitu:
Thanks for the contribution @Lewuathe !

Code lgtm but static check is failing due to unformatted files.

Lewuathe:
@AndyTitu Thank you for letting me know. I fixed the format. Could you please take a look at this one again?

plugins/treasuredata/access_key.go
```diff
@@ -0,0 +1,68 @@
...
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			TryTreasureDataConfigFile("~/.td/td.conf"),
+		)}
```
hculea, line 39: If provisioning can be done using envvars, I would suggest also providing an importer option for envvars as well!

plugins/treasuredata/access_key.go
```diff
@@ -0,0 +1,68 @@
...
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
```
hculea, line 37: On a second glance, looks that there are multiple ways to provision the API Key:

```
API key specified by td -k option, which allows you to specify an API key instead of reading the config file

TREASURE_DATA_API_KEY environment variable

TD_API_KEY environment variable

TD_API_KEY in config file which path is specified by "td -c" option

TD_API_KEY in config file which path is specified by TREASURE_DATA_CONFIG_PATH environment variable

TD_API_KEY in config file which path is specified by TD_CONFIG_PATH environment variable

TD_API_KEY in config file locates at ~/.td/td.conf
```

This would mean the following:
* plugin should opt out of authentication when `-k` or `-c` flags are provided
* plugin should also try to import from `TREASURE_DATA_API_KEY` envvar
* plugin should also import from files found under `TREASURE_DATA_CONFIG_PATH` and `TD_CONFIG_PATH`

I think that the functionality can very easily be adapted to support all these use-cases, let me know if I'm missing anything! 😄

plugins/treasuredata/access_key.go
```diff
@@ -0,0 +1,68 @@
...
+func AccessKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.AccessKey,
+		DocsURL:       sdk.URL("https://docs.treasuredata.com/display/public/PD/TD+Toolbelt"),
+		ManagementURL: sdk.URL("https://console.treasuredata.com/app/mp/ak"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.User,
+				MarkdownDescription: "User name specified by email",
+			},
```
hculea, line 23: Is there any reason to require this field in this plugin specification, given that the username is not used in the provisioning step?

plugins/treasuredata/access_key.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		fields := make(map[sdk.FieldName]string)
+		for _, section := range credentialsFile.Sections() {
+			if section.HasKey("user") && section.Key("user").Value() != "" {
+				fields[fieldname.User] = section.Key("user").Value()
+			}
```
hculea, line 58: Same comment as above - if this field does not have to be provisioned, I think it is best not to import it either.

hculea (commented):
Thank you for this contribution! <3 I've left a few comments that we should address before merging.

plugins/treasuredata/access_key.go
Lewuathe: > plugin should opt out of authentication when -k or -c flags are provided

`td` toolbelt automatically prioritizes the key given in the options (e.g. `-c` or `-k`) so I think we do not need to opt-out the authentication by the plugin by itself.

> plugin should also try to import from TREASURE_DATA_API_KEY envvar
> plugin should also import from files found under TREASURE_DATA_CONFIG_PATH and TD_CONFIG_PATH

That makes sense. I'll try to support this case too.

Lewuathe (commented):
@hculea Thanks for the review. I updated the plugin accordingly.

Lewuathe (commented):
@AndyTitu @hculea Sorry for bothering you but could you take a look at the change after the last review and please give me the feedback if there is something.

plugins/treasuredata/access_key.go
hculea: > td toolbelt automatically prioritizes the key given in the options (e.g. -c or -k) so I think we do not need to opt-out the authentication by the plugin by itself.

Even if it prioritises the key given as flags, the 1Password CLI would still prompt for credentials saved in 1Password. This leads to a degraded user experience - user shouldn't have to pick a credential from 1P if they already provided one as a flag.

hculea:
Terribly sorry for the long turnaround! Only one other comment on an older thread, otherwise, this looks good to go! Thank you for your contribution!!

plugins/treasuredata/access_key.go
Lewuathe: @hculea Okay. But I'm not sure how to opt-out of the authentication if `-c` or `-k` options are provided. Could you give me a pointer to the resource or guide to do so?

Lewuathe: @hculea Okay. But I'm not sure how to opt-out of the authentication if `-c` or `-k` options are provided. Could you give me a pointer to the resource or guide to do so?

hculea: Of course! This would be tackled in the `schema.Executable` struct, specifically within the `NeedsAuth` field. Have a look at how e.g. the `fastly` plugin approaches that:
```
		NeedsAuth: needsauth.IfAll(
			needsauth.NotWithoutArgs(),
			needsauth.NotWhenContainsArgs("-t"),
			needsauth.NotWhenContainsArgs("--token"),
			needsauth.NotWhenContainsArgs("profile"),
			needsauth.NotWhenContainsArgs("config"),
		),
```

Let me know if this makes sense!

Lewuathe: Great. That's what I needed. Thanks!

hculea: Awesome! Let me know when to have another look 😄

Lewuathe: I have updated the patch to skip the auth if given `-c` or `-k` option.

https://github.com/1Password/shell-plugins/pull/176/commits/a1be37e6a32b0e2c9d616d33d6c0eda888efd953

plugins/treasuredata/access_key.go
```diff
@@ -0,0 +1,68 @@
...
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryEnvVarPair(defaultEnvVarMapping),
+			importer.TryEnvVarPair(map[string]sdk.FieldName{
+				"TREASURE_DATA_API_KEY": fieldname.APIKey,
+			}),
+			TryTreasureDataConfigFile(os.Getenv("TREASURE_DATA_CONFIG_PATH")),
```
SimonBarendse, line 39: If these envvars are unset, we're calling `importer.TryFile` with an empty string for the path. Now I believe that leads to the expected behavior because the operating system will return a not found error for the file because the file "" doesn't exist. I'm wondering if we should depend on this to happen at the operating system level, or if we should already abort trying to import when the envvar is unset.

@florisvdg do you forese any issues with the current approach?

plugins/treasuredata/td.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func TreasureDataCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Treasure Data Toolbelkt",
```
SimonBarendse, line 12:
```suggestion
		Name:    "Treasure Data Toolbelt",
```

SimonBarendse (commented):
Thank you for the contribution @lewuathe! Welcome Treasure Data to the ecosystem! ❤️

I think I've found a small typo, and there's one last thing where I think we're good, but I'd like to confirm with a teammate who knows more about that area.

plugins/treasuredata/access_key.go
hculea: Thank you!

plugins/treasuredata/access_key.go
florisvdg: The `importer.TryFile` helper already ignores files that don't exist, so I don't think there's an issue there. We can look into ways to make this a bit more rigid in the framework, but that doesn't have to block this PR.

SimonBarendse: Thank you for the quick check, Floris!

SimonBarendse (approved):

Lewuathe (commented):
Thanks for the review!

But I see the lint error in CI.

https://github.com/1Password/shell-plugins/actions/runs/4323502899/jobs/7547182221


I could not reproduce the issue on my machine with the following version. Is there anything wrong with the format?
```
$ golangci-lint run

$ golangci-lint --version
golangci-lint has version 1.51.2 built from 3e8facb on 2023-02-19T20:54:52Z
```

SimonBarendse:
Hi @Lewuathe, since Friday lint jobs started to fail across this repo, and even in other repos on GitHub Actions as well. So it appears to be independent of the changes introduced in this pull request. We're looking into the issue https://github.com/1Password/shell-plugins/issues/195

florisvdg (commented):
Looks good! Left a few small comments below.

plugins/treasuredata/access_key.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AccessKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.AccessKey,
```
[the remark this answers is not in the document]
Lewuathe: We can call it "API Key". I'll rename it.

plugins/treasuredata/test-fixtures/td.conf
```diff
@@ -0,0 +1,3 @@
+[account]
+  user = user@example.com
+  apikey = 1/xxx
```
florisvdg, line 3: Final nit: it would be nice if the test fixture would use a more realistically looking value, like we do in [other plugins](https://github.com/1Password/shell-plugins/blob/8b715d0cd6f469f5142ccfad9df49ead3c149945/plugins/aws/test-fixtures/credentials). You can generate such a value using the `make treasuredata/example-secrets` command.

SimonBarendse (approved):
Thank you for the contribution @Lewuathe! ❤️ And welcome to the ecosystem Treasure Data 🎉

Lewuathe (commented):
@SimonBarendse @hculea Thanks for the tailored review!

Lewuathe (commented):
@florisvdg @SimonBarendse Is there any specific reason not to merge this PR even after it's approved? Is there something I can do?

Lewuathe:
@SimonBarendse Thanks!

PR #177, opened 2023-02-12
New Plugin:  SAPBTP cli 

derekvincent:
# Summary

Provides support for a new plugin for [SAP's BTP CLI](https://help.sap.com/docs/btp/sap-business-technology-platform/account-administration-using-sap-btp-command-line-interface-btp-cli).

**NOTE:** Plugin name is SAPBTP, but the CLI command is `btp`. 

# Setup 
1.  Setup an [SAP BTP account](https://account.hana.ondemand.com/) - (trial, free or paid) 
2. Provision user - standard using SAP's Account IDP
3.  Download and setup the [SAP BTP Command Line Interface (btp CLI)](https://tools.hana.ondemand.com/#cloud)
4. Build and test the SAPBTP plugin locally according to [these instructions](https://developer.1password.com/docs/cli/shell-plugins/contribute/).
```bash
$ make sapbtp/validate
$ make sapbtp/build
```

# Importer

The BTP CLI config file maintains the current setting and the refresh token created on the initial credentialed login.  This is not particularly useful for populating credentials, and an importer is not currently implemented. 

# Provisioner 

The BTP CLI authentication is done with an explicit login using the `login` command.  The current plugin will only trigger on that CLI command and adds the `--user` and `--password` arguments to the command line.  Additional login arguments such as `--url` and `--subdomain` can still be provided or entered as per the standard flow. 

## Example
```zsh
$ btp login --url https://cpcli.cf.eu10.hana.ondemand.com --subdomain t5822952trial-ga

###############################################################################
# WARNING: 'sapbtp' is not from the official registry.                        #
# Only proceed if you are the developer of 'sapbtp'.                          #
# Otherwise, delete the file at /Users/username/.op/plugins/local/sapbtp. #
###############################################################################
Connecting to CLI server at https://cpcli.cf.eu10.hana.ondemand.com...
Autocomplete script has been updated to include the latest commands. Restart your terminal session to load the newest autocomplete information.

Authentication successful

Current target:
  t5822952trial (global account, subdomain: t5822952trial-ga)

We stored your configuration file at: /Users/username/Library/Application Support/.btp/config.json

Tips:
    Commands are executed in the target, unless specified otherwise using a parameter. To change the target, use 'btp target'.
    To provide feedback about the btp CLI, use 'btp feedback' to open our survey.

OK
```

### DIFF
--- a/plugins/sapbtp/btp.go
+++ b/plugins/sapbtp/btp.go
@@ -1,0 +1,26 @@
+package sapbtp
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func SAPBTPCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "SAP BTP CLI",
+		Runs:    []string{"btp"},
+		DocsURL: sdk.URL("https://help.sap.com/docs/btp/sap-business-technology-platform/account-administration-using-sap-btp-command-line-interface-btp-cli"), // TODO: Replace with actual URL
+		NeedsAuth: needsauth.IfAll(
+			needsauth.ForCommand("login"),
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
+		Uses: []schema.CredentialUsage{
+			{
+				Name: credname.Credentials,
+			},
+		},
+	}
+}
--- a/plugins/sapbtp/credentials.go
+++ b/plugins/sapbtp/credentials.go
@@ -1,0 +1,39 @@
+package sapbtp
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func Credentials() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.Credentials,
+		DocsURL:       sdk.URL("https://help.sap.com/docs/BTP/65de2977205c403bbc107264b8eccf4b/cc1c676b43904066abb2a4838cbd0c37.html"),
+		ManagementURL: sdk.URL("https://account.hana.ondemand.com/"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Username,
+				MarkdownDescription: " SAP BTP Username (email address)",
+				Optional:            false,
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{Uppercase: true, Lowercase: true, Digits: true},
+					Prefix:  "",
+				},
+			},
+			{
+				Name:                fieldname.Password,
+				MarkdownDescription: " SAP BTP Password",
+				Optional:            false,
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{Uppercase: true, Lowercase: true, Digits: true},
+					Prefix:  "",
+				},
+			},
+		},
+		DefaultProvisioner: BTPProvisioner(),
+	}
+}
--- a/plugins/sapbtp/credentials_test.go
+++ b/plugins/sapbtp/credentials_test.go
@@ -1,0 +1,28 @@
+package sapbtp
+
+import (
+	"testing"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func TestCredentialsProvisioner(t *testing.T) {
+	plugintest.TestProvisioner(t, Credentials().DefaultProvisioner, map[string]plugintest.ProvisionCase{
+		"default": {
+			ItemFields: map[sdk.FieldName]string{ // TODO: Check if this is correct
+				fieldname.Username: "john.doe@email.com",
+				fieldname.Password: "CpfSh78WLKpO6teQitmYbH7EXAMPLE",
+			},
+			ExpectedOutput: sdk.ProvisionOutput{
+				CommandLine: []string{
+					"--user",
+					"john.doe@email.com",
+					"--password",
+					"CpfSh78WLKpO6teQitmYbH7EXAMPLE",
+				},
+			},
+		},
+	})
+}
--- a/plugins/sapbtp/plugin.go
+++ b/plugins/sapbtp/plugin.go
@@ -1,0 +1,22 @@
+package sapbtp
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "sapbtp",
+		Platform: schema.PlatformInfo{
+			Name:     "SAP BTP",
+			Homepage: sdk.URL("https://help.sap.com/docs/BTP"),
+		},
+		Credentials: []schema.CredentialType{
+			Credentials(),
+		},
+		Executables: []schema.Executable{
+			SAPBTPCLI(),
+		},
+	}
+}
--- a/plugins/sapbtp/provisioner.go
+++ b/plugins/sapbtp/provisioner.go
@@ -24,13 +24,12 @@ func (p btpProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, ou
 		out.AddArgs("--password", password)
 	}
 
-	return
 }
 
 func (p btpProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput, out *sdk.DeprovisionOutput) {
-	// Nothing to do here: environment variables get wiped automatically when the process exits.
+	// Nothing to do here: command line args are wiped when process exits
 }
 
 func (p btpProvisioner) Description() string {
-	return "Provision command line variables --user and --password with BTP cli"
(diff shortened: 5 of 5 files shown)


AndyTitu (commented):
Thanks for the contribution @derekvincent !

> The BTP CLI config file maintains the current setting and the refresh token created on the initial credentialed login. 
Is this file by any chance used for authentication via the refresh token? My understanding is that the `--username  --password` approach is only to be used once for getting this config file, while all subsequent requests should rely on the persisted file for authentication. In this case, we should find a way of provisioning this file, rather than using the username and password for each and every command.

derekvincent:
@AndyTitu this file does contain the refresh token that is created during the `btp login` command.  The plugin should only pass the username and password parameters on the `login` command, as the other authenticated commands do not accept specifying login parameters and rely on the generated config file and refresh token. 

> Is this file by any chance used for authentication via the refresh token? My understanding is that the `--username --password` approach is only to be used once for getting this config file, while all subsequent requests should rely on the persisted file for authentication. In this case, we should find a way of provisioning this file, rather than using the username and password for each and every command.

AndyTitu:
@derekvincent oh I see! This also explains `needsauth.ForCommand("login")`. This makes sense.

An additional question: 
> other authenticated commands do not accept specifying login parameters and rely on the generated config file and refresh token
So this means that the refresh token could be used by someone else to authenticate on your behalf. In this case, is it safe to store this in plain-text on disk in the config file? This is indeed a temporary credential so the risk decreases a bit. Would it be worth looking into a file-provisioner approach together with using the encrypted plugin cache for serving the temporary refresh token to the config file, while using the `login` (username + password) logic only when needing to refresh the temporary config file, from a security perspective @DCKcode ?

DCKcode:
@AndyTitu Based on the description here I think the file provisioner approach would be well suited here 🙂 

> If the risk of that temporary token is not perceived to be too high, I am okay to ship this as is.

My understanding here is that that token is your actual authorization, so we should avoid storing it to disk (as a real file at least, a temporary file would be okay). But if we can leverage the file provisioner approach for that, we should be fine to do that 🙂

derekvincent:
@DCKcode and @AndyTitu, thanks for the feedback I am been thinking it over and want to make sure I am clear on the process and have some thoughts on it. 

Potential process flow: 
1. Use the plugin to fill in the `--user` and `--password` on the `btp login` command
2. BTP cli will login and create/modify its config state json file 
3. The plugin will then read the file back in, store the contents in the cache and remove the config file 
4. When issuing another authenticated btp cli command, use the cached file contents to create a temporary config file and then remove the file after the command is run

Going over the plugin structure I do not see how to make that work but it could simply be my understanding. Also, the config file is more than just for auth/credentials but also a state file with values populated during the login from the server, certain cli commands will also change values in the file. I really like the idea of the file not being persistent on disk, I am not sure it is a great idea. 

Thoughts?

AndyTitu:
> 1. Use the plugin to fill in the --user and --password on the btp login command
2. BTP cli will login and create/modify its config state json file
3. The plugin will then read the file back in, store the contents in the cache and remove the config file
4. When issuing another authenticated btp cli command, use the cached file contents to create a temporary config file and then remove the file after the command is run

@derekvincent Thanks for looking into this! I would suggest the following workflow which addresses your concerns (we also used this approach when dealing with [AWS MFA temporary credentials](https://github.com/1Password/shell-plugins/blob/main/plugins/aws/sts_provisioner.go)):
0. Research time: Identify which fields in the config file are actually used for the auth. Refresh token is more for refreshing the session when needed, there should also exist and `auth token` of some sort there, and that's what we are mainly interested in. 

> Also, the config file is more than just for auth/credentials but also a state file with values populated during the login from the server, certain cli commands will also change values in the file

The ideal scenario would be that there exists a SAP supported environment variable that you can specify with that auth token. This would nicely separate the responsibility of authentication from the one of settings, as we could solely focus on provisioning that env var.

1. use the plugin to fetch username/password from 1Password. Take these credentials and use a [SAP SDK](https://github.com/sap/cloud-security-client-go) for fetching the temporary auth credentials, by providing these username/password credentials.

2. Provision the required env var/ file with the temporary credentials. If there is no env var we could use, things become more difficult to handle with the in-memory file provisioning method that we are using, as you also noted:

> the config file is more than just for auth/credentials but also a state file with values populated during the login from the server, certain cli commands will also change values in the file

so if that is the only alternative, there ought to to be some more discussion around what we are going to do, so please do return here 😄 

3. Store these temporary credentials in cache so we do not retrieve them from remote all the time. We should start retrieving them only after they expire.

Hope this helps a bit!

derekvincent:
@AndyTitu, thanks for the response and guidance. I have been having some fun digging into this. 

I explored more into what is going on, and the CLI service API that is used is not documented, and we should only be using the btp cli command line tool.  Using the `fs_usgae` command, I do not see any additional files that the application is creating that may store additional data.  I also used a tool to log the web API service calls the cli was using and have been able to sort out what is going on and replicate the calls with `curl` commands. 

The btp cli is calling a cli-specific login service with the provided authentication credentials, and the only thing returned is the refresh token. The next call will use the refresh token and return a new refresh token in the return header. This new token is then used for the next call, and so on.  

I suspect, based on the cli design, the approaches above won't work as the cli returns a new refresh token (not sure it is a refresh token).  

Thoughts or suggestions on how to proceed? 

Derek

derekvincent:
@AndyTitu, thanks for the response and guidance. I have been having some fun digging into this. 

I explored more into what is going on, and the CLI service API that is used is not documented, and we should only be using the btp cli command line tool.  Using the `fs_usgae` command, I do not see any additional files that the application is creating that may store additional data.  I also used a tool to log the web API service calls the cli was using and have been able to sort out what is going on and replicate the calls with `curl` commands. 

The btp cli is calling a cli-specific login service with the provided authentication credentials, and the only thing returned is the refresh token. The next call will use the refresh token and return a new refresh token in the return header. This new token is then used for the next call, and so on.  

I suspect, based on the cli design, the approaches above won't work as the cli returns a new refresh token (not sure it is a refresh token).  


Derek

AndyTitu:
@derekvincent thanks for investigating this further and I'm glad you had some fun while at it! Did you double check that the refresh token is used for authentication in this case, by trying to edit your config and removing the refresh token? Then if following commands would keep working without saying `you are not authorized`, this would be mean that the Auth Token exists somewhere else. Otherwise, if commands that worked when the refresh token was there stop working when the refresh token is not there, we have pinpointed with certainty that the refresh token is the secret we are looking to provision. However, I personally don't believe this to be the case, because usually refresh tokens are used to refresh the session. If the refresh token would act as an authentication credential at the same time, it would mean that with each cli command you are doing, a call to SAP BTP server is made, and this would not be too good for their server load, nor the speed of the command. I also saw there is an `Authentication` section in the config file. 

This is awesome work you are doing! I am just trying to make sure we are following the right lead.

AndyTitu:
Hi @derekvincent , sorry for the delayed response, but after some internal analysis, we've realised that because the refresh token gets refreshed and written back to the file with each call, the file provisioning approach will not be suitable here. The file we are provisioning is an in-memory FIFO which blocks for all reads and writes. When **we** write to it, it's okay because we have the control over the file, but when another process (in this case the sap btp cli) writes to it, that cannot be predicted in a deterministic way, so the process will just remain blocked. 

The only way around this would be to provision an environment variable using the env var provisioner, but I'm not sure if such an env var exists. Pursuing this, for now, would mean looking into what environment variables are out there for the sap btp cli and maybe trying to find one called REFRESH_TOKEN. If this does not exist, a long shot would be to ask the developer/support team of SAP to expose one. Apologies for the hardships encountered with the plugin for this specific tool.

derekvincent:
> Hi @derekvincent , sorry for the delayed response, but after some internal analysis, we've realised that because the refresh token gets refreshed and written back to the file with each call, the file provisioning approach will not be suitable here. The file we are provisioning is an in-memory FIFO which blocks for all reads and writes. When **we** write to it, it's okay because we have the control over the file, but when another process (in this case the sap btp cli) writes to it, that cannot be predicted in a deterministic way, so the process will just remain blocked.
> 
> The only way around this would be to provision an environment variable using the env var provisioner, but I'm not sure if such an env var exists. Pursuing this, for now, would mean looking into what environment variables are out there for the sap btp cli and maybe trying to find one called REFRESH_TOKEN. If this does not exist, a long shot would be to ask the developer/support team of SAP to expose one. Apologies for the hardships encountered with the plugin for this specific tool.

@AndyTitu, sorry for the delay; I did some digging and do not see anything around an env variable to support this. I sent an enhancement request to SAP, but because of the size that SAP is, I am unsure when or if this would be addressed. 

Also, no concerns about hardship; it is a great learning exercise. That being said, I don't know where to go from here. Do we kill off the PR or allow the sign-in as designed as a help service to use a password manager for logins?


Derek

AndyTitu:
> > Hi @derekvincent , sorry for the delayed response, but after some internal analysis, we've realised that because the refresh token gets refreshed and written back to the file with each call, the file provisioning approach will not be suitable here. The file we are provisioning is an in-memory FIFO which blocks for all reads and writes. When **we** write to it, it's okay because we have the control over the file, but when another process (in this case the sap btp cli) writes to it, that cannot be predicted in a deterministic way, so the process will just remain blocked.
> > The only way around this would be to provision an environment variable using the env var provisioner, but I'm not sure if such an env var exists. Pursuing this, for now, would mean looking into what environment variables are out there for the sap btp cli and maybe trying to find one called REFRESH_TOKEN. If this does not exist, a long shot would be to ask the developer/support team of SAP to expose one. Apologies for the hardships encountered with the plugin for this specific tool.
> 
> @AndyTitu, sorry for the delay; I did some digging and do not see anything around an env variable to support this. I sent an enhancement request to SAP, but because of the size that SAP is, I am unsure when or if this would be addressed.
> 
> Also, no concerns about hardship; it is a great learning exercise. That being said, I don't know where to go from here. Do we kill off the PR or allow the sign-in as designed as a help service to use a password manager for logins?
> 
> Derek

I have raised the case of this PR internally and we'll come up with a direction asap.

AndyTitu:
I will be closing this for the time being, as the ecosystem is not yet equipped for provisioning regular files.

PR #178, opened 2023-02-12
feat: check for AWS_SHARED_CREDENTIALS_FILE

Volatus:
Check if the environment has set this variable and if so, attempt the
import using that file.

Signed-off-by: Ismayil Mirzali <ismayilmirzeli@gmail.com>


AndyTitu (approved):
LGTM code wise

DCKcode (approved):
Thanks for the contribution! GIven that this configuration is respected by the AWS CLI, it makes sense to respect it here as well.

SimonBarendse (approved):
Thank you for the contribution @Volatus! :heart:

Issue #128, opened 2022-12-24
Separate plugin and executable


osiegmar:
### op CLI version

2.12.0

### Goal or desired behavior

It would be great being able to use plugins with configurable executables.

I'm using Terraform in an AWS environment for example. I'd like to use the aws plugin to authenticate when using Terraform.

Maybe a new flag can be introduced:

```sh
alias aws="op plugin run -- aws"
alias terraform="op plugin run --exec /usr/bin/terraform -- aws"
```

That should also help on https://github.com/1Password/shell-plugins/issues/120

Merry Christmas! 🎅 

### Current behavior

Currently I'm using https://github.com/broamski/aws-mfa which unfortunately lacks 1Password support for obtaining the MFA token.

SimonBarendse:
@scottlamb shared a similar use case in https://github.com/1Password/shell-plugins/pull/180#issuecomment-1453176260:

> The biggest problem I see with this plugin is something else entirely: AFAICT it only runs aws commands, when really there are tons of other programs that use ~/.aws, including e.g. npx aws-cdk and whatever else is written with AWS SDKs. Those don't get the STS provisioner's environment variables, and so setting up ~/.aws to require that environment means they don't work (unless I supply a similar environment by hand / in some other way when invoking them). That's a deal-breaker. I need an easy way to invoke arbitrary commands in this same environment.

PR #180, opened 2023-02-13
Add assume role support for AWS via profiles

AndyTitu:
### Summary
This is a continuation of @shyim 's initiative for [adding AWS Assume role support](https://github.com/1Password/shell-plugins/pull/92) (full credits to @shyim !)

The solution described by this workflow is to use the `~/.aws/config` file we are all used to for setting up various profiles, with different roles. Given that role ARNs are not secrets, users can configure these on disk. Of course, all other sensitive information is recommended to remain inside 1Password:

```
[profile user1]
role_arn=arn:aws:iam::account_id_number:role/role-name-you-want-to-assume

[profile user2]
role_arn=arn:aws:iam::account_id_number:role/another-role-name-you-want-to-assume
```

Then assuming a role is as simple as running `aws s3 ls --profile user1` given that all other credentials were properly configured inside 1P. Assume role can be used with MFA support or without via the shell plugin.

### Challenges
The expiring of the OTP token error (`failed with invalid MFA one time pass code`) seems unavoidable (for 30 seconds - worst case scenario) in the case where a user would quickly switch between `assume role with MFA` and `MFA auth` workflows, but only when cache is not populated for both. This is because we would be now using different cache entries for the credentials of these 2 different workflow, while they both rely on the same OTP code. The alternative is even worse: having both these workflows rely on the same cache entry would result in the user not being able to switch between these 2 workflows for 15 minutes - worst case scenario.

This can be solved in time by adding support for different credential types per plugin.

### DIFF
--- a/plugins/aws/provisioner.go
+++ b/plugins/aws/provisioner.go
@@ -2,6 +2,10 @@ package aws
 
 import (
 	"context"
+	"fmt"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"os"
+	"strings"
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/provision"
@@ -25,9 +29,15 @@ func (p awsProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, ou
 	totp, hasTotp := in.ItemFields[fieldname.OneTimePassword]
 	mfaSerial, hasMFASerial := in.ItemFields[fieldname.MFASerial]
 
-	if hasTotp && hasMFASerial {
+	roleArn := findRoleArnIfSpecified(in, out)
+	if len(out.Diagnostics.Errors) > 0 {
+		return
+	}
+
+	if (hasTotp && hasMFASerial) || roleArn != "" {
 		p.stsProvisioner.MFASerial = mfaSerial
 		p.stsProvisioner.TOTPCode = totp
+		p.stsProvisioner.RoleArn = roleArn
 		p.stsProvisioner.Provision(ctx, in, out)
 	} else {
 		p.envVarProvisioner.Provision(ctx, in, out)
@@ -40,4 +50,58 @@ func (p awsProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput
 
 func (p awsProvisioner) Description() string {
 	return p.envVarProvisioner.Description()
+}
+
+func findRoleArnIfSpecified(in sdk.ProvisionInput, out *sdk.ProvisionOutput) string {
+	for i, arg := range out.CommandLine {
+		if arg == "--profile" {
+			// Read config file from the location set in AWS_CONFIG_FILE env var or from  ~/.aws/config
+			configPath := os.Getenv("AWS_CONFIG_FILE")
+			if configPath != "" {
+				if strings.HasPrefix(configPath, "~") {
+					configPath = in.FromHomeDir(configPath[1:])
+				}
+			} else {
+				configPath = in.FromHomeDir(".aws", "config") // default config file location
+			}
+			contents, err := os.ReadFile(configPath)
+			if os.IsNotExist(err) {
+				out.AddError(fmt.Errorf("you've specified the --profile flag but no profiles to choose from were present at: %s", configPath))
+				return ""
+			} else if err != nil {
+				out.AddError(err)
+				return ""
+			}
+
+			configFile, err := importer.FileContents(contents).ToINI()
+			if err != nil {
+				out.AddError(err)
+				return ""
+			}
+
+			if configFile != nil {
+				if i+1 == len(out.CommandLine) {
+					return ""
+				}
+				profileSection := getConfigSectionByProfile(configFile, out.CommandLine[i+1])
+				if profileSection != nil {
+					if profileSection.HasKey("role_arn") {
+						key, err := profileSection.GetKey("role_arn")
+						if err != nil {
+							out.AddError(err)
+							return ""
+						}
+
+						// remove the --profile flag so the aws cli does not use it
+						out.CommandLine[i] = ""
+						out.CommandLine[i+1] = ""
+
+						return key.Value()
+					}
+				}
+			}
+			break
+		}
+	}
+	return ""
 }
--- a/plugins/aws/provisioner.go
+++ b/plugins/aws/provisioner.go
@@ -93,9 +93,7 @@ func findRoleArnIfSpecified(in sdk.ProvisionInput, out *sdk.ProvisionOutput) str
 						}
 
 						// remove the --profile flag so the aws cli does not use it
-						out.CommandLine[i] = ""
-						out.CommandLine[i+1] = ""
-
+						out.CommandLine = out.CommandLine[0:i]
 						return key.Value()
 					}
 				}
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -14,6 +14,11 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/sts/types"
 )
 
+const (
+	MFACacheKey               = "sts-mfa"
+	assumeRoleWithMFACacheKey = "sts-assume-role-mfa"
+)
+
 type STSProvisioner struct {
 	TOTPCode  string
 	MFASerial string
@@ -21,19 +26,6 @@ type STSProvisioner struct {
 }
 
 func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
-	var cached types.Credentials
-	if ok := in.Cache.Get("sts", &cached); ok {
-		out.AddEnvVar("AWS_ACCESS_KEY_ID", *cached.AccessKeyId)
-		out.AddEnvVar("AWS_SECRET_ACCESS_KEY", *cached.SecretAccessKey)
-		out.AddEnvVar("AWS_SESSION_TOKEN", *cached.SessionToken)
-
-		if region, ok := in.ItemFields[fieldname.DefaultRegion]; ok {
-			out.AddEnvVar("AWS_DEFAULT_REGION", region)
-		}
-
-		return
-	}
-
 	config := aws.NewConfig()
 	config.Credentials = credentials.NewStaticCredentialsProvider(in.ItemFields[fieldname.AccessKeyID], in.ItemFields[fieldname.SecretAccessKey], "")
 
@@ -42,33 +34,100 @@ func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, ou
 		region = os.Getenv("AWS_DEFAULT_REGION")
 	}
 	if len(region) == 0 {
-		out.AddError(errors.New("region is required for the AWS Shell Plugin MFA workflow: set 'default region' in 1Password or set the 'AWS_DEFAULT_REGION' environment variable yourself"))
+		out.AddError(errors.New("region is required for the AWS Shell Plugin MFA or Assume Role workflows: set 'default region' in 1Password or set the 'AWS_DEFAULT_REGION' environment variable yourself"))
 		return
 	}
 	config.Region = region
 
 	stsProvider := sts.NewFromConfig(*config)
-	input := &sts.GetSessionTokenInput{
-		DurationSeconds: aws.Int32(900), // minimum expiration time - 15 minutes
-		SerialNumber:    aws.String(p.MFASerial),
-		TokenCode:       aws.String(p.TOTPCode),
+	var awsTemporaryCredentials *types.Credentials
+
+	if p.RoleArn != "" && p.MFASerial != "" && p.TOTPCode != "" {
+		if useCache := tryUsingCachedCredentials(assumeRoleWithMFACacheKey, in, out); useCache {
+			return
+		}
+
+		input := &sts.AssumeRoleInput{
+			DurationSeconds: aws.Int32(900), // minimum expiration time - 15 minutes
+			SerialNumber:    aws.String(p.MFASerial),
+			TokenCode:       aws.String(p.TOTPCode),
+			RoleArn:         aws.String(p.RoleArn),
+			RoleSessionName: aws.String("1password-shell-plugin"),
+		}
+
+		resp, err := stsProvider.AssumeRole(ctx, input)
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		awsTemporaryCredentials = resp.Credentials
+		err = out.Cache.Put(assumeRoleWithMFACacheKey, *awsTemporaryCredentials, *awsTemporaryCredentials.Expiration)
+		if err != nil {
+			out.AddError(fmt.Errorf("failed to serialize aws sts credentials: %w", err))
+		}
+	} else if p.RoleArn == "" {
+		if useCache := tryUsingCachedCredentials(MFACacheKey, in, out); useCache {
+			return
+		}
+		input := &sts.GetSessionTokenInput{
+			DurationSeconds: aws.Int32(900), // minimum expiration time - 15 minutes
+			SerialNumber:    aws.String(p.MFASerial),
+			TokenCode:       aws.String(p.TOTPCode),
+		}
+
+		resp, err := stsProvider.GetSessionToken(ctx, input)
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		awsTemporaryCredentials = resp.Credentials
+		err = out.Cache.Put(MFACacheKey, *awsTemporaryCredentials, *awsTemporaryCredentials.Expiration)
+		if err != nil {
+			out.AddError(fmt.Errorf("failed to serialize aws sts credentials: %w", err))
+		}
+	} else {
+		input := &sts.AssumeRoleInput{
+			DurationSeconds: aws.Int32(900), // minimum expiration time - 15 minutes
+			RoleArn:         aws.String(p.RoleArn),
+			RoleSessionName: aws.String("1password-shell-plugin"),
+		}
+
+		resp, err := stsProvider.AssumeRole(ctx, input)
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		awsTemporaryCredentials = resp.Credentials
 	}
 
-	result, err := stsProvider.GetSessionToken(ctx, input)
-	if err != nil {
-		out.AddError(err)
-		return
-	}
-
-	out.AddEnvVar("AWS_ACCESS_KEY_ID", *result.Credentials.AccessKeyId)
-	out.AddEnvVar("AWS_SECRET_ACCESS_KEY", *result.Credentials.SecretAccessKey)
-	out.AddEnvVar("AWS_SESSION_TOKEN", *result.Credentials.SessionToken)
+	out.AddEnvVar("AWS_ACCESS_KEY_ID", *awsTemporaryCredentials.AccessKeyId)
+	out.AddEnvVar("AWS_SECRET_ACCESS_KEY", *awsTemporaryCredentials.SecretAccessKey)
+	out.AddEnvVar("AWS_SESSION_TOKEN", *awsTemporaryCredentials.SessionToken)
 	out.AddEnvVar("AWS_DEFAULT_REGION", region)
 
-	err = out.Cache.Put("sts", *result.Credentials, *result.Credentials.Expiration)
+	err := out.Cache.Put("sts", *awsTemporaryCredentials, *awsTemporaryCredentials.Expiration)
 	if err != nil {
 		out.AddError(fmt.Errorf("failed to serialize aws sts credentials: %w", err))
 	}
+}
+
+func tryUsingCachedCredentials(cacheKey string, in sdk.ProvisionInput, out *sdk.ProvisionOutput) bool {
+	var cached types.Credentials
+	if ok := in.Cache.Get(cacheKey, &cached); ok {
+		out.AddEnvVar("AWS_ACCESS_KEY_ID", *cached.AccessKeyId)
+		out.AddEnvVar("AWS_SECRET_ACCESS_KEY", *cached.SecretAccessKey)
+		out.AddEnvVar("AWS_SESSION_TOKEN", *cached.SessionToken)
+
+		if region, ok := in.ItemFields[fieldname.DefaultRegion]; ok {
+			out.AddEnvVar("AWS_DEFAULT_REGION", region)
+		}
+
+		return true
+	}
+	return false
 }
 
 func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput, out *sdk.DeprovisionOutput) {
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -17,6 +17,7 @@ import (
 type STSProvisioner struct {
 	TOTPCode  string
 	MFASerial string
+	RoleArn   string
 }
 
 func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {


AndyTitu:
This might require security review because it introduces external functions for handling `.aws/config` files and `AWS_` env vars.

plugins/aws/provisioner.go
```diff
@@ -2,42 +2,155 @@ package aws
...
-	// Nothing to do here: environment variables get wiped automatically when the process exits.
+	p.selectedProvisioner.Deprovision(ctx, in, out)
 }
 
 func (p awsProvisioner) Description() string {
-	return p.envVarProvisioner.Description()
+	if p.selectedProvisioner != nil {
+		return p.selectedProvisioner.Description()
+	}
+	return "Provisions AWS with temporary credentials."
```
AndyTitu, line 39: I am not sure where the `Description` function is supposed to be called. If this function is meant to be called always before `Provision` this line is redundant because `Provision` is the one that sets the `selectedProvisioner`

AndyTitu:
@shyim Feel free to review this PR. After all, it all started from your initiative 😄 . I am especially interested if this is working as you would expect.

ahamez:
@AndyTitu Thanks for the follow-up on #92!

If it's as easy as using the profiles in `~/.aws/config`, then 👍👍👍👍 😁

plugins/aws/provisioner.go
```diff
@@ -2,42 +2,155 @@ package aws
...
+	}
+
+	if hasTotp && mfaSerial != "" {
+		return newMFAProvisioner(mfaSerial, totp, awsConfig.Region, awsConfig.NonChainedGetSessionTokenDuration)
+	}
+
+	return provision.EnvVarProvisioner{
+		Schema: defaultEnvVarMapping,
+	}
+}
```
florisvdg, line 127: I don't really see the value of the split between the MFA provisioner and role provisioner. It probably makes more sense to merge those two into a single `STSProvisioner` which takes an optional `profile` arg, and then have an `AWSCLIProvisioner` as a lightweight wrapper that captures the `--profile` flag and passes it down to the `STSProvisioner`. What do you think?

plugins/aws/mfa_provisioner.go
```diff
@@ -41,39 +48,59 @@ func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, ou
...
+
+		return true
+	}
+	return false
 }
 
-func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput, out *sdk.DeprovisionOutput) {
-	// Nothing to do here: environment variables get wiped automatically when the process exits.
+func (p mfaProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput, out *sdk.DeprovisionOutput) {
+	// nothing to do here: 1Password CLI removes env vars
```
SimonBarendse, line 101: This sounds to me like 1Password CLI is actively pruning the envvars. This isn't what's happening, instead, they are wiped by the operating system when the process exits. For that reason, I liked the previous version of this comment better.

plugins/aws/provisioner.go
```diff
@@ -2,42 +2,155 @@ package aws
...
+	return configuration
+}
+
+func selectProvisionerBasedOnPresentValues(itemFields map[sdk.FieldName]string, awsConfig *confighelpers.Config, out *sdk.ProvisionOutput) sdk.Provisioner {
+	totp, hasTotp := itemFields[fieldname.OneTimePassword]
+	mfaSerial, hasMFASerial := itemFields[fieldname.MFASerial]
+
+	// Give priority to the mfa serial specified in 1Password
+	if hasMFASerial && awsConfig.HasMfaSerial() && awsConfig.MfaSerial != mfaSerial {
+		out.AddError(fmt.Errorf("your local AWS configuration has a different MFA serial than the one specified in 1Password"))
```
SimonBarendse, line 110: It feels inconsistent that [we use the region from the 1Password item](https://github.com/1Password/shell-plugins/pull/180/files#diff-4899eeb666d138152378c9be8b70bcee7287a247ab1e504a3385044550017794R56) when both the config file and the item in 1Password have a region specified, but when both the config and the 1Password item have an MFA serial defined, there is no conflict resolution and instead an error is thrown.

To make the product work intuitively, we need the behavior to be consistent. I have a preference for raising an error in both situations, as I don't see a reason why a user would want to have different definitions in both places, so it's probably a user error. Having the product throw an error in this case is helpful, because it allows the user to fix the conflict and choose which of the two they actually meant to use.

plugins/aws/provisioner.go
```diff
@@ -2,42 +2,155 @@ package aws
...
+
+func (p *awsProvisioner) selectCorrespondingProvisioner(itemFields map[sdk.FieldName]string, out *sdk.ProvisionOutput) sdk.Provisioner {
+	awsConfig := getAWSConfiguration(out)
+	if len(out.Diagnostics.Errors) > 0 {
+		return nil
+	}
+
+	err := errorIfUnsupportedConfigurationIsPresent(awsConfig)
+	if err != nil {
+		out.AddError(err)
```
SimonBarendse, line 50: These error messages don't mention that the unsupported config was detected in the AWS config file. Can we assume users are aware/remember that they've configured it there, or would it be worth mentioning that here in the error message?

plugins/aws/provisioner.go
SimonBarendse: I like this suggestion. I think it would clean up this implementation

scottlamb:
> The expiring of the OTP token error (failed with invalid MFA one time pass code) seems unavoidable (for 30 seconds - worst case scenario) in the case where a user would quickly switch between assume role with MFA and MFA auth workflows, but only when cache is not populated for both. This is because we would be now using different cache entries for the credentials of these 2 different workflow, while they both rely on the same OTP code. The alternative is even worse: having both these workflows rely on the same cache entry would result in the user not being able to switch between these 2 workflows for 15 minutes - worst case scenario - as the same tokens from cache will be used.

Similarly: with this PR, won't you also see this failed with invalid MFA one time pass code failure when assuming several roles in rapid succession, because it's sending the TOTP code for each? I've hit that problem with plain `aws` (no 1Password plugin), which I believe is doing essentially the same thing.

I just did a quick experiment, and it looks like if instead you create temporary credentials with GetSessionToken using an MFA, you can use those temporary credentials with N AssumeRoles without supplying a new MFA, and `"aws:MultiFactorAuthPresent": "true"` conditions on `AssumeRole` trust relationships will be satisfied. So I think it's much a better experience to get/cache a single GetSessionToken using mfa and then pass it to each AssumeRole as necessary. Additionally caching the AssumeRole credentials obtained with it would avoid the extra round trips which is nice but not a huge deal.

In fact, this seems to work already via the setup in [this comment](https://github.com/1Password/shell-plugins/pull/92#issuecomment-1420870019) with `credential_source = Environment` in each profile. So I think this PR is a regression: going from a setup that (in my quick experiments) appears to work well for `aws --profile ...` to one that has the 30 second problem.

The biggest problem I see with this plugin is something else entirely: AFAICT it only runs `aws` commands, when really there are tons of other programs that use `~/.aws`, including e.g. `npx aws-cdk` and whatever else is written with AWS SDKs. Those don't get the STS provisioner's environment variables, and so setting up `~/.aws` to require that environment means they don't work (unless I supply a similar environment by hand / in some other way when invoking them). That's a deal-breaker. I need an easy way to invoke arbitrary commands in this same environment.

plugins/aws/provisioner.go
AndyTitu: That configuration can also be from env vars not only from config file. I will state both possibilities for clarity.

SimonBarendse:
Thank you @scottlamb for sharing your thoughts and catching this regression!

> The biggest problem I see with this plugin is something else entirely: AFAICT it only runs aws commands, when really there are tons of other programs that use ~/.aws, including e.g. npx aws-cdk and whatever else is written with AWS SDKs. Those don't get the STS provisioner's environment variables, and so setting up ~/.aws to require that environment means they don't work (unless I supply a similar environment by hand / in some other way when invoking them). That's a deal-breaker. I need an easy way to invoke arbitrary commands in this same environment.

Thank you for sharing your use case! We'll follow this up in https://github.com/1Password/shell-plugins/issues/128.

plugins/aws/provisioner.go
AndyTitu: Sounds good. I am happy about having this as a global rule for resolving this type of conflicts!

plugins/aws/provisioner.go
AndyTitu: Refactored accordingly, thanks!

plugins/aws/mfa_provisioner.go
AndyTitu: Alright, makes sense, done!

AndyTitu:
> In fact, this seems to work already via the setup in https://github.com/1Password/shell-plugins/pull/92#issuecomment-1420870019 with credential_source = Environment in each profile. So I think this PR is a regression: going from a setup that (in my quick experiments) appears to work well for aws --profile ... to one that has the 30 second problem.

We have now fixed this in this PR as well by using different cache keys for each role. Now the cache key is computed as `"assume-role-sts" + roleARN`.

plugins/aws/provisioner.go
AndyTitu: not applicable anymore

scottlamb:
> We fixed this in this PR as well by using different cache keys for each role. Now the cache key is computed as "assume-role-sts" + roleARN.

That doesn't fix it. You only get one TOTP every 30 seconds. If you do the following:

```
role_a_tmp_creds = assume_role(role=a, token_code=12345, mfa_serial=mfa_serial, ...permanent_creds)
role_b_tmp_creds = assume_role(role=b, token_code=12345, mfa_serial=mfa_serial, ...permanent_creds)
```

the second operation that passes `12345` will fail with `An error occurred (AccessDenied) when calling the AssumeRole operation: MultiFactorAuthentication failed with invalid MFA one time pass code.` If you need to pass a token code more than once, you have to wait across a 30-second step boundary.

In contrast, this will succeed because it only passes `12345` once. Apparently the role credentials are still considered to have a MFA; that property is transitive:

```
base_tmp_creds = get_session_token(token_code=12345, ...permanent_creds)
role_a_tmp_creds = assume_role(role=a, ...base_tmp_creds)
role_b_tmp_creds = assume_role(role=b, ...base_tmp_creds)
```

AndyTitu:
@scottlamb thank you for the further insights! I gave it some more thought myself and this makes sense indeed! After some internal discussion we decided to move forward with the solution proposed by you for assuming a role with MFA.

plugins/aws/providers.go
```diff
@@ -0,0 +1,142 @@
...
+			RoleSessionName:   awsConfig.RoleSessionName,
+			ExternalID:        awsConfig.ExternalID,
+			Duration:          900 * time.Second, // minimum duration of 15 minutes
+			Tags:              awsConfig.SessionTags,
+			TransitiveTagKeys: awsConfig.TransitiveSessionTags,
+			SourceIdentity:    awsConfig.SourceIdentity,
+			Mfa:               &confighelpers.Mfa{},
+		},
+		stsCacheWriter: NewStsCacheWriter(roleCacheKey, out.Cache),
+	}
```
SimonBarendse, line 70: There is a lot of duplication between these two returns which:
1. Makes it hard to see what's different
2. Makes it easier for unintended inconsistencies to be introduced in the future.

plugins/aws/providers.go
```diff
@@ -0,0 +1,142 @@
...
+	return credentials, nil
+}
+
+func NewAssumeRoleProvider(awsConfig *confighelpers.Config, in sdk.ProvisionInput, out *sdk.ProvisionOutput) aws.CredentialsProvider {
+	roleCacheKey := getRoleCacheKey(awsConfig.RoleARN)
+	if in.Cache.Has(roleCacheKey) {
+		return NewStsCacheProvider(roleCacheKey, in.Cache)
+	}
+
+	if awsConfig.HasMfaSerial() && awsConfig.MfaToken != "" {
```
SimonBarendse, line 40: What if there is a serial but no token? Or vice-versa? I think we'd want to return an error in this case to make it easier to debug for the user. As when they have either set, they clearly want to use MFA, so I believe falling back to not using MFA would make things harder for these users.

SimonBarendse (commented):
I like the general direction where this is heading. :raised_hands: So I've started looking more to the details as well and left a bunch of comments about that for your consideration.

plugins/aws/cli_provisioner.go
```diff
@@ -0,0 +1,58 @@
...
+func NewAwsCLIProvisioner() sdk.Provisioner {
+	return awsCLIProvisioner{}
+}
+
+func (p awsCLIProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
+	profileName := defaultProfileName
+
+	if profile := stripAndReturnProfileFlag(out); profile != nil {
+		profileName = *profile
+	} else if profile := os.Getenv("AWS_PROFILE"); profile != "" {
```
florisvdg, line 25: I don't think `AWS_PROFILE` is CLI-exclusive, so we should probably move that one down to the STS provisioner.

plugins/aws/providers.go
```diff
@@ -0,0 +1,142 @@
...
+	err = p.stsCacheWriter.persist(credentials)
+	if err != nil {
+		return aws.Credentials{}, err
+	}
+
+	return credentials, nil
+}
+
+func NewAssumeRoleProvider(awsConfig *confighelpers.Config, in sdk.ProvisionInput, out *sdk.ProvisionOutput) aws.CredentialsProvider {
+	roleCacheKey := getRoleCacheKey(awsConfig.RoleARN)
```
jpcoenen, line 35: Is this unique enough? What if the user has 2 credentials that can assume the same role? They would lead to different assumed roles, but would share 1 cache entry.

plugins/aws/providers.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	if awsConfig.HasMfaSerial() && awsConfig.MfaToken != "" {
+		return &assumeRoleProvider{
+			AssumeRoleProvider: confighelpers.AssumeRoleProvider{
+				StsClient:         getSTSClient(awsConfig.Region, NewMFASessionTokenProvider(awsConfig, in, out)),
+				RoleARN:           awsConfig.RoleARN,
+				RoleSessionName:   awsConfig.RoleSessionName,
+				ExternalID:        awsConfig.ExternalID,
+				Duration:          900 * time.Second, // minimum duration of 15 minutes
```
jpcoenen, line 47: Can we dictate this? For example, what if the user wants to run `aws logs tail --follow <log-group>`? That command can run for more than 15 minutes.

Could we use the session duration from the config instead to allow the user to configure this?

plugins/aws/providers.go
```diff
@@ -0,0 +1,142 @@
...
+
+func NewMFASessionTokenProvider(awsConfig *confighelpers.Config, in sdk.ProvisionInput, out *sdk.ProvisionOutput) aws.CredentialsProvider {
+	if in.Cache.Has(mfaCacheKey) {
+		return NewStsCacheProvider(mfaCacheKey, in.Cache)
+	}
+
+	return &mfaSessionTokenProvider{
+		SessionTokenProvider: confighelpers.SessionTokenProvider{
+			StsClient: getSTSClient(awsConfig.Region, NewMasterCredentialsProvider(in.ItemFields)),
+			Duration:  900 * time.Second, // minimum duration of 15 minutes,
```
jpcoenen, line 100: I've seen this pop up in 3 places now. Let's make sure to use one source of truth for this.

plugins/aws/cli_provisioner.go
```diff
@@ -0,0 +1,58 @@
...
+		profileName = profile
+	}
+	NewSTSProvisioner(profileName).Provision(ctx, in, out)
+}
+
+func stripAndReturnProfileFlag(out *sdk.ProvisionOutput) *string {
+	for i, arg := range out.CommandLine {
+		if arg == "--profile" {
```
jpcoenen, line 33: I am not sure if the AWS CLI uses this anywhere, but if `--profile` is used after `--`, it should not be interpreted as a flag by the AWS CLI. Just like how in `op run -- aws --profile <profile>` `--profile` is not interpreted as a flag for `op`, but as an argument instead.

plugins/aws/providers.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func NewStsCacheProvider(key string, cache sdk.CacheState) aws.CredentialsProvider {
+	return stsCacheProvider{
+		awsCacheKey: key,
+		cache:       cache,
+	}
+}
+
+type masterAwsCredentialsProvider struct {
```
jpcoenen, line 114: We generally try to steer away from using "master" because of its relation to slavery. 

Can we use "AWS credentials" or "Access Keys" (that's what AWS uses for that keypair) instead?

florisvdg: Also, there is no such thing as a "master AWS credential" as far as I'm aware.

plugins/aws/cli_provisioner.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (p awsCLIProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
+	profileName := defaultProfileName
+
+	if profile := stripAndReturnProfileFlag(out); profile != nil {
+		profileName = *profile
+	} else if profile := os.Getenv("AWS_PROFILE"); profile != "" {
+		profileName = profile
+	}
+	NewSTSProvisioner(profileName).Provision(ctx, in, out)
```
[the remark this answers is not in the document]
florisvdg: We'll need it to support other tools that use AWS (SDK), like Terraform. We don't want to look for the `--profile` flag in those tools.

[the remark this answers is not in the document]
jpcoenen: Ah, okay, that explains why `NewSTSProvisioner()` exists :-) In that case it becomes more of a nitpick, but it'd still be nicer to not construct a full provisioner if we are only interested in its `Provision()` method.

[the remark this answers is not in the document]
AndyTitu: I get it that line 28 looks weird for now, but the `STSProvisioner` is meant to be used in other parts of the codebase as well, pretty soon, in places where the contained functionality will need to be passed specifically as a provisioner.

To be more specific, in the case of terraform, when the terraform shell plugin knows it needs to provision AWS, it will go directly for the `STSProvisioner` and not the `AwsCliProvisioner`. And for re-usability, we have have `AwsCliProvisioner` rely on `StsProvisioner` logic. Does this make sense? Do you think there could be another way of doing it?

plugins/aws/providers.go
AndyTitu: I saw this get called like that in `aws-vault` (https://github.com/99designs/aws-vault/blob/b15cc22022184d7835311e77cd59b720e6007876/vault/vault.go#L243). I'll rename it to `accessKeysProvider`

plugins/aws/cli_provisioner.go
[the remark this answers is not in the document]
AndyTitu: Haha, I've never seen that second usage before but apparently `aws "--profile" my-profile` does work with the current implementation which means you were right, the command line parser takes care of that.

As for the first usage: do we want to support that as well, or do we want to leave that as an 'escape route' for users, i.e. if they want to use a profile without the shell-plugin and we document that, or is that too much? 
@jpcoenen @florisvdg

plugins/aws/providers.go
AndyTitu: Plugin cache entries are specific to a certain item. In the case the user specifies the other pair of credentials in another item, we are covered. However, in the (arguably rare) case where a user would just rotate their access keys in the same item, we could indeed run into cache collisions. We could use the ACCESS_KEY_ID value to further isolate entries in this case. ( Access key ids are not sensitive by themselves, but we shouldn't even care about that because these plugin defined keys are encrypted anyway)

AndyTitu: Introduced Access Key Id as to further differentiate cache keys both for assume role and mfa

plugins/aws/cli_provisioner.go
AndyTitu: Correct. Done, thanks!

plugins/aws/providers.go
AndyTitu: Yes, that's right. Added that extra check now. thanks!

plugins/aws/providers.go
AndyTitu: Done, thanks!

plugins/aws/sts_provisioner.go
```diff
@@ -2,78 +2,146 @@ package aws
...
-func (p STSProvisioner) Description() string {
+func (p stsProvisioner) Description() string {
 	return "Provision environment variables with temporary STS credentials AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN"
 }
+
+// chooseTemporaryCredentialsProvider returns the aws provider that fits the scenario described by the current configuration, alongside the corresponding stsCacheWriter for encrypting temporary credentials to disk to be used in next runs.
+func (p *stsProvisioner) chooseTemporaryCredentialsProvider(awsConfig *confighelpers.Config, in sdk.ProvisionInput, out *sdk.ProvisionOutput) aws.CredentialsProvider {
+	unsupportedMessage := "%s is not yet supported by the AWS Shell Plugin"
+	if awsConfig.HasSSOStartURL() {
+		out.AddError(fmt.Errorf(unsupportedMessage, "SSO Authentication"))
```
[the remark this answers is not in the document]
AndyTitu: I have changed this function (and others) so that they return an error, to avoid such mistakes.

plugins/aws/sts_provisioner.go
```diff
@@ -2,78 +2,146 @@ package aws
...
+	return configuration
+}
+
+// resolveLocalAnd1PasswordConfigurations goes over configurations present in both local settings and 1Password and resolves conflicts.
+func resolveLocalAnd1PasswordConfigurations(itemFields map[sdk.FieldName]string, awsConfig *confighelpers.Config, out *sdk.ProvisionOutput) {
+	mfaSerial, hasMFASerial := itemFields[fieldname.MFASerial]
+	totp, hasOTP := itemFields[fieldname.OneTimePassword]
+	region, hasRegion := itemFields[fieldname.DefaultRegion]
+
+	// Give priority to the mfa serial specified in 1Password
```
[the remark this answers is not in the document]
AndyTitu: Yep good catch, changed that now, thanks!

plugins/aws/providers.go
AndyTitu: Done, thanks!

plugins/aws/providers.go
AndyTitu: Done

plugins/aws/providers.go
AndyTitu: I did not think of such a use case. Indeed it makes sense to allow the user to override this. I still kept the 15 minutes expiration (the minimum allowed one) as a default in case none is specified because this is more secure than the 1 hour default expiration of AWS.

AndyTitu: Done

plugins/aws/cli_provisioner.go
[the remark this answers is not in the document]
florisvdg: I'm all for offering an easy one-off opt-out for shell plugins, but the way to trigger that should definitely _not_ be a magic equals sign. 😅

AndyTitu: Makes sense, handled that case now!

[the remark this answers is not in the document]
AndyTitu: Alright, makes sense, now we also support the `--profile=` syntax

plugins/aws/cli_provisioner.go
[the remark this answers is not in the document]
AndyTitu: Actually I changed the code now so it makes use of the provisioners as I explained.

plugins/aws/providers.go
jpcoenen: Ah, good point about it being linked to the item. I was not actively aware of that. That indeed makes this a lot less problematic.

Technically, the Access Key ID is a bit too over-specific since it can change even if the user does not change, but it should work fine in practice and make it slightly more robust :+1: After all, it's not a problem if we drop all cache entries if someone rotates their access key.

plugins/aws/cli_provisioner.go
```diff
@@ -25,22 +26,36 @@ func (p awsCLIProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput,
 
 func stripAndReturnProfileFlag(args *[]string) (string, error) {
```
jpcoenen, line 28: This is somewhat of a matter of taste, but passing the args as a pointer makes it somewhat difficult to figure out what is happenign with them. Can we instead just return the modified slice? Three return values is also not super nice, but since the return value is different from the profile, it's not likely that the profile and args return value will get mxied up.

plugins/aws/cli_provisioner.go
```diff
@@ -25,22 +26,36 @@ func (p awsCLIProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput,
...
 			if i+1 == len(*args) {
 				return "", fmt.Errorf("--profile flag was specified without a value")
 
 			}
 			profile := (*args)[i+1]
 			// Remove --profile flag so aws cli doesn't attempt to assume role by itself
-			*args = removeProfileFlagFromArgs(i, *args)
+			*args = removeArgSequenceFromArgList(i, i+1, *args)
+			return profile, nil
+		} else if strings.Contains(arg, "--profile=") {
```
jpcoenen, line 44: Let's use `strings.HasPrefix()`. Otherwise `"This is a story about the --profile= flag"` is also accepted.

plugins/aws/cli_provisioner.go
```diff
@@ -25,22 +26,36 @@ func (p awsCLIProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput,
...
+			split := strings.Split(arg, "=")
+			if len(split) != 2 {
+				return "", fmt.Errorf("--profile flag was specified without a value")
+			}
+			profile := split[1]
+			// Remove --profile flag so aws cli doesn't attempt to assume role by itself
```
jpcoenen, line 50: Mostly as an FYI: Go also has a specific function for this:
```go
profile := strings.TrimPrefix(arg, "--profile=")
if profile == "" {
    return "", fmt.Errorf("--profile flag was specified without a value")
}
```

Feel free to leave it as-is though, the current implementation works just fine :)

plugins/aws/cli_provisioner.go
```diff
@@ -25,22 +26,36 @@ func (p awsCLIProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput,
...
 	}
 	return "", nil
 }
 
-func removeProfileFlagFromArgs(argIndex int, args []string) []string {
-	result := append(args[0:argIndex], args[argIndex+2:]...)
+func removeArgSequenceFromArgList(startIndex, endIndex int, args []string) []string {
+	result := append(args[0:startIndex], args[endIndex+1:]...)
 	return result
 }
```
jpcoenen: nit: While I appreciate pulling this out into a separate function. Because of the relatively complex arguments, the code might actually become easier to read if we perform the `append()` inline.

plugins/aws/cache_utils.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return aws.Credentials{}, fmt.Errorf("did not find cached credentials")
+}
+
+func getRoleCacheKey(roleArn string, accessKeyID string) string {
+	return assumeRoleCacheKeyID + accessKeyID + roleArn
+}
+
+func getMfaCacheKey(accessKeyID string) string {
+	return mfaCacheKeyID + accessKeyID
```
jpcoenen, line 52: nit: it's generally a good practice to use a separatator between these elements. In this case it is not likely to lead to any problems because of the specific format of the Access Key ID and Role ARN, but at least it makes the cache keys easier to interpret.

For example:
```
fmt.Sprintf("%s|%s|%s", assumeRoleCacheKeyID, accessKeyID, roleArn)
```

jpcoenen (changes requested):
Seeing quite a lot of improvement 👌 

Most of these comments are just some nitpicks and are not blocking. Only the thing about the number of arguments of `NewSTSProvisioner` and the use of `strings.Contains()` are blocking imo.

plugins/aws/cli_provisioner.go
AndyTitu: Makes sense, done, thanks!

plugins/aws/cache_utils.go
AndyTitu: Done!

plugins/aws/sts_provisioner.go
```diff
@@ -2,78 +2,168 @@ package aws
...
+	}
 
-		return
+	if profile := os.Getenv("AWS_PROFILE"); profile != "" {
+		return stsProvisioner{profileName: profile}
 	}
 
-	config := aws.NewConfig()
-	config.Credentials = credentials.NewStaticCredentialsProvider(in.ItemFields[fieldname.AccessKeyID], in.ItemFields[fieldname.SecretAccessKey], "")
+	return stsProvisioner{profileName: defaultProfileName}
```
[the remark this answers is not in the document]
AndyTitu: Okay, the init functions are a bit redundant. Refactored as advised!

plugins/aws/sts_provisioner.go
```diff
@@ -2,78 +2,146 @@ package aws
...
 }
 
-func (p STSProvisioner) Description() string {
+func (p stsProvisioner) Description() string {
 	return "Provision environment variables with temporary STS credentials AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN"
 }
+
+// chooseTemporaryCredentialsProvider returns the aws provider that fits the scenario described by the current configuration, alongside the corresponding stsCacheWriter for encrypting temporary credentials to disk to be used in next runs.
+func (p *stsProvisioner) chooseTemporaryCredentialsProvider(awsConfig *confighelpers.Config, in sdk.ProvisionInput, out *sdk.ProvisionOutput) aws.CredentialsProvider {
+	unsupportedMessage := "%s is not yet supported by the AWS Shell Plugin"
```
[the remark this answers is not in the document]
AndyTitu: I have created issues for all these authentication methods and I've linked them in the error messages alongside an invitation to take a look! Thanks for the suggestion!

plugins/aws/providers.go
```diff
@@ -0,0 +1,143 @@
+package aws
```
florisvdg, line 1: I'd be more in favor of putting these "provider" util functions at the bottom of `sts_provisioner.go` to make that more of an implementation detail. By having `providers.go` in this package, it looks like we're introducing yet another concept in the shell plugins SDK (plugins, provisioners, importers, providers, etc.), which is not the case.

plugins/aws/sts_provisioner.go
```diff
@@ -2,78 +2,171 @@ package aws
...
-}
+const defaultProfileName = "default"
 
-func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
-	var cached types.Credentials
-	if ok := in.Cache.Get("sts", &cached); ok {
-		out.AddEnvVar("AWS_ACCESS_KEY_ID", *cached.AccessKeyId)
-		out.AddEnvVar("AWS_SECRET_ACCESS_KEY", *cached.SecretAccessKey)
-		out.AddEnvVar("AWS_SESSION_TOKEN", *cached.SessionToken)
+type StsProvisioner struct {
```
florisvdg, line 16: Nit: `STSProvisioner` would be more Go'esque.

plugins/aws/providers.go
AndyTitu: Done, thanks!

plugins/aws/sts_provisioner.go
```diff
@@ -77,3 +63,260 @@ func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput
...
+
+// getProfile returns the profile to be used on this run based on specified profile information
+func (p STSProvisioner) getProfile() string {
+	if len(p.profileName) != 0 {
+		return p.profileName
+	}
+
+	if profile := os.Getenv("AWS_PROFILE"); profile != "" {
+		return profile
+	}
```
jpcoenen, line 75: Shouldn't we strip this environment variable if it is being used? Is there anything that stops the AWS CLI from trying to re-assume a role using this profile?

jpcoenen:
Code seems to be in a good state now :+1: 

However, I only just noticed that we're not adding any tests in PR, while there is quite a bit of new logic. Are there any plans to add some tests to confirm that everything is working correctly now and we can easily confirm things are still functional in the future?

plugins/aws/sts_provisioner.go
```diff
@@ -77,3 +63,260 @@ func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput
...
+func (c stsCacheProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
+	var cached aws.Credentials
+	if ok := c.cache.Get(c.awsCacheKey, &cached); ok {
+		return cached, nil
+	}
+
+	return aws.Credentials{}, fmt.Errorf("did not find cached credentials")
+}
+
+func NewStsCacheProvider(key string, cache sdk.CacheState) aws.CredentialsProvider {
```
florisvdg, line 286: Same comment as before about `NewSTSCacheProvider` being more Go'esque.

plugins/aws/cache_utils.go
```diff
@@ -0,0 +1,38 @@
...
+	assumeRoleCacheKeyID = "sts-assume-role"
+)
+
+// stsCacheWriter writes aws temp credentials to cache using the awsCacheKey
+type stsCacheWriter struct {
+	awsCacheKey string
+	cache       sdk.CacheOperations
+}
+
+func (c stsCacheWriter) persist(credentials aws.Credentials) error {
```
florisvdg, line 21: Nit: Could be nicer to have these cache wrappers use the same terminology as the underlying cache: `Put`/`Get`.

plugins/aws/sts_provisioner.go
AndyTitu: Ooh, that's right, very good catch!

AndyTitu:
I have now added unit tests for functions: `resolveLocalAnd1PasswordConfigurations`, `getProfile` and `stripAndReturnProfileFlag`. I have also uncovered some uncovered edge cases while doing so and I've fixed them.

plugins/aws/sts_provisioner.go
```diff
@@ -77,3 +67,270 @@ func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput
...
+func (p STSProvisioner) getProfile() (string, error) {
+	if len(p.profileName) != 0 {
+		return p.profileName, nil
+	}
+
+	if profile := os.Getenv("AWS_PROFILE"); profile != "" {
+		err := os.Unsetenv("AWS_PROFILE")
+		if err != nil {
+			return "", err
+		}
```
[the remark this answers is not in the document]
jpcoenen: We could consider moving this to a separate issue as it probably does not affect the plugin as included in the CLI.

@AndyTitu Have you tested if it does work if the plugin is compiled into the CLI?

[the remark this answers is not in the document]
jpcoenen: I did some experimentation and it seems that the AWS CLI ignores `AWS_PROFILE` if it is provided with temporary credentials. That means that unsetting it is not strictly required. I am not sure if that also applies to all AWS SDK's (and therefore tools build on top of them), but it does mean we can leave out this piece as long as only the AWS CLI is supported.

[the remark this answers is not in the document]
AndyTitu: Also tested this out now, and indeed we can just remove the unsetting logic here.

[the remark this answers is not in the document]
AndyTitu: Done!

AndyTitu:
> However, I only just noticed that we're not adding any tests in PR, while there is quite a bit of new logic. Are there any plans to add some tests to confirm that everything is working correctly now and we can easily confirm things are still functional in the future?

@jpcoenen I have now added some extra end-to-end tests for the case in which we are using AssumeRole, MFA and AssumeRole with MFA providers, with the purpose of ensuring in the future that all our use cases will continue to work as expected.

plugins/aws/sts_provisioner.go
```diff
@@ -16,10 +16,33 @@ import (
 const defaultProfileName = "default"
 
 type STSProvisioner struct {
-	profileName string
+	profileName     string
+	providerFactory STSProviderFactory
```
jpcoenen, line 20: 🙌 for adding the tests that use this. 

A small suggestion to simplify the interface a bit:
```suggestion
    newProviderFactory func(cacheState sdk.CacheState, cache sdk.CacheOperations, fields map[sdk.FieldName]string) STSProviderFactory
```

I think that allows the removal of the `SetState()` method and thereby makes it impossible to misuse the interface (calling `NewAssumeRoleProvider()` before `SetState()` is illegal in the current form).

However, I don't see any big issues with the current form. So you're also free to leave it as-is.

plugins/aws/sts_provisioner.go
```diff
@@ -77,3 +90,273 @@ func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput
...
+}
+
+func (m CacheProviderFactory) NewMFASessionTokenProvider(awsConfig *confighelpers.Config) aws.CredentialsProvider {
+	mfaCacheKey := getMfaCacheKey(m.ItemFields[fieldname.AccessKeyID])
+	if m.InCache.Has(mfaCacheKey) {
+		return NewStsCacheProvider(mfaCacheKey, m.InCache)
+	}
+
+	if awsConfig.NonChainedGetSessionTokenDuration == 0 {
+		awsConfig.NonChainedGetSessionTokenDuration = 900 * time.Second // default to minimum duration of 15 minutes for security
```
jpcoenen, line 169: super nit: we have `time.Minute` and I am not a fan of "for security" as it does not really explain any reasoning.
```suggestion
		// Set a low session TTL by default. 
		// 15 minutes should be enough to complete most operations.
		awsConfig.NonChainedGetSessionTokenDuration = 15 * time.Minute
```

jpcoenen (dismissed):
AWeSome work, @AndyTitu! Code looks good to me 👌 Left two tiny suggestions, but I am happy to merge it as-is.

plugins/aws/sts_provisioner.go
AndyTitu: Sounds good, changed to minutes and removed the "for security"

jpcoenen (approved):
Still approved! 🚀

plugins/aws/sts_provisioner.go
```diff
@@ -2,71 +2,89 @@ package aws
...
-func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
-	var cached types.Credentials
-	if ok := in.Cache.Get("sts", &cached); ok {
-		out.AddEnvVar("AWS_ACCESS_KEY_ID", *cached.AccessKeyId)
-		out.AddEnvVar("AWS_SECRET_ACCESS_KEY", *cached.SecretAccessKey)
-		out.AddEnvVar("AWS_SESSION_TOKEN", *cached.SessionToken)
+func NewSTSProvisioner(profileName string) STSProvisioner {
+	return STSProvisioner{
+		profileName: profileName,
+		newProviderFactory: func(cacheState sdk.CacheState, cacheOps sdk.CacheOperations, fields map[sdk.FieldName]string) STSProviderFactory {
```
SimonBarendse, line 26: Do I interpret this name correctly that we have a factory for a factory?

My current understanding based on the names:
- A providerfactory returns a provider.
- The provider in turn provides a credential

One point of confusion is that the return type of this function is also a provider factory. Based on the fieldname, I'd expect the return type to be a provider, as it's what the providerfactory (the func) returns.

go.sum
```diff
@@ -43,8 +56,10 @@ github.com/golang/protobuf v1.3.2/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5y
 github.com/golang/protobuf v1.3.4 h1:87PNWwrRvUSnqS4dlcBU/ftvOIBep4sYuBLlh6rX2wk=
 github.com/golang/protobuf v1.3.4/go.mod h1:vzj43D7+SQXF/4pzW/hwtAqwc6iTitCiVSaWz5lYuqw=
 github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5aqRK0M=
-github.com/google/go-cmp v0.5.8 h1:e6P7q2lk1O+qJJb4BtCQXlK8vWEO8V1ZeuEdJNOqZyg=
 github.com/google/go-cmp v0.5.8/go.mod h1:17dUlkBOakJ0+DkrSSNjCkIjxS6bF9zb3elmeNGIjoY=
+github.com/google/go-cmp v0.5.9 h1:O2Tfq5qg4qc4AmwVlvv0oLiVAGB7enBSJ2x2DqQFi38=
+github.com/gsterjov/go-libsecret v0.0.0-20161001094733-a6f4afe4910c h1:6rhixN/i8ZofjG1Y75iExal34USq5p+wiN1tpie8IrU=
+github.com/gsterjov/go-libsecret v0.0.0-20161001094733-a6f4afe4910c/go.mod h1:NMPJylDgVpX0MLRlPy15sqSwOFv/U1GZ2m21JhFfek0=
```
SimonBarendse, line 62: These are all coming in via `aws-vault` dependency: https://github.com/99designs/aws-vault/blob/master/go.mod

In the current version of aws-vault and the current way we're calling it (just using the config from aws-vault), we're not using these dependencies.

Not ideal to pull in all these extra dependencies (to maintain), but great that we're re-using the work done by the community in the aws-vault project on AWS config. Overall I think the trade-off makes sense to pull in these dependencies 👍

SimonBarendse (approved):
Great work, Andi! 🚀

plugins/aws/sts_provisioner.go
AndyTitu: Yes basically a `newProviderFactory` is a function that returns a provider factory. I thought the `new` in the beginning of the name would signal that it's a function and not a provider factory. A factory of a  provider factory, as you noted, if we choose to extrapolate. This is a better alternative to the `SetState` pattern, because the specific provider factory used by the `STSProvisioner` is a `CacheProviderFactory` and needs some initialization with `CacheState`, `CacheOperations` and `ItemFields`.

Do you think we could come up with better naming than `newProviderFactory`? The only other name I can come up with is `newProviderFactoryFunc` which doesn't seem to add a lot of value 🤔

plugins/aws/sts_provisioner.go
```diff
@@ -77,3 +95,263 @@ func (p STSProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput
...
+		return fmt.Errorf("only 1Password-backed OTP authentication is supported by the MFA worklfow of the AWS shell plugin")
+	}
+	// make sure 1Password OTP is used
+	if hasOTP {
+		awsConfig.MfaToken = totp
+	}
+
+	if hasMFASerial && awsConfig.HasMfaSerial() && awsConfig.MfaSerial != mfaSerial {
+		return fmt.Errorf("your local AWS configuration (config file or environment variable) has a different MFA serial than the one specified in 1Password")
+	} else if !awsConfig.HasMfaSerial() {
```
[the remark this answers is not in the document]
AndyTitu: Yes, that is true, but the reason I expressed this as an `else if` is because if I wanted to achieve the same logic without the else I would have had to write:

```
if !awsConfig.HasMfaSerial() && hasMfaSerial {
}
```
which was a bit longer. 
Given this is more a preference detail, I'll choose to not commit a change because I'll be loosing the acquired approvals.

Issue #163, opened 2023-01-24
New plugin: IBM Cloud CLI


williamhpark:
### Platform or tool

IBM Cloud

### Relevant links

* https://cloud.ibm.com/docs/cli?topic=cli-getting-started
* https://cloud.ibm.com/docs/cli?topic=cli-cri-login

### Goal or desired behavior

Authenticate the IBM Cloud CLI using credentials so IBM Cloud users can quickly authenticate into and access their IBM Cloud cloud computing services.

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

The IBM Cloud CLI seems to use environment variables: https://cloud.ibm.com/docs/cli?topic=cli-cri-login.

PR #181, opened 2023-02-13
Draft: New plugin: IBM Cloud CLI

williamhpark:
## Summary

Build a new shell plugin for the IBM Cloud CLI.

Addresses: #163 

### Setup

1. Create an IBM Cloud account [here](https://cloud.ibm.com/login).
1. Build and test the IBM Cloud plugin locally according to [these instructions](https://developer.1password.com/docs/cli/shell-plugins/contribute/), i.e. `make ibmcloud/build`

### Importer

1. Go to https://cloud.ibm.com/iam/apikeys, create a new API key, and copy/download it.
1. Initialize the IBM Cloud plugin and import credentials by entering the API key manually in the terminal. The credentials should be saved in 1Password successfully.

### Provisioner

1. After initializing the IBM Cloud plugin and having your credential saved in 1Password, run `ibmcloud login`. You should be automatically authenticated with the API key stored from environment variables.
1. Try running an `ibmcloud` command with each of the following arguments (descriptions can be found [here](https://cloud.ibm.com/docs/cli?topic=cli-ibmcloud_cli#ibmcloud_login)):
```
-u
-p
--apikey
--cr-token
--profile
--sso
--no-account
```
You should not be prompted to authenticate through the 1Password CLI.

### DIFF
--- a/plugins/ibmcloud/api_key.go
+++ b/plugins/ibmcloud/api_key.go
@@ -1,0 +1,70 @@
+package ibmcloud
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://ibmcloud.com/docs/api_key"), // TODO: Replace with actual URL
+		ManagementURL: sdk.URL("https://console.ibmcloud.com/user/security/tokens"), // TODO: Replace with actual URL
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "API Key used to authenticate to IBM Cloud.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 44,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer: importer.TryAll(
+			importer.TryEnvVarPair(defaultEnvVarMapping),
+			TryIBMCloudConfigFile(),
+		)}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"IBMCLOUD_API_KEY": fieldname.APIKey, // TODO: Check if this is correct
+}
+
+// TODO: Check if the platform stores the API Key in a local config file, and if so,
+// implement the function below to add support for importing it.
+func TryIBMCloudConfigFile() sdk.Importer {
+	return importer.TryFile("~/path/to/config/file.yml", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		// var config Config
+		// if err := contents.ToYAML(&config); err != nil {
+		// 	out.AddError(err)
+		// 	return
+		// }
+
+		// if config.APIKey == "" {
+		// 	return
+		// }
+
+		// out.AddCandidate(sdk.ImportCandidate{
+		// 	Fields: map[sdk.FieldName]string{
+		// 		fieldname.APIKey: config.APIKey,
+		// 	},
+		// })
+	})
+}
+
+// TODO: Implement the config file schema
+// type Config struct {
+//	APIKey string
+// }
--- a/plugins/ibmcloud/api_key.go
+++ b/plugins/ibmcloud/api_key.go
@@ -14,8 +14,8 @@ import (
 func APIKey() schema.CredentialType {
 	return schema.CredentialType{
 		Name:          credname.APIKey,
-		DocsURL:       sdk.URL("https://ibmcloud.com/docs/api_key"), // TODO: Replace with actual URL
-		ManagementURL: sdk.URL("https://console.ibmcloud.com/user/security/tokens"), // TODO: Replace with actual URL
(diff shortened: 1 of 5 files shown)


williamhpark:
Currently, there are a few things I'm unsure about with regards to my implementation.

1. IBM Cloud does create a configuration file at `~/.bluemix/config.json`, however, from what I know there's no mechanism through which API keys are ever automatically saved to this config file through the IBM CLI. I wrote a file importer so that if a user were to manually enter their API key in `~/.bluemix/config.json` with `APIKey` as the key, it could be imported into 1Password. However, if this is too much of a stretch I could remove the file importer altogether.
2. With the current implementation, in order to authenticate you need to run `ibmcloud login`. If you were to run `ibmcloud account list` before authenticating with `ibmcloud login` first, for example, the command would fail. Is there a way that if the user isn't authenticated and they run a command besides `ibmcloud login`, we would run `ibmcloud login` under the hood first?

SimonBarendse:
> IBM Cloud does create a configuration file at ~/.bluemix/config.json when the CLI is set up, however, from what I know there's no mechanism through which API keys are ever automatically saved to this config file through the IBM CLI. I wrote a file importer so that if a user were to manually enter their API key in ~/.bluemix/config.json with APIKey as the key, it could be imported into 1Password. However, if this is too much of a stretch I could remove the file importer altogether.

Does the IBM CLI read the API key from said config file? I'm asking because that would be a reason for users to manually add it.

SimonBarendse:
> With the current implementation, in order to authenticate you need to run ibmcloud login. If you were to run ibmcloud account list before authenticating with ibmcloud login first, for example, the command would fail. If the user isn't authenticated and they run a command besides ibmcloud login, do we want to run ibmcloud login under the hood first so that the command doesn't fail? If this is the desired behaviour, could I get some guidance on how to implement this?

Your current implementation is a great first version, that's already super valuable. It allows users to remove the API key from plaintext on disk, and instead load it through 1Password when and where needed. Not having to run `ibmcloud login` first would be a great additional improvement, but it shouldn't block this current implementation from moving forward already.

plugins/ibmcloud/api_key.go
```diff
@@ -0,0 +1,67 @@
...
+			TryIBMCloudConfigFile(),
+		)}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"IBMCLOUD_API_KEY": fieldname.APIKey,
+}
+
+func TryIBMCloudConfigFile() sdk.Importer {
+	return importer.TryFile("~/.bluemix/config.json", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
```
SimonBarendse, line 46: Since the name of this file completely deviates from the name of the plugin, it warrants a comment why that's okay and it is actually meant to be used with IBM Cloud. Shall we include the following?

```suggestion
	// Bluemix is a former name of IBM Cloud: https://www.ibm.com/cloud/blog/announcements/bluemix-is-now-ibm-cloud
	return importer.TryFile("~/.bluemix/config.json", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
```

plugins/ibmcloud/api_key.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://www.ibm.com/docs/en/app-connect/container?topic=servers-creating-cloud-api-key"),
```
SimonBarendse, line 17: > Use this information to create an IBM® Cloud API key that your deployed integration can use to authenticate

1Password Shell Plugins are built to be used by users, not by deployed integrations. Should we another type of authentication? Maybe username/password?

SimonBarendse (commented):
Love it! :cloud: I have left one comment to further look into and one small suggestion to help future readers of this code.

plugins/ibmcloud/api_key.go
florisvdg: I think this is the URL we're looking for:
```suggestion
		DocsURL:       sdk.URL("https://cloud.ibm.com/docs/account?topic=account-userapikey"),
```

SimonBarendse: That looks right, thank you, Floris

williamhpark:
> Does the IBM CLI read the API key from said config file? I'm asking because that would be a reason for users to manually add it.

@SimonBarendse No there aren't any instances where the CLI reads the API key directly from the config file from what I know. The CLI stores a bearer and refresh token in the config file to authenticate, which may contain information extracted or derived from the API key, but it never reads an API key directly.

After some more thought, I feel like removing the file importer may be the right choice. Most likely, the user would be storing their API key in some other unknown location on their local disk since there isn't even a recommended location that IBM suggests users keep their API key. So in most cases, scanning the config file for an API key would probably not be useful. Can I get your thoughts on  this @SimonBarendse?

plugins/ibmcloud/api_key.go
williamhpark: Updated!

williamhpark:
Added one more small change to skip auth through 1Password when user runs `ibmcloud logout`, to avoid the edge case where a user runs `ibmcloud logout` in a new terminal session and needs to authenticate through 1Password only to be logged out right away.

SimonBarendse:
If the way to pass the key is through an envvar and not through the config file, the envvar importer will be sufficient.

Most likely users are storing the key in their `~/.zshrc` or equivalent to set the environment.

plugins/ibmcloud/api_key.go
williamhpark: Marking as resolved since I removed the file importer code.

AndyTitu:
So the current approach we are taking in this PR will not work. Last time I checked we were waiting on IBM's team to introduce support for an env var in `ibmcloud`. The reason why we can't use a file provisioner here is because the `ibmcloud` cli keeps writing back to the config file and this will block our FIFO indeterminately. Since none of this issue have an ETA on their solutions, I will go ahead and close this PR for the time being.

PR #183, opened 2023-02-15
Add file provisioner helper to provision output dir

florisvdg:
Some CLIs only allow a config file path to be set using the directory it's in. This PR adds a variation to the existing `provision.SetPathAsEnvVar("FOO_CONFIG_FILE_PATH")` helper which lets you provision the output directory as an environment variable, which can be used like this: `provision.SetOutputDirAsEnvVar("FOO_CONFIG_DIR")`.

### DIFF
--- a/sdk/provision/file_provisioner.go
+++ b/sdk/provision/file_provisioner.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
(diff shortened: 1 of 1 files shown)


SimonBarendse (approved):
Thanks @florisvdg! This will come in handy for https://github.com/1Password/shell-plugins/pull/181

PR #184, opened 2023-02-16
Adds support for the Confluent CLI against Confluent Cloud

nerdynick:
This adds support for Confluent's Official CLI. However due to current limits of the Plugin framework, only support for Confluent Cloud has been enabled. However the code is made ready for the eventual support for Confluent Platform installations.


plugins/confluent/credentials.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+const (
+	MDS_URL                      = sdk.FieldName("MDS URL")
+	ConfluentCloudCredentials    = sdk.CredentialName("Confluent Cloud Credentials")
+	ConfluentPlatformCredentials = sdk.CredentialName("Confluent Platform Credentials")
```
hculea, line 14: I propose we move these to `credentials/names.go`, and exclude the name of the platform. I believe that keeping both will result in the double appearance of the platform name within any imported credential.

plugins/confluent/credentials.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+const (
+	MDS_URL                      = sdk.FieldName("MDS URL")
```
hculea, line 12: I'd propose we also move this one to `fieldname/names.go` for consistency with other plugins. We could also consider renaming this to `URL`, since it contains the secret that would be passed via `--url` to `confluent login`, but I'll defer this decision to you.

plugins/confluent/credentials.go
```diff
@@ -0,0 +1,94 @@
...
+				Optional:            true,
+			},
+			{
+				Name:                fieldname.Certificate,
+				MarkdownDescription: "TLS Certificate File used to authenticate to Confluent Platform MDS.",
+				Secret:              false,
+				Optional:            true,
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultPlatformEnvVarMapping),
```
hculea, line 78: I would propose, for better code readability, that we comment out all the bits that relate to Confluent Platform, with a TODO about revisiting this once the Shell Plugins ecosystem adds support for multiple credential types per plugin.

plugins/confluent/plugin.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "confluent",
+		Platform: schema.PlatformInfo{
+			Name:     "Confluent",
+			Homepage: sdk.URL("https://confluent.io"), // TODO: Check if this is correct
```
hculea, line 13: Let's remove this TODO

hculea (commented):
Thank you for this contribution! It's awesome that you thought about adding all the credential types from the get-go. We hope to support multiple credential types in the future, you've really made it easy to ensure that having support for both Confluent Cloud and Platform will be a breeze, at that point 😄 

Only a few comments and good to go!

SimonBarendse (commented):
Thank you for the contribution! Confluent will make for a great addition to the ecosystem! ❤️

plugins/confluent/confluent.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func ConfluentCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Confluent CLI",
+		Runs:    []string{"confluent"},
+		DocsURL: sdk.URL("https://docs.confluent.io/confluent-cli/current/overview.html"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.ForCommand("login"),
```
SimonBarendse, line 15: Does the `login` command write credentials to disk to pass them to the other commands? One of the goals of Shell Plugins is to avoid credentials ever having to touch the disk in plaintext. Instead they're stored encrypted in 1Password and are only passed to the tools that need the credential in memory and only to authorized terminal sessions.

In case Confluent uses a file to pass the credentials between the login and other commands, we can use the [file-provisioner](https://github.com/1Password/shell-plugins/blob/main/sdk/provision/file_provisioner.go) to pass the crednetials directly to the commands that require them (without needing to run `login` beforehand). It uses a temporary in-memory file to pass the credentials to the CLI. As a reference example, a.o. the [MySQL plugin use a temporary file to pass the credential](https://github.com/1Password/shell-plugins/blob/e6148953ffd3a24a79e3fa4eb513fa8bd44d96a5/plugins/mysql/database_credentials.go#L46).

Let us know if we can help out. In addition to here on GitHub we're also on [Slack](https://join.slack.com/t/1password-devs/shared_invite/zt-1halo11ps-6o9pEv96xZ3LtX_VE0fJQA) for quick back-and-forth and you can book a [pairing session](https://calendly.com/d/grs-x2h-pmb/1password-shell-plugins-pairing-session) with us.

nerdynick: The login command performs the exchange of credentials for a JWT Token that has a TTL (Default 1hr if I recall). So without implementing the login API calls within the OP Plugin itself and writing out the token to a temp file. This was the only way I could see fit to connect the 2. As credentials aren't able to be provide in any other way at this time.

It also worth adding that the APIs themselves are rate limited. So even attempting to login, get the token, write it to disk, then performance the user intended action. Would be problematic for the end power user.

plugins/confluent/credentials.go
nerdynick: Are you talking moving them into the root version of that file or creating my own version within the plugin? I didn't add it to the root because none of the others appeared to be doing so. So figured it wasn't allowed.

plugins/confluent/credentials.go
nerdynick: I debated on just using URL, but since I couldn't complete the work for the Platform side of things I just left it alone. Can just move this to URL and rely on the Description to declare what URL we are talking about.

plugins/confluent/credentials.go
nerdynick: I just removed these and replaced them with the generic Credential for now.

plugins/confluent/credentials.go
nerdynick: This is all commented out now and a TODO left about it.

plugins/confluent/confluent.go
SimonBarendse: > It also worth adding that the APIs themselves are rate limited. So even attempting to login, get the token, write it to disk, then performance the user intended action. Would be problematic for the end power user.

For this the plugin ecosystem offers an [encrypted cache](https://github.com/1Password/shell-plugins/blob/e6148953ffd3a24a79e3fa4eb513fa8bd44d96a5/sdk/provisioner.go#L169) to persist the token between command executions. An example of this in use are [temporary AWS credentials](https://github.com/1Password/shell-plugins/blob/e6148953ffd3a24a79e3fa4eb513fa8bd44d96a5/plugins/aws/sts_provisioner.go#L67).

> So without implementing the login API calls within the OP Plugin itself and writing out the token to a temp file.

Is there a Confluent SDK we could use? Or could we possibly capture the output of the `login` command before it hits the disk?

nerdynick: As it stands, there currently is no way to provide the token to the CLI. So while there is a GoLang SDK, there wouldn't be a way at this time to then translate that token to the CLI. I will open a separate ticket/issue with the Confluent CLI team and see about addressing this at a later date. But for now this seems to be the only route forward.

hculea: A possible reason would be that this code currently lives in the `internal` dir (see https://github.com/golang-standards/project-layout#internal). 🤔 

@nerdynick what would be the way to go about asking for some changes in the Confluent CLI's codebase, if we were to go down this route?

nerdynick: I'd be slightly hesitant about attempting to manipulate the internal state cache that the CLI manages. It's constantly changing without backward compatibility, since it's deemed ephemeral and internal to the CLI itself and would require keeping up with new releases, which happen regularly. Outside of going the route that I did with this PR so far. Would be to file an issue against the project or do the work ourselves and hope a PR gets accepted to extend support for a Token to be provided via ENV Vars like the User/Pass can be. For which I started exploring down that route when this thread was brought up. It'd take a bit of time to complete and then also hope to get the PR accepted. 

The follow on question would then be, if we go down the route of asking the Confluent CLI team to make the needed change, would this project, 1Pass CLI Plugins, be OK with importing an external project (SDK) to acquire that token or would we need to duplicate that logic in a minimal fashion within this plugin to achieve that?

AndyTitu: >The follow on question would then be, if we go down the route of asking the Confluent CLI team to make the needed change, would this project, 1Pass CLI Plugins, be OK with importing an external project (SDK) to acquire that token or would we need to duplicate that logic in a minimal fashion within this plugin to achieve that?

The best way forward here seems to start work on the Confluent CLI to be able to take in the token via the env var, as you said. After this is done, we are more than open to import any external SDK that would retrieve that token.

nerdynick: Alright, I was able to modify the plugin and make use of what is essentially an undocumented feature. In that the Confluent CLI is able to take the credentials on every command. So the plugin now provides credentials on all calls but the Help and Version ones. This should mitigate the security concern that have been brought up I believe. This should also be a future proof route as the API documents the use of Credentials on all calls rather then the use of a Token.

plugins/confluent/credentials.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func CloudCredentials() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.Credentials,
```
florisvdg, line 14: Looking at the docs, I see that the Confluent CLI also supports API keys. To avoid confusion with that and to keep forwards compatibility for if we'd ever want to support these API keys in the future, it would be best to pick a less generic name for this username/password credential type.

For example:

```suggestion
		Name:          credname.UserLogin,
```

(Note: this one would need to be added to `sdk/schema/credname/names.go` first)

plugins/confluent/credentials.go
```diff
@@ -0,0 +1,91 @@
...
+		DocsURL:       sdk.URL("https://docs.confluent.io/confluent-cli/current/command-reference/confluent_login.html"),
```
florisvdg, line 15: Maybe [this URL](https://docs.confluent.io/cloud/current/access-management/identity/user-accounts.html#local-user-username-password) would be a better docs URL for this credential type?
```suggestion
		DocsURL:       sdk.URL("https://docs.confluent.io/cloud/current/access-management/identity/user-accounts.html#local-user-username-password"),
```

AndyTitu (approved):
lgtm, excited to get this out! Thank you for the contribution and making it happen with the confluent env vars! @nerdynick

plugins/confluent/credentials.go
```diff
@@ -0,0 +1,91 @@
...
+		ManagementURL: sdk.URL("https://confluent.cloud"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Username,
```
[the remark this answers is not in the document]
florisvdg: For a regular user login it may actually be nicer to use `username` so that it can be used with an existing browser login item.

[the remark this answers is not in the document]
nerdynick: Being able to leverage the same OP Item as the browser was my goal and why I went with it. However in testing, there appears to be a bug with how that import logic works. I assume it's only looking for the field name to be in the name attribute of the field within the saved item. However when you create/save credentials from the browser via the form workflow. The field name value isn't populated but instead only the `designation` field value is populated. This appears to only be for the username field not the password field interestingly.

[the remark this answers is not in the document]
nerdynick: For the curious, 1st is saved via browser 2nd is saved from the OP Plugin workflow

`
{
  "details": {
    "fields": [
      {
        "value": "<redacted>",
        "id": "",
        "name": "",
        "type": "T",
        "designation": "username"
      }
  }
}
`

`
{
  "details": {
    "notesPlain": "",
    "passwordHistory": [],
    "sections": [{
        "fields": [
          {
            "k": "string",
            "n": "<redacted>",
            "t": "username",
            "v": "<redacted>"
          }
     ]}
  ]}
}
`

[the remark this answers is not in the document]
AndyTitu: > I assume it's only looking for the field name to be in the name attribute of the field within the saved item. However when you create/save credentials from the browser via the form workflow. The field name value isn't populated but instead only the designation field value is populated. This appears to only be for the username field not the password field interestingly.

Yes this is an inconsistency that has been singled before in [this PR](https://github.com/1Password/shell-plugins/pull/172) as well, and we should fix. 

I agree then that we should keep this field called `username`.

jpcoenen (approved):
Thanks for your contribution! 🙌  This looks good to me.

accraw:
Hi @nerdynick !

We're about ready to merge this in, but you have some unsigned commits preventing that from happening. Are you able to take a minute to sign them? `git commit -S --amend --no-edit && git push --force-with-lease` or something similar should do the trick.

Thank you!
Amanda

accraw (approved):

accraw:
Actually that doesn't look like it did the trick - I think you might need a rebase: https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits

nerdynick:
Alright, it took some creativity but I believe it's all Rebased, squashed, and signed now.

accraw (approved):

plugins/confluent/confluent.go
accraw: Hi @nerdynick !
I was just trying to use this shell plugin but can't get the username + password to work with commands without first running `confluent login` and writing an auth_token to disk. Is there a trick to get this to work without that step? 

```➜  ~ confluent kafka topic list --url https://<my-domain>.us-east-2.aws.confluent.cloud:443
No session token found, please enter user credentials. To avoid being prompted, run `confluent login`.
Username:```

PR #186, opened 2023-02-16
fix(ngrok): updating formatting for Authtoken description

russorat:
noticed the word authtoken was not formatted correctly when installing the plugin.

### DIFF
--- a/plugins/ngrok/credentials.go
+++ b/plugins/ngrok/credentials.go
@@ -19,7 +19,7 @@ func Credentials() schema.CredentialType {
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.AuthToken,
-				MarkdownDescription: "Auth Token used to authenticate to ngrok.",
+				MarkdownDescription: "Authtoken used to authenticate to ngrok.",
 				Optional:            false,
 				Secret:              true,
 				Composition: &schema.ValueComposition{


claraconney (approved):

arunsathiya (approved):

arunsathiya:
For 1Password devs: When this PR is merged, the graphic and the text on this doc will need to be updated too:

https://developer.1password.com/docs/cli/shell-plugins/ngrok/

plugins/ngrok/credentials.go
```diff
@@ -19,7 +19,7 @@ func Credentials() schema.CredentialType {
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.AuthToken,
```
SimonBarendse, line 21: This is currently `Auth Token` as well and needs to be updated to reflect `authtoken` name used throughout ngrok.

The Name field is also used for the name of the 1Password field in which the authtoken is stored, and where the plugin looks for the authtoken when provisioning. Because we've already released the plugin in stable and folks will have already created items with `Auth Token` as the field name, we're going to add ability to the ecosystem to define alternative fields, so that the plugin can still look for `Auth Token` for backwards compatibility, while using `Authtoken` in all text, and as the default field name for any newly initialized plugins.

As this `Name` field is also used to recognize the field stored in 1Password, and this plugin has already been released in a stable version, we're going to add ability for alternative names, so that `Auth Token` is still accepted for backwards compatibility in addition to the new and default `Authtoken` field.

SimonBarendse: This requires some changes in the 1Password CLI, and @AndyTitu will be taking it on.

arunsathiya: I love seeing the care that goes into small things like this one, thanks team. 🙇🏽 

When I built this plugin, I noticed how ngrok calls their token `Authtoken` instead of the traditional name `Auth Token`, but decided to proceed with `Auth Token` except for using `Authtoken` in the documentation/description fields. That's because I didn't want to have a redundant field name that means the same thing except for how it's formatted.

But now, I have noted that we'd want to be very specific in using the same naming scheme that CLI programs have. 👍🏽

AndyTitu:
This is now treated in https://github.com/1Password/shell-plugins/pull/188 !

AndyTitu:
Thanks @russorat for pointing this out! Since some `op` platform changes were required for making this change work nicely with pre-existing workflows, we took the changes on ourselves. Now this is fixed in #188 .

Issue #185, opened 2023-02-16
1password wrapper breaks some of the ngrok cli commands that do not require authtoken


russorat:
### Platform or tool

ngrok

### Desired or expected behavior

i would like to be able to run `ngrok config edit` without my terminal crashing.

### Current behavior

after installing the ngrok plugin, when i run `ngrok config edit`, i am prompted for my 1password login, and then to select the credentials to use for this command. But this command doesn't need any credentials to run, so this is unnecessary. 

after selecting the correct ngrok credentials, the shell just hangs with an empty screen. hitting ctl+c has no effect.

![image](https://user-images.githubusercontent.com/4805997/219479052-57b453c3-e3a5-42aa-8661-34f9737d60bb.png)

### Relevant log output

```shell
i am using iterm 3.4.19, running ngrok agent 3.1.1, on MacOS 13.2.1 (22D68)
```

### op CLI version

2.14.0

arunsathiya:
Thanks for flagging this, @russorat! This is my fault and a PR to fix this is ready:

https://github.com/1Password/shell-plugins/pull/187

arunsathiya:
I believe this should be resolved as of @AndyTitu's work in https://github.com/1Password/shell-plugins/pull/194, and we are also nearly about to switch to envvars for provisioning credentials: https://github.com/1Password/shell-plugins/pull/222

scottisloud:
Hi all,
Given that this should be resolved and no further comments have been made in the past three years, this issue will be closed. Thanks for raising it!

PR #187, opened 2023-02-16
fix(ngrok): Do not require 1Password authentication for certain (sub)commands

arunsathiya:
Fixes https://github.com/1Password/shell-plugins/issues/185

We currently require 1Password authentication for `ngrok config`'s and `ngrok update`'s subcommands, while it isn't necessary at all. Requiring authentication results in breakage for certain commands, as described here:

- https://github.com/1Password/shell-plugins/issues/185

This PR fixes that issue by not requiring 1Password authentication for `ngrok config`, `ngrok update` and their subcommands.

### DIFF
--- a/plugins/ngrok/ngrok.go
+++ b/plugins/ngrok/ngrok.go
@@ -20,6 +20,7 @@ func ngrokCLI() schema.Executable {
 			needsauth.NotWhenContainsArgs("update"),     // skip 1Password authentication for "ngrok update" and "ngrok update --channel=beta"
 			needsauth.NotWhenContainsArgs("completion"), // skip 1Password authentication for "ngrok completion"
 			needsauth.NotWhenContainsArgs("credits"),    // skip 1Password authentication for "ngrok credits"
+			needsauth.NotWhenContainsArgs("service"),    // skip 1Password authentication for "ngrok service" subcommands because that setup involves defining/knowing in advance the config file path, which isn't something we can handle with 1Password Shell Plugins.
 		),
 		Uses: []schema.CredentialUsage{
 			{
--- a/plugins/ngrok/ngrok.go
+++ b/plugins/ngrok/ngrok.go
@@ -15,9 +15,9 @@ func ngrokCLI() schema.Executable {
 		NeedsAuth: needsauth.IfAll(
 			needsauth.NotForHelpOrVersion(),
 			needsauth.NotWithoutArgs(),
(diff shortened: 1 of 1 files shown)


russorat:
other subcommands that do not require authtoken:

`ngrok completion`
`ngrok credits`
`ngrok help`
`ngrok version`

it's unclear how this would interact with `ngrok service` which will install ngrok as a background service

arunsathiya:
>other subcommands that do not require authtoken:

Thanks Russ, I have updated this PR to reflect that behavior: 3c6fdb29139d6e3571c127f1e4928ea190be1ea8

>it's unclear how this would interact with ngrok service which will install ngrok as a background service

I have just had a look at the [ngrok service docs](https://ngrok.com/docs/ngrok-agent/ngrok#ngrok-service) (and tested), and it seems that the initial setup (`ngrok service start --config path`) will require knowing the config file's path in advance. That's not possible when 1Password wrapper/shell plugin is used. Even when this shell plugin is reworked to support environment variables-based provisioning, we cannot securely generate a config file and use its path. 1Password Shell Plugin's [TempFile provisioner](https://pkg.go.dev/github.com/1Password/shell-plugins/sdk/provision#TempFile) is not relevant is my understanding, because that's 1Password generating a secure config file at runtime and injecting that path into the command.

To sum things up, I am opting to not use 1Password shell plugin for `ngrok service` commands: de09071abe177cf0d4d3cc79f642be86692e0976. The user will need to use a locally store config file for that command.

Open to others feedback/thoughts!

SimonBarendse:
For `ngrok config` to be usable for the non-secret configuration, we also need to merge the on-disk configuration file with the in-memory configuration file we're provisioning, so that any changes you make to the on-disk file persist when the plugin is used for auth.

We could use ngrok's built-in merge functionality for this: https://ngrok.com/docs/ngrok-agent/config#how-config-files-get-merged. @AndyTitu could you give this a look since you recently worked on a similar problem for another plugin?

SimonBarendse:
@russorat for `ngrok service`, does `ngrok service start` accept the `NGROK_AUTHTOKEN` environment variable? Or does that need to be in the configuration file that was used with `ngrok service install`?

AndyTitu:
Thanks @arunsathiya for starting this, I rebased off your branch and continued work for supporting `--config` in https://github.com/1Password/shell-plugins/pull/194

arunsathiya:
I'll close this PR in favor of the ongoing work at #194.

PR #188, opened 2023-02-20
Add alternative field names

AndyTitu:
Related PR: https://github.com/1Password/shell-plugins/pull/186

Add ability to specify various names for a `CredentialField`, so that users aren't constrained to map a field to only one name.

The order in which these names are specified matters towards which 1Password item field will be selected. 
Example:

A `CredentialField` with `Name` set to `d` and `AternativeNames` set to `[]sdk.FieldName{"b", "a", "c"}` and an item with fields named `a`, `b` and `c` will result into a `ProvisionInput` with `ItemFields` containing the value of `b`. If the item were to be added with a field `d` then the value in `ItemFields` would become that of field `d` since the `Name` takes priority over `AlternativeNames`.

For this to work in all cases, a constraint for all fields should be that they do not share any alternative names. Otherwise a conflict will arise in `ItemFields`.

### DIFF
--- a/plugins/ngrok/credentials.go
+++ b/plugins/ngrok/credentials.go
@@ -2,8 +2,6 @@ package ngrok
 
 import (
 	"context"
-	"fmt"
-
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
 	"github.com/1Password/shell-plugins/sdk/provision"
@@ -21,7 +19,7 @@ func Credentials() schema.CredentialType {
 			{
 				Name:                fieldname.Authtoken,
 				AlternativeNames:    []string{"Auth Token"},
-				MarkdownDescription: fmt.Sprintf("%s used to authenticate to ngrok.", fieldname.Authtoken),
+				MarkdownDescription: "Authtoken used to authenticate to ngrok.",
 				Optional:            false,
 				Secret:              true,
 				Composition: &schema.ValueComposition{
--- a/plugins/ngrok/credentials.go
+++ b/plugins/ngrok/credentials.go
@@ -13,17 +13,15 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const ngrokAuthToken = "Authtoken"
-
 func Credentials() schema.CredentialType {
 	return schema.CredentialType{
 		Name:    credname.Credentials,
 		DocsURL: sdk.URL("https://ngrok.com/docs/ngrok-agent/config"),
 		Fields: []schema.CredentialField{
 			{
-				Name:                ngrokAuthToken,
-				AlternativeNames:    []string{fieldname.AuthToken.String()},
-				MarkdownDescription: fmt.Sprintf("%s used to authenticate to ngrok.", ngrokAuthToken),
+				Name:                fieldname.Authtoken,
+				AlternativeNames:    []string{"Auth Token"},
+				MarkdownDescription: fmt.Sprintf("%s used to authenticate to ngrok.", fieldname.Authtoken),
 				Optional:            false,
 				Secret:              true,
 				Composition: &schema.ValueComposition{
@@ -64,7 +62,7 @@ func Credentials() schema.CredentialType {
 
 func ngrokConfig(in sdk.ProvisionInput) ([]byte, error) {
 	config := Config{
-		AuthToken: in.ItemFields[ngrokAuthToken],
+		AuthToken: in.ItemFields[fieldname.Authtoken],
 		APIKey:    in.ItemFields[fieldname.APIKey],
 		Version:   "2", // required field for ngrok CLI to work when file-based configuration is used; automatically configured by the CLI program and is not configurable by the user
 	}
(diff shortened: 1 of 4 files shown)


sdk/schema/credential_type.go
```diff
@@ -33,6 +33,9 @@ type CredentialField struct {
 	// The name of the field, e.g. "Token", "Password", or "Username".
 	Name sdk.FieldName
 
+	// Other names this field can be recognized after. The order in which these names are specified matters towards which 1Password item field will be selected.
```
florisvdg, line 36: How about:
```suggestion
	// Alternative names for this field. Can be used to deprecate field names without breaking existing setups.
	// If there are values present for multiple entries, the first match will be chosen.
```

sdk/schema/credential_type.go
```diff
@@ -33,6 +33,9 @@ type CredentialField struct {
...
+	AlternativeNames []sdk.FieldName
```
florisvdg, line 37: Maybe this should be `[]string` to give us the freedom to potentially clean out the `fieldname` package without being restricted by deprecated values.
```suggestion
	AlternativeNames []string
```

sdk/schema/credential_type.go
```diff
@@ -169,6 +184,12 @@ func (c CredentialType) Validate() (bool, ValidationReport) {
 		Severity:    ValidationSeverityError,
 	})
 
+	report.AddCheck(ValidationCheck{
+		Description: "Has no separate fields that could be identified by the same name",
```
florisvdg, line 188: How about:
```suggestion
		Description: "Has no duplicate field names",
```

sdk/schema/credential_type.go
AndyTitu: Makes sense! Thanks

sdk/schema/credential_type.go
AndyTitu: Thanks, done!

sdk/schema/credential_type.go
AndyTitu: Yes, this more concise, thanks!

plugins/ngrok/credentials.go
```diff
@@ -12,14 +13,17 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+const ngrokAuthToken = "Authtoken"
```
florisvdg, line 16: Let's move this into the `fieldname` package, even if it's just being used by one plugin.

plugins/ngrok/credentials.go
```diff
@@ -12,14 +13,17 @@ import (
...
 func Credentials() schema.CredentialType {
 	return schema.CredentialType{
 		Name:    credname.Credentials,
 		DocsURL: sdk.URL("https://ngrok.com/docs/ngrok-agent/config"),
 		Fields: []schema.CredentialField{
 			{
-				Name:                fieldname.AuthToken,
-				MarkdownDescription: "Auth Token used to authenticate to ngrok.",
+				Name:                ngrokAuthToken,
+				AlternativeNames:    []string{fieldname.AuthToken.String()},
```
florisvdg, line 25: I think it would make more sense to be more explicit here:
```suggestion
				AlternativeNames:    []string{"Auth Token"},
```

plugins/ngrok/credentials.go
AndyTitu: Done, thanks!

plugins/ngrok/credentials.go
AndyTitu: Done!

sdk/schema/fieldname/names.go
```diff
@@ -17,6 +17,7 @@ const (
 	AppSecret       = sdk.FieldName("App Secret")
 	AppToken        = sdk.FieldName("App Token")
 	AuthToken       = sdk.FieldName("Auth Token")
+	Authtoken       = sdk.FieldName("Authtoken")
```
florisvdg, line 20: We should also add it to the `ListAll` function below.

plugins/ngrok/credentials.go
```diff
@@ -18,8 +19,9 @@ func Credentials() schema.CredentialType {
 		DocsURL: sdk.URL("https://ngrok.com/docs/ngrok-agent/config"),
 		Fields: []schema.CredentialField{
 			{
-				Name:                fieldname.AuthToken,
-				MarkdownDescription: "Auth Token used to authenticate to ngrok.",
+				Name:                fieldname.Authtoken,
+				AlternativeNames:    []string{"Auth Token"},
+				MarkdownDescription: fmt.Sprintf("%s used to authenticate to ngrok.", fieldname.Authtoken),
```
florisvdg, line 24: We're not dynamically inserting field names in descriptions in any other plugin, so I don't really see a reason to start doing it now in this plugin.

sdk/schema/fieldname/names.go
AndyTitu: Done, thanks for catching that

plugins/ngrok/credentials.go
AndyTitu: Fair enough, modified back

florisvdg (approved):

hculea (approved):

PR #191, opened 2023-02-28
Draft: Add capability to set a default credential when there are multiple

accraw:
also map to a command

Using ngrok as an axample, add functionality to have multiple credential types, such as auth token and api token. This means also adding the ability to set a credential type as the default, and mapping commands to a credential type. There will be op changes to match to these.

### DIFF
--- a/plugins/ngrok/api_credentials.go
+++ b/plugins/ngrok/api_credentials.go
@@ -1,0 +1,86 @@
+package ngrok
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+	"gopkg.in/yaml.v3"
+)
+
+func APICredentials() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.APIKey,
+		DocsURL: sdk.URL("https://ngrok.com/docs/ngrok-agent/config"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "API Key used to authenticate to ngrok API.",
+				Optional:            true,
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 48,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.TempFile(ngrokApiConfig, provision.Filename("config.yml"), provision.AddArgs("--config", "{{ .Path }}")),
+		Importer: importer.TryAll(
+			importer.TryEnvVarPair(defaultApiEnvVarMapping),
+			importer.MacOnly(
+				TryngrokAPIConfigFile("~/Library/Application Support/ngrok/ngrok.yml"),
+			),
+			importer.LinuxOnly(
+				TryngrokAPIConfigFile("~/.config/ngrok/ngrok.yml"),
+			),
+		)}
+}
+
+func ngrokApiConfig(in sdk.ProvisionInput) ([]byte, error) {
+	config := APIConfig{
+		APIKey:  in.ItemFields[fieldname.APIKey],
+		Version: "2", // required field for ngrok CLI to work when file-based configuration is used; automatically configured by the CLI program and is not configurable by the user
+	}
+	contents, err := yaml.Marshal(&config)
+	if err != nil {
+		return nil, err
+	}
+	return []byte(contents), nil
+}
+
+var defaultApiEnvVarMapping = map[string]sdk.FieldName{
+	"NGROK_API_KEY": fieldname.APIKey,
+}
+
+func TryngrokAPIConfigFile(path string) sdk.Importer {
+	return importer.TryFile(path, func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		var config APIConfig
+		if err := contents.ToYAML(&config); err != nil {
+			out.AddError(err)
+			return
+		}
+
+		if config.APIKey == "" || config.Version == "" {
+			return
+		}
+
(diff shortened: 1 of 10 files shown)


plugins/ngrok/auth_credentials.go
```diff
@@ -61,8 +48,7 @@ func Credentials() schema.CredentialType {
 
 func ngrokConfig(in sdk.ProvisionInput) ([]byte, error) {
 	config := Config{
-		AuthToken: in.ItemFields[fieldname.Authtoken],
-		APIKey:    in.ItemFields[fieldname.APIKey],
+		AuthToken: in.ItemFields[fieldname.AuthToken],
```
AndyTitu, line 51: This should be `fieldname.Authtoken` because ngrok calls this secret `authtoken` as opposed to `Auth Token`

plugins/ngrok/auth_credentials.go
```diff
@@ -73,8 +59,7 @@ func ngrokConfig(in sdk.ProvisionInput) ([]byte, error) {
 }
 
 var defaultEnvVarMapping = map[string]sdk.FieldName{
-	"NGROK_AUTHTOKEN": fieldname.Authtoken,
-	"NGROK_API_KEY":   fieldname.APIKey,
+	"NGROK_AUTHTOKEN": fieldname.AuthToken,
```
AndyTitu, line 62:
```suggestion
	"NGROK_AUTHTOKEN": fieldname.Authtoken,
```

plugins/ngrok/api_credentials.go
```diff
@@ -0,0 +1,86 @@
...
+					Length: 48,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.TempFile(ngrokApiConfig, provision.Filename("config.yml"), provision.AddArgs("--config", "{{ .Path }}")),
```
AndyTitu, line 35: I recently noticed that ngrok now supports env var for the api credential : https://ngrok.com/docs/ngrok-agent/#supported-env-variables. It could be more straightforward to use that here and get rid of hardcoding `Version: 2` in the file

plugins/ngrok/auth_credentials.go
```diff
@@ -32,21 +32,8 @@ func Credentials() schema.CredentialType {
...
-					Charset: schema.Charset{
-						Uppercase: true,
-						Lowercase: true,
-						Digits:    true,
-					},
-				},
-			},
 		},
+		DefaultCredential:  true,
 		DefaultProvisioner: provision.TempFile(ngrokConfig, provision.Filename("config.yml"), provision.AddArgs("--config", "{{ .Path }}")),
```
AndyTitu, line 37: same as above, we could use the AUTH_TOKEN env var: https://ngrok.com/docs/ngrok-agent/#supported-env-variables

sdk/schema/executable.go
```diff
@@ -37,6 +40,9 @@ type CredentialUsage struct {
 	// set in the credential schema, so should only be used if this executable requires a custom configuration, that deviates
 	// from the way the credential is usually provisioned.
 	Provisioner sdk.Provisioner
+
+	// (Optional) Whether the credential is specific to a specific argument
```
AndyTitu, line 44:
```suggestion
	// (Optional) Whether the credential is needed for authentication by a specific argument
```

sdk/schema/executable.go
```diff
@@ -23,6 +23,9 @@ type Executable struct {
 
 	// (Optional) Whether the exectuable needs authentication for certain args.
 	NeedsAuth sdk.NeedsAuthentication
+
+	// (Optional) words to prompt about a credential
+	AuthPrompt string
```
AndyTitu, line 28: I am failing to see what this does. Could you please elaborate on why this is needed in this scenario?

sdk/schema/plugin.go
```diff
@@ -21,6 +21,16 @@ type Plugin struct {
 	Executables []Executable
 }
 
+// Return the default credential for a plugin, or the first one in the list if it isn't set
+func (p Plugin) DefaultCredentialType() CredentialType {
```
AndyTitu, line 25: what if all credentials are set to be the default one? Should we even allow this? I would argue we would need to enforce that only one credential can be the default one, in `plugin.go` via the validation checks. 

The alternative would be to let the user set as many credentials as they want as default only to select the first one in order as the actual default one which could be confusing. The only scenario I could think of right now for which a plugin could have more than 1 default credential type defined, is because it also has more than 1 executable, and one of those executable has a certain default credential whereas another has a different default credential. However, in this case we could drop the `DefaultCredential` field altogether and solely rely on the order of specification for selecting the default credential for a certain executable. This would need to be documented. What do you think?

AndyTitu (commented):
Looking good, thanks for enabling executables to work with multiple credential types! Left a few comments:

sdk/schema/plugin.go
florisvdg: I don't think there really is such thing as a "default credential". There are just certain credentials needed for certain commands.

If this data is used in `op` to determine the order to configure the credentials, then we can also get that from the order the credentials have in the `Executable` definition, and avoid the whole concept of "default credentials".

AndyTitu: I am fine with this!

sdk/schema/executable.go
accraw: Does the new comment clarify enough?

plugins/ngrok/api_credentials.go
AndyTitu: Actually, I opened an issue for this, and we can treat it with another occasion: https://github.com/1Password/shell-plugins/issues/216

plugins/ngrok/auth_credentials.go
AndyTitu: See issue: https://github.com/1Password/shell-plugins/issues/216

sdk/schema/executable.go
florisvdg: To not create confusion with the biometric auth prompt functionality, how about:
```suggestion
	Description string
```

sdk/schema/plugin.go
florisvdg: So that would also apply to the `DefaultCredential` field in the credential schema.

AndyTitu:
Closing this as a lot of the same functionality will be added via https://github.com/1Password/shell-plugins/pull/256

PR #194, opened 2023-03-02
Be aware of already existing Ngrok config file 

AndyTitu:
Based on @arunsathiya 's [additions](https://github.com/1Password/shell-plugins/pull/187) this PR looks deeper into handling an already existing config file.

## Thought process

check if `--config` flag is specified with a path to the already existing config file. If yes, read that file, inject it with sensitive information from 1Password and provision it via in-memory file provisioning. If not, use the default locations for the config file. In this way, configs are also present alongside credentials for each run.


AndyTitu:
@arunsathiya @russorat it would be awesome if you could test these changes against your local ngrok setup as well. You might have some configuration that I missed (i.e. is it okay to use `time.Duration` for `connect_timeout` in config, or should I just stick to `int`?).

arunsathiya:
This looks like a great next step, @AndyTitu, thank you. 🙌🏽 

I haven't taken a proper look at the changes so far, but some quick tests seem to run well for me. 

I ran into an error but I am unable to reproduce it at the moment, nor do I remember the exact error message. I'll keep testing for another day or two and followup with my findings.

A minor note in the meantime though: I see several "AuthToken" in the struct definition. Should we update that to "Authtoken", in alignment with the field name?

AndyTitu:
Linter fails unexpectedly: https://github.com/1Password/shell-plugins/issues/195. Solving this in https://github.com/1Password/shell-plugins/pull/196

plugins/ngrok/credentials.go
```diff
@@ -98,8 +84,37 @@ func TryngrokConfigFile(path string) sdk.Importer {
...
+	WebAddress         string                  `yaml:"web_addr"`
+	WebAllowHosts      string                  `yaml:"web_allow_hosts"`
+}
+
+// TunnelConfig struct is non-exhaustive. It only covers few most common fields, for now.
+type TunnelConfig struct {
+	Address  string `yaml:"addr"`
+	Metadata string `yaml:"metadata"`
+	Proto    string `yaml:"proto"`
+	HostName string `yaml:"hostname"`
```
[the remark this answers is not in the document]
AndyTitu: Ah, you are right, I updated it now. Should've thought of this before researching the whole ngrok documentation for all those fields 😅

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+	args = append(args, "--config")
+	args = append(args, newFilePath)
+	out.CommandLine = args
+	return ""
+}
+
+func (f fileProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput, out *sdk.DeprovisionOutput) {
+	// nothing to do here: files get deleted automatically by the CLI
```
SimonBarendse, line 76:
```suggestion
	// nothing to do here: files get deleted automatically by 1Password CLI
```

To distinguish from ngrok's CLI

SimonBarendse (dismissed):

SimonBarendse (approved):

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,81 @@
...
+		if arg == "--config" {
+			if i+1 != len(args) {
+				existingFilePath := args[i+1]
+				args[i+1] = newFilePath
+				return existingFilePath
+			}
+		}
+	}
+	args = append(args, "--config")
+	args = append(args, newFilePath)
```
hculea, line 70: From what I understand, if it happens that `--config` is the last element in the command line and that nothing is passed to it, we simply append `--config $newFilePath` to the command? The command would be `--config --config  $newFilePath` in this case. This seems rather wrong, I'd expect some handling for when `--config` doesn't receive a value.

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	out.AddSecretFile(provisionedConfigFilePath, newContents)
+}
+
+func processConfigFlag(out *sdk.ProvisionOutput, newFilePath string) string {
+	args := out.CommandLine
+	for i, arg := range args {
+		if arg == "--config" {
+			if i+1 != len(args) {
```
hculea, line 62: This condition can be simplified by chaining these two statements with `&&`.

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,81 @@
...
+		return
+	}
+
+	out.AddSecretFile(provisionedConfigFilePath, newContents)
+}
+
+func processConfigFlag(out *sdk.ProvisionOutput, newFilePath string) string {
+	args := out.CommandLine
+	for i, arg := range args {
+		if arg == "--config" {
```
hculea, line 61: What happens when `--config=/path/to/config` is used instead of `--config /path/to/config`?

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,81 @@
...
+	newContents, err := yaml.Marshal(&config)
+	if err != nil {
+		out.AddError(err)
+		return
+	}
+
+	out.AddSecretFile(provisionedConfigFilePath, newContents)
+}
+
+func processConfigFlag(out *sdk.ProvisionOutput, newFilePath string) string {
```
hculea, line 58: This function name doesn't really reflect what it does. Let's consider changing it to reflect that it replaces the path passed by the user with a temp path and returns the initial path, and add a comment to it.

AndyTitu: Done, thanks!

plugins/ngrok/provisioner.go
AndyTitu: Good point, done. Also added some tests

plugins/ngrok/provisioner.go
AndyTitu: Done, thanks! Added some tests to validate we don't run into issues like this

SimonBarendse (approved):

hculea (approved):

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,104 @@
...
+	config := make(map[string]interface{})
+
+	var existingConfigFilePath string
+	// use default locations, depending on the OS
+	switch runtime.GOOS {
+	case "darwin":
+		existingConfigFilePath = filepath.Join(in.HomeDir, "/Library/Application Support/ngrok/ngrok.yml")
+	case "linux":
+		existingConfigFilePath = filepath.Join(in.HomeDir, "/.config/ngrok/ngrok.yml")
+	}
```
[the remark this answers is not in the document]
hculea: Nvm, figured out this is only for importers, in fact.

plugins/ngrok/provisioner.go
```diff
@@ -0,0 +1,104 @@
...
+				return "", append(args, newFilePath)
+			}
+
+		} else if strings.HasPrefix(arg, "--config=") {
+			existingFilePath := strings.TrimPrefix(arg, "--config=")
+			args[i] = fmt.Sprintf("--config=%s", newFilePath)
+			return existingFilePath, args
+		}
+	}
+	return "", append(args, fmt.Sprintf("--config=%s", newFilePath))
```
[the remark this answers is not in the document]
AndyTitu: That's a good point, ad we should be aware of that in the future when handling flags! However, for ngrok this does not seem to be the case. All paths specified through `--config` will be considered and merged.  This merging behaviour clashed with the purpose of our file provisioning so we are not looking to support that for now. A good way to fully support this use case in the future is to switch to env var provisioners.

PR #198, opened 2023-03-13
Add support for MongoDB Atlas CLI, with environment variable based importing and provisioning

joqim:
This PR adds support for the [MongoDB Atlas CLI](https://www.mongodb.com/tools/atlas-cli). This is different from the [core MongoDB CLI](https://www.mongodb.com/docs/mongocli/stable/install/). 

## Core features

- The authentication setup currently is "Public Key" and "Private Key". These are obtained from the Altas organization > Access Manager > API Keys. (One part that's not clear to me yet is, does the key generated here apply to all the projects within the organization? I am failing to a project-specific API generator)
- The shell plugin looks for the Atlas CLI's environment variables and imports into 1Password.
- If there are no environment variables in the current shell session, 1Password prompts the user to manually enter them.
- Provisions the shell plugin for every "atlas" command (except help and version commands) with the environment variables.

## Main issues

- Atlas CLI supports config file too, but we don't look for it and import the credentials just yet because it uses a different type of authentication: `access token`, and `refresh token`.
- When there is no Atlas CLI config file, but if a MongoDB CLI config file exists, the Atlas commands show a prompt asking if the user would like to import the config file from MongoDB CLI path to the Atlas CLI path. This kind of hinders the 1Password shell plugin's authentication flow. 

This is my first time contributing to Shell Plugins, feedback welcome. Thanks, @arunsathiya, for getting me started with 1Password and Shell Plugins.


plugins/atlas/atlas.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func MongoDBAtlasCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "MongoDB Atlas CLI", // TODO: Check if this is correct
```
AndyTitu, line 12: this looks correct, it's safe to remove the TODO

plugins/atlas/atlas.go
```diff
@@ -0,0 +1,25 @@
...
+		Runs:      []string{"atlas"},
+		DocsURL:   sdk.URL("https://www.mongodb.com/docs/atlas/cli/v1.2/"),
```
AndyTitu, line 14: nit: how often do we expect atlas cli to release new versions? Given this link contains a specific version, we could consider using this link: https://www.mongodb.com/docs/atlas/getting-started/ which shows how to get started including directions to the how to install page

plugins/atlas/credentials.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func Credentials() schema.CredentialType {
+	return schema.CredentialType{
+		Name: credname.Credentials,
```
AndyTitu, line 14: Should we create another credname here for understandability? `Credentials` is kind of general so I was thinking `PrivateKeyPair` (a name for the public/private key pair) could have more significance here.

plugins/atlas/plugin.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "atlas",
+		Platform: schema.PlatformInfo{
+			Name:     "MongoDB Atlas",
+			Homepage: sdk.URL("https://www.mongodb.com/"), // TODO: Check if this is correct
```
AndyTitu, line 13: this is right, we could remove the TODO

plugins/atlas/credentials_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func TestCredentialsProvisioner(t *testing.T) {
+	plugintest.TestProvisioner(t, Credentials().DefaultProvisioner, map[string]plugintest.ProvisionCase{
+		"default": {
+			ItemFields: map[sdk.FieldName]string{ // TODO: Check if this is correct
```
AndyTitu, line 14: this is correct, no TODO needed

AndyTitu (commented):
Code looks good to me overall, I just left a few nitpicks. Awesome job @joqim , thanks for contributing!

AndyTitu:
I think supporting the environment variables - public/private key-pair - workflow for now is a good approach. We could introduce the config file auth flow further down the line.

plugins/atlas/credentials.go
arunsathiya: My only concern is that `PrivateKeyPair` wouldn't be a relevant credential name when we introduce config file-based authentication in the future, because it uses a different type of authentication -- based on Joqim's PR description.

arunsathiya (approved):
Great to see your first PR, Joqim! 🎉 

Overall, it looks good to me but I left [one comment here to discuss further about the credential name](https://github.com/1Password/shell-plugins/pull/198#discussion_r1139816690).

joqim:
I've addressed your changes, please take a look into it.

plugins/atlas/credentials.go
AndyTitu: Indeed, but would it not remain relevant for the existing use case of the env vars?

AndyTitu (approved):
lgtm

plugins/atlas/credentials.go
arunsathiya: I agree with shipping this for now. 👍🏼 I am keen to understand whether renaming the credential name in the future would be considered a breaking change though.

arunsathiya (approved):
The current changes look good to me too.

plugins/atlas/credentials.go
AndyTitu: Generally, yes, this would make for a breaking change.

However, it should not happen that we have to change that name here. In this specific case, when we introduce a new way of authentication, we'll likely want to create a second `CredentialType` altogether, one that is not a private key pair, and will be called `access token` from what I read.

plugins/atlas/privatekeypair.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func PrivateKeyPair() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.PrivateKeyPair,
```
florisvdg, line 14: Question: Glancing at the docs, isn't this credential what Mongo calls the "API Key"? Which consists of a private and public key: https://www.mongodb.com/docs/atlas/configure-api-access/#std-label-atlas-admin-api-access

florisvdg: image

joqim: I agree with your suggestion. I've updated the credential names/ filenames accordingly. Please let me know if you have any other feedback.

florisvdg: And would it make sense to use [this page](https://www.mongodb.com/docs/atlas/configure-api-access/) as the API Key's `DocsURL`?

plugins/atlas/apikey.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://www.mongodb.com/docs/atlas/cli/stable/atlas-cli-env-variables/"),
```
florisvdg, line 15: And would it make sense to use [this page](https://www.mongodb.com/docs/atlas/configure-api-access/) as the API Key's `DocsURL`?

```suggestion
		DocsURL:       sdk.URL("https://www.mongodb.com/docs/atlas/configure-api-access/"),
```

SimonBarendse (approved):
Appreciate the contribution! Thank you for taking the time to submit ❤️

AndyTitu (approved):
re-approving

Issue #204, opened 2023-03-13
New plugin: Gitea/Tea


14zombies:
### Platform or tool

Gitea

### Relevant links

* Platform Homepage: https://gitea.io
* CLI docs: https://gitea.com/gitea/tea

### Goal or desired behavior

Tea is the CLI for Gitea the self-hosted git service (similar to the GitHub cli). It stores credentials in a config file on disk at ~/.config/tea/config.yml.

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

Digging into the source https://gitea.com/gitea/tea/src/branch/main/cmd/login/add.go it appears that environmental variables could be used, however with a bit of experimentation that seems to not be the case. 
<img width="863" alt="Screenshot 2023-03-13 at 21 11 49" src="https://user-images.githubusercontent.com/17486387/224833343-93f711dd-5aca-473d-b5ee-084f9eee50c3.png">

PR #205, opened 2023-03-13
Add support for Gitea (tea)

14zombies:
Plugin for Gitea CLI (aka tea)

Addresses #204 

### Prerequisites
- Sign up for [Gitea](https://gitea.com) (This step can be skipped if you're already hosting your own Gitea instance)
- Install [tea](https://gitea.com/gitea/tea)
- Build the plugin locally: `make gitea/build`

### Setup
- Create an application access token either at (Gitea.com)[https://gitea.com/user/settings/applications] or on your own instance (User Settings → Applications)
- Initialise the Tea plugin or use `tea login add` to sign in with a username & password which will then be imported.

## Provisioner
Tea stores a config file in ~/.config/tea/config.yml with the token, user, & url amongst other preferences.

## Issues
Tea reads the credentials from a set location which creates a couple of issues:
- If the config file already exists the plugin will fail. Removing this file once the plugin is configured is a good idea anyway.
- Skimming through the source it reads the file from $XDG_CONFIG_HOME if this is set to a location other than ~/.config the plugin will not work.
- After running `op plugin init tea` the alias is not setup and I'm not sure why.

## Notes
As far as I can tell there is no way to pass credentials through ENV or to pass the location of the config though ENV/flags either. 
Flags can be used to set the token, user, & url along with a blank password (--password " ") however the credentials are then stored in the config.yml.

### DIFF
--- a/plugins/gitea/personal_access_token.go
+++ b/plugins/gitea/personal_access_token.go
@@ -1,0 +1,120 @@
+package gitea
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+	"gopkg.in/yaml.v2"
+)
+
+var configPath string = "~/.config/tea/config.yml"
+
+func PersonalAccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.PersonalAccessToken,
+		DocsURL:       sdk.URL("https://gitea.com/user/settings/applications"),
+		ManagementURL: nil,
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Token,
+				MarkdownDescription: "Token used to authenticate to Gitea.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 40,
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+			{
+				Name:                fieldname.HostAddress,
+				MarkdownDescription: "The Gitea host to connect to, should start with https://",
+				Secret:              false,
+				Composition: &schema.ValueComposition{
+					Charset: schema.Charset{
+						Lowercase: true,
+						Uppercase: true,
+						Digits:    true,
+						Symbols:   true,
+					},
+					Prefix: "https://",
+				},
+			},
+			{
+				Name:                fieldname.User,
+				MarkdownDescription: "Your Gitea username",
+				Secret:              false,
+			},
+		},
+		DefaultProvisioner: provision.TempFile(giteaConfig, provision.AtFixedPath(configPath)),
+		Importer: importer.TryAll(
+			TryGiteaConfigFile(),
+		)}
+}
+
+func giteaConfig(in sdk.ProvisionInput) ([]byte, error) {
+	config := Config{
+		Logins: []Login{
+			{
+				Name:    in.ItemFields[fieldname.HostAddress],
+				URL:     in.ItemFields[fieldname.HostAddress],
+				Token:   in.ItemFields[fieldname.Token],
+				Default: true,
+				User:    in.ItemFields[fieldname.User],
+			},
+		},
+	}
+	contents, err := yaml.Marshal(&config)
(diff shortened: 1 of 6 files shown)


plugins/gitea/personal_access_token.go
```diff
@@ -0,0 +1,120 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+var configPath string = "~/.config/tea/config.yml"
+
+func PersonalAccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.PersonalAccessToken,
+		DocsURL:       sdk.URL("https://gitea.com/user/settings/applications"),
+		ManagementURL: nil,
```
SimonBarendse, line 21: For the DocsURL we expect a documentation page that is not behind a login. The management URL is the in-product page where you can create and manage the credential(s).

See for example the [GitLab plugin](https://github.com/1Password/shell-plugins/blob/main/plugins/gitlab/personal_access_token.go) for an example.

```suggestion
		DocsURL:       sdk.URL("https://docs.gitea.io/en-us/command-line/"),
		ManagementURL: sdk.URL("https://gitea.com/user/settings/applications"),
```

plugins/gitea/plugin.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "gitea",
+		Platform: schema.PlatformInfo{
+			Name:     "Gitea",
+			Homepage: sdk.URL("https://gitea.com/"),
```
SimonBarendse, line 13: We usually link to the project homepage here
```suggestion
			Homepage: sdk.URL("https://gitea.io/"),
```

plugins/gitea/tea.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func giteaCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Gitea CLI",
+		Runs:    []string{"tea"},
+		DocsURL: sdk.URL("https://gitea.com/gitea/tea/src/branch/main/README.md"),
```
SimonBarendse, line 14:
```suggestion
		DocsURL: sdk.URL("https://docs.gitea.io/en-us/command-line/"),
```

plugins/gitea/personal_access_token.go
```diff
@@ -0,0 +1,120 @@
...
+	SSHHost           string `yaml:"ssh_host"`
+	SSHKey            string `yaml:"ssh_key"`
+	Insecure          bool   `yaml:"insecure"`
+	SSHCertPrincipal  string `yaml:"ssh_certificate_principal"`
+	SSHAgent          bool   `yaml:"ssh_agent"`
+	SSHKeyFingerprint string `yaml:"ssh_key_agent_pub"`
+	SSHPassphrase     string `yaml:"-"`
+	VersionCheck      bool   `yaml:"version_check"`
+	User              string `yaml:"user"`
+	Created           int64  `yaml:"created"`
```
SimonBarendse, line 115: we could consider omitting all the fields we're not using from this struct, for simplicity. The yaml parser will skip/ignore these fields in the yaml file.

It is great that the test fixtures do contain these fields, so we ensure a realistic configuration file used in practice will work 👍 So I'd definitely keep the fields in there.

SimonBarendse (commented):
Thank you for your contribution! 🙌

plugins/gitea/personal_access_token.go
14zombies: I had a bit of indecision about this. The Gitea command-line section in their docs is for running/administering the server. The documentation for users in general is fairly sparse, the only place I've been able to find documentation about tea is that Readme in its repo. However there is no mention of Tokens. Perhaps https://docs.gitea.io/en-us/api-usage/#generating-and-listing-api-tokens would be better, there is a small mention at the bottom of that section about generating a Token through the UI.

plugins/gitea/tea.go
14zombies: As above the command line refers to the server it's self rather than the end user cli tea.

plugins/gitea/personal_access_token.go
SimonBarendse: I like your suggestion to link to the docs section on generating and listing API tokens 👍 That's indeed more useful for end-users than the command-line doc I proposed.

hculea (approved):

Issue #203, opened 2023-03-13
New plugin: Zendesk


williamhpark:
### Platform or tool

Zendesk

### Relevant links

* Platform homepage: https://www.zendesk.com/in/
* CLI docs: https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/
* Authentication docs: https://developer.zendesk.com/api-reference/introduction/security-and-auth/

### Goal or desired behavior

Authenticate the Zendesk CLI using Login Details so users can quickly authenticate into Zendesk and develop Zendesk apps.

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

The Zendesk CLI uses environment variables for authentication https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/#environment-variables. There are three authentication schemes that are supported:

<img width="651" alt="image" src="https://user-images.githubusercontent.com/53918631/224829768-cc0bddc2-66aa-4c2d-ab54-2f8486043267.png">

Since `API token` takes precedence over `Basic authentication` according to the docs, and implementation complexity is the same, we should probably use the `API token` auth method.

PR #207, opened 2023-03-14
New plugin: Zendesk

williamhpark:
## Summary

Add support for a new plugin for the Zendesk CLI, ZCLI.

Addresses: #203

### Setup

1. Create a Zendesk account (they offer a [free trial](https://www.zendesk.com/register/#step-1))
2. In your Zendesk account, generate an API token following [these instructions](https://support.zendesk.com/hc/en-us/articles/4408889192858-Generating-a-new-API-token)
3. install the Zendesk CLI following [these instructions](https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/)
4. Build and test the ZCLI plugin locally according to [these instructions](https://developer.1password.com/docs/cli/shell-plugins/contribute/). E.g. `make zendesk/build`

### Importer

After installing and setting up ZCLI, import [subdomain](https://support.zendesk.com/hc/en-us/articles/4409381383578-Where-can-I-find-my-Zendesk-subdomain-), email, and API key by entering the credentials manually in the terminal. The credentials should be saved in 1Password successfully.

### Provisioner

After initializing the ZCLI plugin and having your credentials saved in 1Password, run a `zcli` command that requires authentication, i.e. one of the following:

```
apps:create
apps:package
apps:update
apps:validate
```

You should be taken through the authentication flow with the Zendesk credentials that were saved in 1Password, and you shouldn't have to enter the credentials manually. Once authenticated, the command should run successfully.

## Additional Information

Zendesk allows you to authenticate using any of three authentication schemes:

image

The documentation suggests that users don't use basic authentication, API token auth takes precedence over basic auth, and 2FA isn't supported with basic authentication according to [this documentation](https://developer.zendesk.com/api-reference/introduction/security-and-auth/), so we can assume existing users would likely not use this scheme. Also, OAuth implementation is more complicated. So, the API token scheme seemed like the best choice.
 
Be aware that the plugin name is `zendesk` and the executable name is `zcli`.


plugins/zendesk/api_token.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.APIToken,
```
williamhpark, line 14: The three credentials that are required for authentication are `Subdomain`, `Email`, and `API token`. I'm unsure about the `credname` as well as `fieldname` values I used, please let me know if I should change them to make it clearer which credentials the plugin needs.

arunsathiya (commented):
Overall, this is a great start! 🙌🏽 I'll drop a couple of things to think about.

## Biggest concern: Should `Username` field be appended with `/token`?

image

Based on the [zcli docs](https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/), using API Token for authentication requires that `/token` be appended to the Username. In our current implementation, we are not doing that. So, I suspect authentication should fail. Were you able to get it working without that? 🤔 

Personally, I am not able to get zcli authentication working unfortunately (with and without 1Password Shell Plugins) due to this bug:

- https://github.com/zendesk/zcli/issues/147

So, I can't be of much help with testing. All of my feedback here is only based on the code.

## Validate the OrgUrl field?

What if someone enters `test.zendesk.com` instead of just `test` in the `OrgUrl` field? Would the `zcli` program accept that just fine? Same for `https://test.zendesk.com`?

## Do not require authentication for a few more subcommands

image


Based on the zcli docs, we can skip authentication for `apps:bump`, `apps:clean`, `apps:new` and `apps:server` too.

plugins/zendesk/api_token.go
arunsathiya: Personally, I have defaulted to `Credentials` as the credential name when I am unclear on the best name, but in this case, defining `LoginDetails` as a cred name and using it sounds good to me. I don't see defining new cred names as harmful. Even if, in the future, we need to define alternative credential names, something like defining [alternative field names](https://github.com/1Password/shell-plugins/pull/188) might be possible.

florisvdg: I think `credname.APIToken` is fine as is. That's the actual thing you're creating/deleting and that's how Zendesk refers to it.

williamhpark:
Hey @arunsathiya, thanks so much for the feedback!

> Based on the [zcli docs](https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/), using API Token for authentication requires that /token be appended to the Username. In our current implementation, we are not doing that. So, I suspect authentication should fail. Were you able to get it working without that? 🤔

Yep with the current implementation I was able to get it working. If you take a look at the documentation, you'll see that when using env vars you don't want to append `/token` because Zendesk does it for us:

image

> What if someone enters test.zendesk.com instead of just test in the OrgUrl field? Would the zcli program accept that just fine? Same for https://test.zendesk.com?

All three that you mentioned (`test`, `test.zendesk.com`, and `https://test.zendesk.com`) will be accepted when a user runs `zcli login -i` and `Successfully logged in` will display, but it seems that the `app` commands that require authentication only succeed if the subdomain provided by the user is just `test`. I feel like implementing validation on the 1Password-side is probably not necessary, since it should be up to the user to import the correct credentials that would successfully authenticate the CLI even without using 1Password. I feel that if there is validation, it should be ZCLI that performs the validation itself. Let me know if you disagree.

> Based on the zcli docs, we can skip authentication for apps:bump, apps:clean, apps:new and apps:server too.

Thanks for catching that! I'll make that change right now.

arunsathiya:
>Yep with the current implementation I was able to get it working. If you take a look at the documentation, you'll see that when using env vars you don't want to append /token because Zendesk does it for us:

Oh, that's great! 👍🏽 

>That being said, if OrgURL feels like it would cause users to enter the full URL and not just the subdomain, changing the fieldname to something more specific like Subdomain could be an option.

I think switching to `Subdomain` makes a lot of sense indeed. That's also what the `zcli login -i` flow indicates:

```
➜  ~ zcli login -i
########################################################################
# WARNING: 'zendesk' is not from the official registry.                #
# Only proceed if you are the developer of 'zendesk'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/zendesk. #
########################################################################
Subdomain: example
Email: example@example.com
Password:
```

Also, should we switch from `Username` to `Email` as that matches the usage throughout [this documentation page](https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/) and the `zcli login -i` flow?

plugins/zendesk/api_token.go
hculea: I'd say we should refer to them as Zendesk does, since that's what would make most sense from the user-perspective, when creating a new plugin item.

So if they use `Subdomain` instead of `OrgURL` and `Email` instead of `Username`, let's go with that.

williamhpark: @hculea That makes sense, just made a change to reflect that!

williamhpark:
> I think switching to `Subdomain` makes a lot of sense indeed. That's also what the `zcli login -i` flow indicates:
> 
> ```
> ➜  ~ zcli login -i
> ########################################################################
> # WARNING: 'zendesk' is not from the official registry.                #
> # Only proceed if you are the developer of 'zendesk'.                  #
> # Otherwise, delete the file at /Users/arun/.op/plugins/local/zendesk. #
> ########################################################################
> Subdomain: example
> Email: example@example.com
> Password:
> ```
> 
> Also, should we switch from `Username` to `Email` as that matches the usage throughout [this documentation page](https://developer.zendesk.com/documentation/apps/getting-started/using-zcli/) and the `zcli login -i` flow?

@arunsathiya Just changed both, the fieldnames are now `Subdomain` and `Email`. Thanks for the feedback!

hculea (approved):
Code looks 👌

arunsathiya (approved):

AndyTitu (approved):
lgtm

PR #218, opened 2023-03-23
Make Sourcegraph endpoint optional

florisvdg:
The [Sourcegraph](https://docs.sourcegraph.com) plugin currently _requires_ the endpoint field, but I'm not sure that's necessary because Sourcegraph also has a main API. 

@arunsathiya Was there a reason for this, or can we make it optional?

### DIFF
--- a/plugins/sourcegraph/access_token.go
+++ b/plugins/sourcegraph/access_token.go
@@ -18,6 +18,7 @@ func AccessToken() schema.CredentialType {
 				Name:                fieldname.Endpoint,
 				MarkdownDescription: "Base URL for your Sourcegraph instance.",
 				Secret:              false,
+				Optional:            true,
 				Composition: &schema.ValueComposition{
 					Charset: schema.Charset{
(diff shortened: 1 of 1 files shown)


florisvdg:
image

Yep, seems to be optional indeed.

arunsathiya:
Good spot, @florisvdg. It does seem that if unset, it'd default to the main Sourcegraph instance:

```
➜  ~ src --help
src is a tool that provides access to Sourcegraph instances.
For more information, see https://github.com/sourcegraph/src-cli

Usage:

	src [options] command [command options]

Environment variables
	SRC_ACCESS_TOKEN  Sourcegraph access token
	SRC_ENDPOINT      endpoint to use, if unset will default to "https://sourcegraph.com"
```

I was under the impression that it's mandatory because the default endpoint wasn't mentioned on the docs here:

https://docs.sourcegraph.com/cli/explanations/env

hculea (approved):

PR #219, opened 2023-03-24
fix(tea): use `os.UserConfigDir` to find config

mcornick:
On macOS, `tea` writes its configuration to
`~/Library/Application Support/tea` and not `~/.config/tea`. This breaks the tea plugin. Go provides `os.UserConfigDir` to abstract the location of the config directory, so use that to fix macOS while not breaking Linux. In case `os.UserConfigDir` fails, fall back to `~/.config/tea` (this shouldn't happen unless something weird is going on, like the user's $HOME is not set.)

### DIFF
--- a/plugins/gitea/personal_access_token.go
+++ b/plugins/gitea/personal_access_token.go
@@ -2,6 +2,7 @@ package gitea
 
 import (
 	"context"
+	"os"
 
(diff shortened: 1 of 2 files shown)


AndyTitu:
Looks good to me!

AndyTitu (approved):

PR #220, opened 2023-03-24
Search alternative config directories when building a plugin

williamhpark:
## Summary

Currently, the config directory for shell plugins is hard-coded as `~/.op`. Whenever a user builds a plugin (e.g. `make aws/build`), the plugin executable gets stored at `~/.op/plugins/local`. However, depending on the user's environment, `~/.op` will not always necessarily be the config directory that's used by the CLI. The [CLI documentation](https://developer.1password.com/docs/cli/config-directories/) lists the order of precedence for possible configuration directory locations. We should search for an existing config directory based on that order of precedence and use it if it exists, for the sake of consistency and to avoid situations where multiple config directories get created for users of the CLI that also build their own shell plugins, which could result in issues like this https://www.reddit.com/r/1Password/comments/11jgm6j/error_from_1pw_cli_while_inspecting_gh/.

## How to Test

1. Check the values for the `OP_CONFIG_DIR`, `HOME`, and `XDG_CONFIG_HOME` env variables on your system. 
2. Build a shell plugin, e.g. `make aws/build`. Based on the env var values you found previously, the plugin executable should have been stored in the correct location based on the order of precedence.

## Additional Notes

For general `op` commands when using the CLI, users could specify a custom config directory using the `--config` flag, which takes the highest priority. However, since building for the shell plugins is done using a Makefile, I couldn't think of a good way that a custom config location could be specified in the build flow. Even with this gap, this PR should cover the majority of use cases since I don't anticipate many users rely on the `--config` flag. However, if this is too big of an assumption please let me know.

### DIFF
--- a/cmd/contrib/scripts/config_dir_getter.go
+++ b/cmd/contrib/scripts/config_dir_getter.go
@@ -10,16 +10,18 @@ import (
 
 // Retrieve the config directory to read or write to
 func main() {
-	opConfigDir, _ := os.LookupEnv("OP_CONFIG_DIR")
-	xdgConfigHome, _ := os.LookupEnv("XDG_CONFIG_HOME")
-	home, _ := homedir.Dir()
-
 	// This logic is based on the order of precedence outlined in the CLI documentation:
 	// https://developer.1password.com/docs/cli/config-directories
-	configDirPaths := []string{}
-	if opConfigDir != "" {
-		configDirPaths = append(configDirPaths, opConfigDir)
+
+	// If OP_CONFIG_DIR is set, use it immediately
+	if opConfigDir, _ := os.LookupEnv("OP_CONFIG_DIR"); opConfigDir != "" {
+		fmt.Print(opConfigDir)
+		return
 	}
+
+	xdgConfigHome, _ := os.LookupEnv("XDG_CONFIG_HOME")
+	home, _ := homedir.Dir()
+	configDirPaths := []string{}
 	if home != "" {
 		// Legacy home
 		configDirPaths = append(configDirPaths, filepath.Join(home, ".op"))
@@ -45,8 +47,8 @@ func main() {
 		}
 	}
 
-	// If we reach this point then none of those directories exist (op is executed
-	// for the first time). Default to the last entry in the list.
+	// None of those directories exist and OP_CONFIG_DIR is not set
+	// Default to the last entry in the list
 	if len(configDirPaths) > 0 {
 		fmt.Print(configDirPaths[len(configDirPaths)-1])
 	}
--- a/cmd/contrib/scripts/config_dir_getter.go
+++ b/cmd/contrib/scripts/config_dir_getter.go
@@ -16,8 +16,6 @@ func main() {
 
 	// This logic is based on the order of precedence outlined in the CLI documentation:
 	// https://developer.1password.com/docs/cli/config-directories
-	// Note: At the moment, if the --config flag is being used to specify the config
-	// directory then this script does not search in that location
 	configDirPaths := []string{}
 	if opConfigDir != "" {
 		configDirPaths = append(configDirPaths, opConfigDir)
--- a/cmd/contrib/scripts/main.go
+++ b/cmd/contrib/scripts/main.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
+)
+
+// Retrieve the config directory to read or write to
+func main() {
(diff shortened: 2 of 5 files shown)


cmd/contrib/scripts/main.go
```diff
@@ -0,0 +1,55 @@
...
+// Retrieve the config directory to read or write to
+func main() {
+	opConfigDir, _ := os.LookupEnv("OP_CONFIG_DIR")
+	xdgConfigHome, _ := os.LookupEnv("XDG_CONFIG_HOME")
+	home, _ := homedir.Dir()
+
+	// This logic is based on the order of precedence outlined in the CLI documentation:
+	// https://developer.1password.com/docs/cli/config-directories
+	// Note: At the moment, if the --config flag is being used to specify the config
+	// directory then this script does not search in that location
```
AndyTitu, line 20: This script would only be executed in the `make plugin/build` command, where no `--profile flags` are to be set. Did you mean that this function could detect what `--profile` flag the user usually runs their `op` commands with? I think that is not something we could or should be looking at anyway. If a user keeps setting the `--profile flag` all the time, then they'll probably want to consider setting the `OP_CONFIG_DIR` env var anyway. 

So I think we could safely remove this Note:

cmd/contrib/scripts/main.go
AndyTitu, line 1: I would call this file `config_dir_getter.go`

AndyTitu:
Awesome addition, thanks! This should fix most config dir inconsistencies related to local shell plugin development.

cmd/contrib/scripts/main.go
williamhpark: I elaborated on this in the `Additional Notes` section of the PR. It's related to the use of the `--config` flag to specify the config directory location, not `--profile`. But yep I agree that removing the Note should be safe, done!

cmd/contrib/scripts/config_dir_getter.go
```diff
@@ -0,0 +1,53 @@
...
+	for _, configDir := range configDirPaths {
+		fileInfo, err := os.Stat(configDir)
+		if err == nil && fileInfo.IsDir() {
+			fmt.Print(configDir)
+			return
+		}
+	}
+
+	// If we reach this point then none of those directories exist (op is executed
+	// for the first time). Default to the last entry in the list.
```
hculea, line 49: I suggest we should skip this step, if an environment variable has been set, but the directory has not been found.

williamhpark: Would you mind elaborating? 

From my understanding, this code is necessary. If a user has just started working with the 1Password CLI for example and runs `make <plugin>/build` for the first time, in which case no directories would be found, this block of code should run and create the directory.

hculea: In its current form, the code would still create one of our predefined locations, if the directory under `OP_CONFIG_DIR` doesn't exist. That is, if the user mistypes the value under `OP_CONFIG_DIR`, they would expect no side effects if the directory pointed to by the envvar doesn't exist. However, a directory (the last in the list of viable dirs) gets created, if no other exists.

williamhpark: Sorry I'm still a bit confused about the desired behaviour. If the user has defined `OP_CONFIG_DIR` but the directory doesn't exist, would you want the `make <plugin>/build` command to fail entirely?

If the end of the `config_dir_getter.go` script is reached and nothing is returned/printed, then when `make <plugin>/build` is called, in the Makefile `config_dir` evaluates to a blank string and `plugins_dir` evaluates to `/plugins/local`. A command is then called to create a directory at `/plugins/local`, which is even more unexpected than creating a new directory at one of the predefined locations, `~/.op/plugins/local` for instance. 

Since error handling isn't really something you can do within a Makefile, the current implementation was the best solution I could think of. Also, I should note that I matched the logic currently used in the `op` repo for consistency.

williamhpark: I re-tested how `op` searches for a config directory and it seems that if `OP_CONFIG_DIR` is set, the location specified by `OP_CONFIG_DIR` is used even if other predefined config directories already exist. This was something I missed when I first created this PR - I pushed a change to match the behaviour seen in `op` since the behaviour in both repos should match.

AndyTitu (approved):
lgtm. Ready to merge after addressing last concerns.

williamhpark:
> lgtm. Ready to merge after addressing last concerns.

I believe that I've addressed the last concerns [here](https://github.com/1Password/shell-plugins/pull/220#discussion_r1157821569). @hculea is there anything else that you think needs to be addressed?

Issue #216, opened 2023-03-23
Switch to env var provisioners for ngrok


AndyTitu:
### Platform or tool

Ngrok

### Desired or expected behavior

Authtoken and api key can be provisioned via [the newly exposed env vars](https://ngrok.com/docs/ngrok-agent/#supported-env-variables) NGROK_AUTH_TOKEN and NGROK_API_KEY.

We need to realise that NGROK_API_KEY was recently introduced so maybe not all ngrok users will benefit from this yet. We should consider defining a `versionChecker` inside our plugins that would only allow the plugin to be run if the user's ngrok version is above a certain threshold.

### Current behavior

Ngrok provisions secrets using the config file. This is not as straightforward as provisioning via env vars because it involves handling already existing non-sensitive config files. It requires logic for looking at default locations as well looking at `--config` flag. The case when multiple config flags are specified is not supported because the merging behaviour does not fit with our FIFO file provisioning.

arunsathiya:
- Just a note that I have a patch ready for it here: https://github.com/1Password/shell-plugins/pull/222

PR #222, opened 2023-03-27
update(ngrok): switch to envvars provisioner for ngrok 3.2.1 and higher

arunsathiya:
Fixes https://github.com/1Password/shell-plugins/issues/216 and fixes https://github.com/1Password/shell-plugins/issues/274

## Description

This PR updates the ngrok Shell Plugin to use environment variables to provision credentials when the ngrok CLI is 3.2.1 or higher. We'd retain the config file-based provisioner for older versions.

## Testing instructions

- Make sure there isn't a ngrok config file on your computer at [the path depending on your OS](https://ngrok.com/docs/ngrok-agent/config/#default-locations). (This is important because as of https://github.com/1Password/shell-plugins/pull/194, we detect such configuration files and merge with 1Password-generated configuration file)
- Check ngrok version by running `ngrok --version`. If it's 3.2.1 or higher, any `ngrok api` command (example `ngrok api tunnels list`) should succeed. During this run, the credentials are provisioned using the NGROK_AUTH_TOKEN and NGROK_API_KEY environment variables. (Note that [only NGROK_API_KEY is new as of 3.2.1](https://ngrok.com/docs/ngrok-agent/changelog/#ngrok-agent-321---2023-03-13), NGROK_AUTH_TOKEN was already supported in previous versions)
- If the version is 3.1.1 or older, `ngrok api tunnels list` should still succeed, but this time the provisioning happens with configuration files.

## Another way to test

- Checkout this branch and switch to it.
- In the file `plugins/ngrok/provisioner.go`, change `return fileProvisioner{}` to `return provision.EnvVars(defaultEnvVarMapping)` (line 64)
- With ngrok 3.1.1, run `ngrok api tunnels list`. It should fail with `ERROR:  API key is missing; either use '--api-key' flag or set it as 'api_key' in the configuration file` because at this point the provisioning happens with the env vars NGROK_AUTH_TOKEN and NGROK_API_KEY of which NGROK_API_KEY is not supported on 3.1.1

## Download ngrok 3.1.1

To download ngrok 3.1.1, download the binary from one of these two links depending on your system architecture:

https://github.com/ngrok/homebrew-ngrok/blob/668bb053898e0f767916083f3fdeaf4a0f6eac84/Casks/ngrok.rb#L4-L10

And then run `sudo unzip ~/Downloads/ngrok-v3-3.1.1-darwin-arm64.zip -d /usr/local/bin`

Ensure you are running the ngrok binary that you downloaded now by running `which ngrok`. You don't want to be testing with another binary, say Homebrew's version if you had it installed previously. `which ngrok` should read `/usr/local/bin/ngrok`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
 	github.com/golang/protobuf v1.3.4 // indirect
 	github.com/gsterjov/go-libsecret v0.0.0-20161001094733-a6f4afe4910c // indirect
 	github.com/hashicorp/go-hclog v0.14.1 // indirect
+	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/yamux v0.0.0-20180604194846-3520598351bb // indirect
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51 // indirect
 	github.com/mattn/go-colorable v0.1.9 // indirect
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,6 @@ require (
 	github.com/golang/protobuf v1.3.4 // indirect
 	github.com/gsterjov/go-libsecret v0.0.0-20161001094733-a6f4afe4910c // indirect
 	github.com/hashicorp/go-hclog v0.14.1 // indirect
-	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/yamux v0.0.0-20180604194846-3520598351bb // indirect
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51 // indirect
 	github.com/mattn/go-colorable v0.1.9 // indirect
@@ -52,6 +51,7 @@ require (
 	github.com/oklog/run v1.0.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 // indirect
+	golang.org/x/mod v0.9.0
 	golang.org/x/net v0.8.0 // indirect
 	golang.org/x/sys v0.6.0 // indirect
 	golang.org/x/term v0.6.0 // indirect
--- a/plugins/ngrok/provisioner.go
+++ b/plugins/ngrok/provisioner.go
@@ -41,9 +41,8 @@ func ngrokProvisioner() sdk.Provisioner {
 	//
 	// semver.Compare resulting in 0 means 3.2.1 is in use
 	// semver.Compare resulting in +1 means >3.2.1 is in use
-	if semver.Compare(currentVersion, envVarAuthVersion) == 0 || semver.Compare(currentVersion, envVarAuthVersion) == +1 {
-		newNgrokEnvVarProvisioner := ngrokEnvVarProvisioner{}
-		newNgrokEnvVarProvisioner.Provision(context.Background(), sdk.ProvisionInput{}, &sdk.ProvisionOutput{})
+	if semver.Compare(currentVersion, envVarAuthVersion) >= 0 {
+		return ngrokEnvVarProvisioner{}
 	}
 
 	// Otherwise use config file to provision credentials
--- a/plugins/ngrok/provisioner.go
+++ b/plugins/ngrok/provisioner.go
@@ -2,6 +2,7 @@ package ngrok
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -11,7 +12,6 @@ import (
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
-	"github.com/1Password/shell-plugins/sdk/provision"
 	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
 	"golang.org/x/mod/semver"
 	"gopkg.in/yaml.v3"
@@ -28,37 +28,24 @@ type fileProvisioner struct {
 }
 
 func ngrokProvisioner() sdk.Provisioner {
-	cmd := exec.Command("ngrok", "--version")
-	ngrokVersion, err := cmd.Output()
+	currentVersion, requiredVersion, err := getNgrokVersion()
 	if err != nil {
+		// When ngrok version check fails for any reason,
+		// use config file to provision as a fallback
 		return fileProvisioner{}
 	}
 
-	// Example: "ngrok version 3.1.1\n" to "3.1.1\n"
-	currentVersion := strings.TrimPrefix(string(ngrokVersion), "ngrok version ")
-
-	// Example: "3.1.1\n" to "3.1.1"
-	currentVersion = strings.Trim(currentVersion, "\n")
-
-	// Example: "3.1.1" to "v3.1.1" as that's the format
-	// the package semver expects
-	currentVersion = "v" + currentVersion
-
-	// NGROK_API_KEY is supported only from ngrok 3.2.1
-	// NGROK_AUTH_TOKEN was already supported
-	requiredVersion := "v3.2.1"
-
 	// If the current ngrok CLI version is 3.2.1 or higher,
-	// use environment variables to provision the Shell Plugin credentials
+	// use environment variables to provision credentials
 	//
 	// semver.Compare resulting in 0 means 3.2.1 is in use
 	// semver.Compare resulting in +1 means >3.2.1 is in use
 	if semver.Compare(currentVersion, requiredVersion) == 0 || semver.Compare(currentVersion, requiredVersion) == +1 {
-		return provision.EnvVars(defaultEnvVarMapping)
+		newNgrokEnvVarProvisioner := ngrokEnvVarProvisioner{}
+		newNgrokEnvVarProvisioner.Provision(context.Background(), sdk.ProvisionInput{}, &sdk.ProvisionOutput{})
 	}
 
-	// If the current ngrok CLI version less than 3.2.1,
-	// use configuration files to provision the credentials
+	// Otherwise use config file to provision credentials
 	return fileProvisioner{}
 }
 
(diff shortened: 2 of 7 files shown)


AndyTitu:
Thanks for this great addition @arunsathiya ! I really like the logic of using the env var provisioning from versions equal or higher than 3.2.1, and still use the file provisioner based on versions lower than that, so that the shell plugin works for ngrok users regardless of their installed version.

> The CI test is failing with 2023/03/27 08:33:22 exec: "ngrok": executable file not found in $PATH but that's expected I think because ngrok binary isn't available on the machine that runs the test.

For this reason, I think an improvement we could make on the OP CLI side is to also provision the tool's version via the `ProvisionInput` struct. It should not be the plugin's responsibility to detect the tool on the user's machine. Even more so since the CLI handles tool detection already, for now just to detect and error if a user is running a shell-plugin for a tool without having the actual tool in their PATH.
What do you think @SimonBarendse @florisvdg ?

florisvdg:
@AndyTitu That fits nicely into a plan I had a while back to make the Provisioner metadata more intelligent, which you can find on [this branch](https://github.com/1Password/shell-plugins/compare/main...provisioner-info).

plugins/ngrok/provisioner.go
```diff
@@ -3,19 +3,23 @@ package ngrok
...
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
 	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+	"github.com/hashicorp/go-version"
```
SimonBarendse, line 17: There is also semver functionality in the stdlib: https://pkg.go.dev/golang.org/x/mod/semver#Compare. What are the differences? Could/should we use that instead?

SimonBarendse (commented):
I think it would be great to pull out the logic for version comparison to make it easier to re-use. Worth exploring if we can go even further than the code on that branch Floris pushed and also take care of parsing the version, and maybe even the comparison func, so all the user needs to do is pass the version number.

In order to support non-standard versioning schemes (i.e. not semver) or with non-standard version command output that requires custom parsing logic, we can make it possible to pass funcs to overwrite the default behavior. but in the most common scenarios the user/plugin should require as little as possible work/code.

I believe making the code re-usable is an additive improvement, and doesn't need to block the improvement in this PR to go out. Can we already release this work from @arunsathiya then we (1Password) can take on the internal changes to add the version meta support in 1Password CLI to make it re-usable for other plugins?

arunsathiya:
Thanks for the review, all of you! This should be ready for one more look, and hopefully good to merge. 🤞🏼

hculea:
I have one concern related to this MR - I think it ingrains too deeply the version comparison functionality in the `ngrok` provisioner. 

The pattern where different versions of a tool would default to different provisioners should not be isolated only to `ngrok` - but rather should be a core feature of the plugin ecosystem.

For example, I can imagine something along the lines of:
```
DefaultProvisioner: provision.ExecDependent(
  provision.ForVersion("<=2.3.5", provision.EnvVars(envVarMapping),
  provision.ForVersion(">2.3.5", provision.TempFile(...)
)
```
together with a validation function that would assert that all versions are covered.

I'd love to hear what y'all think!
---
An additional question: are there any known limitations to the current temp file approach, when it comes to ngrok?

florisvdg:
A helper function to switch provisioners based on certain conditions could be useful indeed, but I agree with Simon that it doesn't have to block this PR.

arunsathiya:
>An additional question: are there any known limitations to the current temp file approach, when it comes to ngrok?

As far as I know/remember, the only concern is that the temp file approach requires a "Version" field, without which the ngrok binary doesn't work. At this point, "version: 2" is a valid field, but when in the future ngrok moves to a different version as the required/default version, we'd have to make that change on the Shell Plugins side too. Environment variables-based approach doesn't require that field.

hculea (approved):

hculea:
Very eager to bring versioning to all plugin provisioners!

Before then, I agree, let's not hold this back. 😄

AndyTitu (approved):
lgtm. I have created an internal issue for tackling the versioning logic discussed here.

plugins/ngrok/provisioner.go
```diff
@@ -24,7 +27,38 @@ const (
...
+	// NGROK_AUTH_TOKEN was already supported
+	requiredVersion := "v3.2.1"
+
+	// If the current ngrok CLI version is 3.2.1 or higher,
+	// use environment variables to provision the Shell Plugin credentials
+	//
+	// semver.Compare resulting in 0 means 3.2.1 is in use
+	// semver.Compare resulting in +1 means >3.2.1 is in use
+	if semver.Compare(currentVersion, requiredVersion) == 0 || semver.Compare(currentVersion, requiredVersion) == +1 {
+		return provision.EnvVars(defaultEnvVarMapping)
```
AndyTitu, line 57: @arunsathiya  another thing I'd like to see tackled in this PR, if you have enough time before a security/Sr. review is performed on this PR, is to :
1. Create a wrapper env var provisioner for ngrok. In its provision function it will check the version (as you already do here) but it will return an error if the version is lower than 3.2.1.
2. Add this wrapper env var provisioner as the `DefaultProvisioner`, while using this function `ngrokProvisioner()` in the credential Usage definition of the `ngrok` executable definition. 

What this means is that: 
- ngrok and only ngrok will use either the file provisioner or the env var provisioner
- any other tool that will use ngrok's credential type in the future will only use the env var provisioner. Imagine that we release a terraform plugin that needs to work with ngrok, and therefore will re-use the ngrok credential to provision the ngrok secrets. If terraform ends up using the file provisioner we'll end up with `terraform plan --config=...`  which is not an actual terraform command, because the ngrok file provisioner also appends this config flag to the command line of the executable. That is why the ngrok file provisioner should just remain specific to the ngrok tool.

plugins/ngrok/provisioner.go
```diff
@@ -95,6 +116,31 @@ func getConfigValueAndNewArgs(args []string, newFilePath string) (string, []stri
...
+	// Example: "3.1.1\n" to "3.1.1"
+	currentVersion = strings.Trim(currentVersion, "\n")
+
+	// Example: "3.1.1" to "v3.1.1" as that's the format
+	// the package semver expects
+	currentVersion = "v" + currentVersion
+
+	// NGROK_API_KEY is supported only from ngrok 3.2.1
+	// NGROK_AUTH_TOKEN was already supported, but NGROK_API_KEY is a new addition
+	requiredVersion := "v3.2.1"
```
AndyTitu, line 139: Given that this is a constant version that won't change, I think it makes more sense to make it a `const` at the top of the file and call it something like `envVarAuthVersion` to make it clear that we are using it to determine when we can use env var provisioning.

plugins/ngrok/provisioner.go
```diff
@@ -95,6 +116,31 @@ func getConfigValueAndNewArgs(args []string, newFilePath string) (string, []stri
 	return "", append(args, fmt.Sprintf("--config=%s", newFilePath))
 }
 
+// Get installed ngrok version and required version
+func getNgrokVersion() (string, string, error) {
```
AndyTitu, line 120: Then this function can return only `(string, error)`

plugins/ngrok/provisioner.go
```diff
@@ -24,7 +27,25 @@ const (
...
+		// use config file to provision as a fallback
+		return fileProvisioner{}
+	}
+
+	// If the current ngrok CLI version is 3.2.1 or higher,
+	// use environment variables to provision credentials
+	//
+	// semver.Compare resulting in 0 means 3.2.1 is in use
+	// semver.Compare resulting in +1 means >3.2.1 is in use
+	if semver.Compare(currentVersion, requiredVersion) == 0 || semver.Compare(currentVersion, requiredVersion) == +1 {
```
AndyTitu, line 43: I believe `compare` can only return -1, 0 or 1 so here we can just say:
```suggestion
	if semver.Compare(currentVersion, requiredVersion) >= 0 {
```

plugins/ngrok/provisioner.go
```diff
@@ -24,7 +27,25 @@ const (
...
+		newNgrokEnvVarProvisioner := ngrokEnvVarProvisioner{}
+		newNgrokEnvVarProvisioner.Provision(context.Background(), sdk.ProvisionInput{}, &sdk.ProvisionOutput{})
```
AndyTitu, line 45: At this point we should not call the `Provision` method but rather just return the `ngrokEnvVarProvisioner{}`. 

Provisioning is generally never done inside the shell plugin code. The CLI code must call the `Provision` function, and the shell plugin's code reponsibility is to provide the CLI with the right provisioner to call.

go.mod
```diff
@@ -51,6 +51,7 @@ require (
 	github.com/oklog/run v1.0.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 // indirect
+	golang.org/x/mod v0.9.0
```
AndyTitu, line 54: why is this required here?

AndyTitu (commented):
This generally looks good! I have left a few minor comments. 

It would be good to test these changes with the Terraform plugin to make sure using `ngrok Credentials` with `Terraform` no longer returns the `not supported` error.

AndyTitu (commented):
This generally looks good! I have left a few minor comments. 

It would be good to test these changes with the Terraform plugin to make sure using `ngrok Credentials` with `Terraform` no longer returns the `not supported` error.

arunsathiya:
Thanks for the guidance, @AndyTitu. I have tested with the Terraform plugin and it works as expected. 

image

My config file is as below:

```
terraform {
  required_providers {
    ngrok = {
      source = "ngrok/ngrok"
      version = "0.1.4"
    }
  }
}

# Configure the ngrok provider
provider "ngrok" {}

resource "ngrok_api_key" "example" {
  description = "ad-hoc dev testing"
  metadata = "{\"environment\":\"dev\"}"
}
```

The resource block is from here:

https://github.com/ngrok/terraform-provider-ngrok/tree/main/examples/resources

arunsathiya:
Fixes https://github.com/1Password/shell-plugins/issues/274

go.mod
AndyTitu: Oh, I see this is for the `semver` dependency

arunsathiya:
Testing instructions:

- Install ngrok 3.2.1, a higher version, and a lower version (using [the original testing instructions here](https://github.com/1Password/shell-plugins/pull/222#issue-1641641735)) and see that provisioning credentials work as expected for all of these. 3.2.1 and higher should use environment variables, while smaller ones, like 3.1.1 should use file provisioner.
- Create a Terraform config file `main.tf` with the content below.
- Run `terraform init`
- Run `op plugin init terraform` and choose ngrok credentials. 
- Run `terraform plan` followed by `terraform apply`. 
- See that the application goes through as expected.

Terraform config file:

```
terraform {
  required_providers {
    ngrok = {
      source = "ngrok/ngrok"
      version = "0.1.4"
    }
  }
}

# Configure the ngrok provider
provider "ngrok" {}

resource "ngrok_api_key" "example" {
  description = "ad-hoc dev testing"
  metadata = "{\"environment\":\"dev\"}"
}
```

PR #228, opened 2023-03-28
Add ManagementURL for Sourcegraph plugin

SimonBarendse:
In the dashboard the management URL is dynamic; It contains the username: https://sourcegraph.com/users/<username>/settings/token

That's probably why it wasn't included in the first iteration of this plugin.

I just found that https://sourcegraph.com/user/settings/tokens redirects to the management URL for the currently logged in user. Sourcegraph also uses this in the 'src login' command, so it should be safe to use that here as well.

### DIFF
--- a/plugins/sourcegraph/access_token.go
+++ b/plugins/sourcegraph/access_token.go
@@ -11,8 +11,9 @@ import (
 
 func AccessToken() schema.CredentialType {
 	return schema.CredentialType{
-		Name:    credname.AccessToken,
-		DocsURL: sdk.URL("https://docs.sourcegraph.com/cli"),
+		Name:          credname.AccessToken,
+		DocsURL:       sdk.URL("https://docs.sourcegraph.com/cli"),
+		ManagementURL: sdk.URL("https://sourcegraph.com/user/settings/tokens"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.Endpoint,


hculea (approved):

AndyTitu (approved):
> I just found that https://sourcegraph.com/user/settings/tokens redirects to the management URL for the currently logged in user.
Nice touch

arunsathiya:
This is a great addition for now, but if I may note something for our future reference, this assumes that access tokens created are for the Dotcom instance (that is, Sourcegraph.com). But for self-hosted users of the Sourcegraph software, the actual link would be https://example.com/users/settings/token. It's for that reason, I hadn't included ManagementURL in the original version:

- https://github.com/1Password/shell-plugins/pull/146

Also, I remember that visiting https://sourcegraph.com/users/settings/token itself had resulted in a 404 at the time of my PR. It's not clear when that changed based on these recent PRs 🤷🏽‍♂️ 🤔 

- https://github.com/sourcegraph/sourcegraph/pulls?q=is%3Apr+%22access+token%22+is%3Aclosed

In any case, I am on board with us retaining the current version of ManagementURL. 💯 When in the future Shell Plugins can dynamically capture the host address/domain at the time of setup (that is, `op plugin init src`), the ManagementURL would need to be updated to support/adapt that new domain. In fact, even Sourcegraph's CLI out of the box assumes that everyone uses the Dotcom version:

image

arunsathiya:
Also, I just stumbled upon this issue which demoes how Sourcegraph CLI captures the Sourcegraph instance address at runtime/setup and then generates the appropriate ManagementURL:

- https://github.com/sourcegraph/src-cli/issues/716

image

That's the approach I'd love to see Shell Plugins take one day. I am happy to take a stab at how we can tackle this as well.

PR #229, opened 2023-03-28
AWS Vault importer for the AWS shell plugin

williamhpark:
## Summary

[AWS Vault](https://github.com/99designs/aws-vault) is a popular tool among AWS users used to store/provision AWS credentials in a vaulting backend such as macOS Keychain. Chances are if we want to convince an AWS user to also use the AWS shell plugin then they would need to migrate their credentials from AWS Vault. Therefore, we need to create an importer for the AWS access keys, that would fetch those from the user's vaulting backend and put them in 1Password. Luckily, AWS Vault is also open source, so we can leverage their own functionality for retrieving credentials.

## Thought Process

When a profile is added through `aws-vault`, a section is added to the AWS config file usually located at `~/.aws/config`. The AWS Vault importer iterates through the config file, searches the user's vaulting backend for matching credentials (taking advantage of functionality defined in the AWS Vault repo), then imports them to 1Password.

~~Since AWS Vault relies on the AWS config file to function, designing the importer as a file importer seemed intuitive.~~

The AWS region name for an AWS profile can be specified in `~/.aws/config`, for specific profiles and/or for the `default` profile. When an AWS Vault profile is imported, the region that's used for the item is determined as follows:

| AWS Vault profile region specified? | `default` profile region specified? | Region used |
|-------------------------------------|-------------------------------------|-------------|
| Yes                                 | Yes                                 | AWS Vault profile   |
| Yes                                 | No                                  | AWS Vault profile   |
| No                                  | Yes                                 | `default`   |
| No                                  | No                                  | None        |

## How To Test

1. Install `aws-vault` following the instructions outlined [here](https://github.com/99designs/aws-vault)
2. Add an AWS profile to your keychain through `aws-vault`, i.e. `aws-vault add <profile>`
3. Run `op plugin init aws` > `Import into 1Password...` > enter your keychain password(s) when prompted. You should see options to import from `<keychain> (<profile>)`. If there are multiple AWS Vault profiles, each one should be listed
4. Select a profile to import and complete the import flow. An `AWS Access Key (<profile>)` item should be created in 1Password
5. Run an `aws` command that requires authentication, e.g. `aws s3 ls`, and use the newly created item to authenticate. The command should run successfully
6. Test that the logic for determining the item's region, outlined in "Thought Process", is followed

image


plugins/aws/aws_vault_importer.go
```diff
@@ -0,0 +1,82 @@
...
+						fields[fieldname.DefaultRegion] = section.Key("region").Value()
+					} else if defaultRegion != "" {
+						fields[fieldname.DefaultRegion] = defaultRegion
+					}
+
+					// Only add candidates with required credential fields
+					if fields[fieldname.AccessKeyID] != "" && fields[fieldname.SecretAccessKey] != "" {
+						out.AddCandidate(sdk.ImportCandidate{
+							Fields:   fields,
+							NameHint: importer.SanitizeNameHint("aws-vault - " + profileName),
```
williamhpark, line 75: Thoughts on using this as the namehint?

williamhpark:
At the moment, `make aws/build` fails in this branch unless I run `go mod vendor` beforehand, which generates around 1.5K files in the `vendor` directory. Shall I commit these files?

plugins/aws/aws_vault_importer.go
```diff
@@ -0,0 +1,87 @@
...
+					}
+				}
+			}
+		}
+	})
+}
+
+func TryAWSVault(result func(ctx context.Context, in sdk.ImportInput, out *sdk.ImportAttempt)) sdk.Importer {
+	return func(ctx context.Context, in sdk.ImportInput, out *sdk.ImportOutput) {
+		attempt := out.NewAttempt(sdk.ImportSource{AWSVault: true})
```
williamhpark, line 84: I added this new `AWSVault` import source. In order for this code to work, a change would need to be made in the 1Password CLI codebase to recognize this import source and output to the terminal accordingly, i.e. `AWS Vault (<profile>)`

plugins/aws/credentials_importer.go
AndyTitu, line 1: Let's create only one `plugins/aws/importers.go` file where we put all our aws custom importer functions.

sdk/importer.go
```diff
@@ -33,8 +33,9 @@ func (c *ImportCandidate) Equal(other ImportCandidate) bool {
 }
 
 type ImportSource struct {
-	Env   []string
-	Files []string
+	Env      []string
+	Files    []string
+	AWSVault bool
```
AndyTitu, line 38: I think we need to figure out a better general way of dealing with import sources that are neither files or environment variables. In this case, we are dealing with importing credentials from password managers. My take is that we should add another option in `ImportSource` called `PasswordManagers []string` which would be a list of names of password managers that contain our desired credentials. 

The goal of this would be for the user to see the following when they are importing from password managers/vaults:

```
> Password Manager ( profile andy)  AWS-Vault - MacOS Keychain
   Password Manager ( profile dev)    AWS-Vault - MacOS Keychain
   File                                                      ~/.aws/credentials
```

However, since this is an SDK change that will require a small change in `op` as well, I'm fine with decoupling this from the current PR. For this PR, I would then be ok with the outcome:

```
> File (profile andy)   AWS-Vault - MacOS Keychain
   File (profile dev)      AWS-Vault - MacOS Keychain
```
Even though this is not nice, you could argue that the keychain works with encrypted files under the hood, so we can keep this until we introduce a more suitable import source.

What do you think @florisvdg ?

plugins/aws/aws_vault_importer.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			out.AddError(err)
+			return
+		}
+
+		// Use the CredentialKeyring struct from aws-vault to retrieve vaulting backend credentials
+		credentialKeyring := &vault.CredentialKeyring{Keyring: keyring}
+
+		// Load the AWS config file (default location at ~/.aws/config)
+		awsConfigFile, err := awsVault.AwsConfigFile()
```
AndyTitu, line 32: nice that you're using the aws-vault functionality for handling config files!

plugins/aws/aws_vault_importer.go
```diff
@@ -0,0 +1,87 @@
...
+		// Get the region specified for the "default" profile
+		var defaultRegion string
+		if defaultSection, ok := awsConfigFile.ProfileSection("default"); ok {
+			defaultRegion = defaultSection.Region
+		}
+
+		// Iterate through the profiles in the AWS config file and
+		// import any matching credentials stored in the vaulting backend
+		for _, profileName := range awsConfigFile.ProfileNames() {
+			if profileName != "default" {
```
AndyTitu, line 48: Is there any reason we are skipping the "default" profile? As far as I know, people can specify credentials under the default profile as well in the `.aws/credentials` file. And `aws-vault` should support working with the default credentials in the keyring as well.

AndyTitu (changes requested):
Nice work! Left a few comments.

plugins/aws/aws_vault_importer.go
```diff
@@ -0,0 +1,87 @@
...
+				creds, err := credentialKeyring.Get(profileName)
+				if err == nil {
```
AndyTitu, line 50: Let's make sure we are adding this error to the output's diagnostics if it is not nil. If a user specified a profile in their config file, they would expect to see an error if that profile's credentials are not to be found in the keyring.

plugins/aws/aws_vault_importer.go
AndyTitu: See my other comment on this: https://github.com/1Password/shell-plugins/pull/229#discussion_r1154138509

sdk/importer.go
florisvdg: As for import sources in this SDK, we can either introduce "Other" and have the plugin author give a name of the source (e.g. "macOS Keychain") to present in `op`. But we could introduce new sources for "macOS Keychain", "GNOME Keyring", and "KWallet" in the SDK. May be the nicer option.

The source value should indeed probably be/contain the profile name. But not sure if "AWS Vault" should even be mentioned at all here.

plugins/aws/aws_vault_importer.go
williamhpark: Thanks for catching that, done! FYI I made it so that the code doesn't `return` if fetching credentials for one profile fails, since it's possible that fetching credentials for other profiles that come later in the config file could still succeed.

By the way, currently, errors during importing/provisioning display like this:

image

Should we make it so that errors display in a more user-friendly way? This would be a change in the 1Password CLI code, from my understanding.

plugins/aws/aws_vault_importer.go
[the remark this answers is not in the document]
williamhpark: The recent updates that I made to the PR generally follow your suggestion, let me know your thoughts!

sdk/importer.go
williamhpark: For now, I made it so that a new `Other` import source is introduced in the SDK and used in the AWS Vault importer.

When a user attempts to import for AWS, the source value for the items retrieved from the vaulting backend are displayed as `<vaulting backend> (profile)`, with no second column. Again, this requires some updates in the 1Password CLI codebase as well:

image

Let me know your thoughts @AndyTitu @florisvdg!

plugins/aws/aws_vault_importer.go
williamhpark: Makes sense, sounds like a good idea to me too. Once the changes for this PR are finalized and merged, I'll create a new issue for the file importer.

By the way, the config file functionality that I used for the AWS Vault importer actually comes from the [AWS Vault SDK](https://pkg.go.dev/github.com/99designs/aws-vault/v7@v7.2.0), not the [AWS SDK](https://aws.amazon.com/sdk-for-go/), in case there was any confusion about that.

sdk/importer.go
florisvdg: Makes sense to me 👍 

One thing about how it's displayed in `op`: it would be nice if it would be spaced just like `File` and `Environment`.
```
macOS Keychain         work
macOS Keychain         home
File                   ~/.aws/credentials
```
Could be worth introducing something like:
```go
type CustomSource struct {
   Type  string
   Value []string
}
```
Or alternatively (maybe simpler), populate the `Other []string` value as: `[]string{"macOS Keychain", "work"}`.

plugins/aws/importers.go
```diff
@@ -0,0 +1,189 @@
...
+		if defaultSection, ok := awsConfigFile.ProfileSection("default"); ok {
+			defaultRegion = defaultSection.Region
+		}
+
+		// Iterate through the profiles in the AWS config file and
+		// import any matching credentials stored in the vaulting backend
+		for _, profileName := range awsConfigFile.ProfileNames() {
+			creds, err := credentialKeyring.Get(profileName)
+			if err != nil {
+				out.AddError(fmt.Errorf("could not retrieve credentials for %s from vaulting backend", profileName))
```
AndyTitu, line 146: small suggestion
```suggestion
				out.AddError(fmt.Errorf("could not retrieve credentials for profile %s from vaulting backend %s", profileName, string(awsVaultBackend)))
```

sdk/importer.go
AndyTitu: > One thing about how it's displayed in op: it would be nice if it would be spaced just like File and Environment.

Yes, let's do this.

> Could be worth introducing something like:
```
type CustomSource struct {
   Type  string
   Value []string
}
```

I like this clear definition the best

@williamhpark

plugins/aws/aws_vault_importer.go
AndyTitu: https://github.com/1Password/shell-plugins/issues/235 Created issue to track these changes.

plugins/aws/aws_vault_importer.go
AndyTitu: On a second consideration, I realised that we should rather treat a not found secret as we currently do for other credentials we are searching for in the import step: if we don't find them, just don't include them in the candidate list. What this means in our case is to check in this block whether or not the error that we get is the `credentials not found` one and in that case do not show an error, but rather simply skip these creds:

```
creds, err := credentialKeyring.Get(profileName)
 			if err != nil && strings.Contains(err.Error(), "not found") {
 				out.AddError(fmt.Errorf("could not retrieve credentials for %s from vaulting backend %s: %w", profileName, string(awsVaultBackend), err))
```

plugins/aws/importers.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		// If AWS_VAULT_BACKEND matches one of the available backends, use it
+		if backendType == awsVaultBackendEnvVar {
+			awsVaultBackend = awsVaultBackendEnvVar
+			break
+		}
+	}
+
+	return TryAWSVault(backendNames[awsVaultBackend], func(ctx context.Context, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		awsVault := &cli.AwsVault{}
```
AndyTitu, line 116:
```suggestion
				var keyringConfigDefaults = keyring.Config{
			ServiceName:              "aws-vault",
			LibSecretCollectionName:  "awsvault",
			KWalletAppID:             "aws-vault",
			KWalletFolder:            "aws-vault",
			KeychainTrustApplication: true,
			WinCredPrefix:            "aws-vault",
			KeychainName:             "aws-vault",
			FileDir:                  "~/.awsvault/keys/",
		}

		awsVault := &cli.AwsVault{KeyringConfig: keyringConfigDefaults}
```

Let's do this so the user doesn't have to manually add the aws-vault keychain to their Mac OS keychain list

plugins/aws/importers.go
williamhpark: Not necessary anymore because of [this change](https://github.com/1Password/shell-plugins/pull/229#discussion_r1156185534).

plugins/aws/importers.go
florisvdg: What does this do?

sdk/importer.go
williamhpark: The reason why I thought it would be better to format in just one column in `op` was to match what is currently done for the file importer, where the AWS profile name is displayed in parentheses next to the source type (`File (test)` in the screenshot):

image

This being said, I think that introducing the `CustomSource` that @florisvdg suggested would be useful for potential future `Other` sources anyways. But in this case, I feel inclined to keep it as one column by passing a blank string as the value, i.e.:

`attempt := out.NewAttempt(importer.SourceOther(keyringBackend, ""))`

plugins/aws/importers.go
williamhpark: Before this change, in order to use the AWS Vault importer on macOS, you would have to manually add the keychain file to your system keychain ("File -> Add Keychain" and select the `aws-vault.keychain-db`). But defining `KeychainName: "aws-vault"`  in the keyring config actually makes it so that this is no longer necessary; even if the user didn't add the keychain, which is definitely possible, the importing succeeds. 

The other config values (`ServiceName`, `LibSecretCollectionName `, etc.) would probably be good to keep as well, since these are the default config values used in the `aws-vault` codebase here: https://github.com/99designs/aws-vault/blob/master/cli/global.go.

sdk/importer.go
williamhpark: I pushed a change based on my last comment, let me know if this seems reasonable @AndyTitu @florisvdg.

plugins/aws/importers.go
```diff
@@ -0,0 +1,189 @@
...
+				out.AddCandidate(sdk.ImportCandidate{
+					Fields:   fields,
+					NameHint: importer.SanitizeNameHint(profileName),
+				})
+			}
+		}
+	})
+}
+
+func TryAWSVault(keyringBackend string, result func(ctx context.Context, in sdk.ImportInput, out *sdk.ImportAttempt)) sdk.Importer {
```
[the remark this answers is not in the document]
williamhpark: I pushed a change that merges the functionality into one function, let me know what you think.

williamhpark:
I realized that we should be searching through all available vaulting backends on the OS, not just one of them, so I pushed changes to the PR to reflect this. I tested this out by storing credentials in both the macOS Keychain and [pass](https://www.passwordstore.org/), and importing from both works.

image

plugins/aws/importers.go
```diff
@@ -0,0 +1,186 @@
...
+			// Get the region specified for the "default" profile
+			var defaultRegion string
+			if defaultSection, ok := awsConfigFile.ProfileSection("default"); ok {
+				defaultRegion = defaultSection.Region
+			}
+
+			// Iterate through the profiles in the AWS config file and
+			// import any matching credentials stored in the vaulting backend
+			for _, profileName := range awsConfigFile.ProfileNames() {
+				profileFound, _ := credentialKeyring.Has(profileName)
```
AndyTitu, line 149: here we should not ignore the error but rather do `out.AddErrror(err)` and `continue` right before we check `profileFound` so that a potential error can be logged even if we skip these credentials. The potential error can be something like `"No keychain is available. You may need to restart your computer."` (extracted from the actual aws-vault code)

sdk/importer.go
florisvdg: Ah gotcha. Okay to keep it consistent with `File` then.

I do think we need to revisit this approach in the future. Ideally the profile name for both of them would be in the column on the right, instead of in parentheses on the left.

plugins/aws/importers.go
florisvdg: Okay, as long as we're not creating any on the user's system by doing this check.

Would be good to link to that file in a code comment so that we always go back to where we got this from.

sdk/importer.go
AndyTitu: Hmm, I see.

I am still inclined to think specifying the value of the import source with the profile name would be slightly more suitable. It doesn't make a huge difference, I'll admit, but it seems just a bit more consistent.

Here is my reasoning based on looking at how a keychain access is organised
Screenshot 2023-04-04 at 20 29 58

1. The name of the keychain is `aws-vault (andy)`. If we strip `aws-vault` as we talked about we remain with just andy. In the case of a `File` importer `.aws/credentials` is also the name of the file. So it seems more consistent to treat the names (i.e "andy" and other profile names) as import source values, like we do with file names

2. We can notice a third element in the keychain which is `Sts.GetSessionToken` (temporary credentials). In this case we don't want to import those as well, but if we did would it make sense to add that as a name hint rather than as a name? That's not a profile anymore.

williamhpark: @AndyTitu I see what you mean for point 1. When stored in Pass through AWS Vault, the credentials are stored under the profile name as well (`pass-profile` in the ss) , which adds to your point that the profile name should be treated as the import source value:

image

To keep consistency with the file importer, I'm still inclined to keep the profile name as the name hint. Although it seems a bit redundant, what are your thoughts on having the profile name as both the name hint and source value? Like so:

image

williamhpark: Or another option is we could format like so, similar to what @AndyTitu suggested at the beginning of this thread:

```
> Password Manager (work)       macOS Keychain
  Password Manager (home)       Pass
  File                          ~/.aws/credentials
```

I think I like this best actually.

florisvdg: I don't think seeing `Password Manager` there is that helpful to the user. I'd suggest to stick with your current approach until we can think of something nicer in the future:

florisvdg: I'd say just the name hint. I don't see the value of duplicating it.

AndyTitu: 👍

AndyTitu (approved):
lgtm

hculea:
Let's make sure we take out the vendor'd dependencies out of the diff, before merging!

hculea (approved):

hculea (changes requested):
Requesting changes - just to make sure the vendored deps are removed. Good to merge after!

PR #232, opened 2023-03-29
New plugin: AWS CDK Toolkit

williamhpark:
## Summary

Add support for a new plugin for the AWS CDK Toolkit.

The AWS Cloud Development Kit (AWS CDK) lets you define your cloud infrastructure as code in one of its supported programming languages. You can use it to create, manage, deploy, etc. AWS CDK apps.

## Thought Process

The AWS CDK Toolkit uses the same credentials as the regular AWS CLI, and the same profile workflow. We can reuse the importers and provisioners that have already been created for the AWS shell plugin.

## How To Test

1. Install the AWS CDK Toolkit following [these instructions](https://docs.aws.amazon.com/cdk/v2/guide/cli.html)
2. Initialize the AWS CDK plugin by running `op plugin init cdk`
3. Create an AWS CDK app following [these instructions](https://docs.aws.amazon.com/cdk/v2/guide/hello_world.html)
4. Run a `cdk` command from the CDK app directory that requires authentication, e.g. `cdk deploy`. Try authenticating in the following ways. The command should be authenticated and run as expected:
  a. Use an existing `AWS Access Key` created for the regular AWS shell plugin
  b. Import credentials from the `~/.aws/credentials` file
  c. Import credentials manually as prompted
5. Run `make aws/validate`. All validation checks should pass and there should be a `# Executable: AWS CDK Toolkit` section


williamhpark:
There are conflicts between this PR and [this other PR for an AWS Vault importer](https://github.com/1Password/shell-plugins/pull/229) that should be resolved when it's time to merge.

AndyTitu (commented):
We would like to leverage the powers of `CredentialUsage` field of an `Executable` to specify that AWS CDK CLI executable could use the exact same `CredentialType` as `AWS` does. This will mean that by default, all default provisioners and importers will be re-used for cdk as well.

plugins/awscdk/plugin.go
```diff
@@ -0,0 +1,22 @@
...
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "awscdk",
+		Platform: schema.PlatformInfo{
+			Name:     "AWS CDK",
+			Homepage: sdk.URL("https://aws.amazon.com/cdk/"),
+		},
+		Credentials: []schema.CredentialType{
+			AccessKey(),
```
SimonBarendse, line 16: I'd expect this to have `aws.AccessKey()` or something along these lines so that it completely re-uses the credential, rather than specify its own version of the credential, since both AWS CLI and AWS CDK CLI are using the exact same AWS SDK to load the credential. It's the same credential definition in AWS, so ideally it could also have a single credential definition in the plugin ecosystem.

williamhpark:
> We would like to leverage the powers of `CredentialUsage` field of an `Executable` to specify that AWS CDK CLI executable could use the exact same `CredentialType` as `AWS` does. This will mean that by default, all default provisioners and importers will be re-used for cdk as well.

@AndyTitu I believe that the change that I just pushed conforms with that you're hoping for. Let me know if you think any other changes are necessary!

plugins/awscdk/plugin.go
williamhpark: I simplified the code so that the `cdk` executable is defined within the existing `aws` directory, rather than within a new plugin directory, which indirectly addresses your comment I believe :)

sdk/schema/plugin.go
```diff
@@ -78,12 +78,6 @@ func (p Plugin) Validate() (bool, ValidationReport) {
 		Severity:    ValidationSeverityError,
 	})
 
-	report.AddCheck(ValidationCheck{
-		Description: "Has no more than one executable defined. Plugins with multiple executables are not supported yet",
-		Assertion:   len(p.Executables) == 1,
-		Severity:    ValidationSeverityError,
-	})
```
hculea, line 85: Let's ensure that we fully add support for multiple executables on the CLI as well, I believe there may be a few instances in the codebase where we may not correctly handle cases where more than one exec is defined per plugin.

hculea, line 85: Another concern that I have about removing this check is that now, we may end up with a situation where a plugin contributor uses the latest version of `shell-plugin`'s main to develop a multi-exec plugin (which is already a requested feature e.g. https://github.com/1Password/shell-plugins/pull/141), but an older version of the CLI, if they were not to update to the latest.

In that case, there would be no actionable error message. I am wondering if we should be artificially triggering an error message by increasing the protocol version? Then they would get a "1Password CLI is outdated" error message when trying to use a locally built plugin against the CLI.

hculea: This, however, is a secondary concern, and a situation that I expect to be rather rare. I'm fine with leaving this for later or even address it only if it becomes a problem, but let's track it with a ticket internally 😄

florisvdg: If we're concerned about version mismatches then we could also upgrade the protocol version, which will result in a nicer error message.

plugins/aws/aws.go
```diff
@@ -23,3 +23,20 @@ func AWSCLI() schema.Executable {
...
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
+		Uses: []schema.CredentialUsage{
+			{
+				Name: credname.AccessKey,
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
williamhpark: Done!

sdk/schema/plugin.go
williamhpark: Following a functional test of the `op plugin` commands, I didn't run into any failures or unexpected behaviour caused by multiple executables. I also looked through the `op` plugin code and couldn't find any areas that require updating, since it was written with the expectation that multiple executables should be supported at some point. But if there's anything in particular you would like me to look into again, please let me know.

One thing though, when running `op plugin list`, or `op plugin init`/`op plugin inspect` without specifying the executable, I noticed that the name defined in `plugin.go` is the one that gets displayed. For instance, since the name defined in `/plugins/aws/plugin.go` is `AWS`, that's what gets displayed for both the `aws` and `cdk` executables, which doesn't seem ideal:

image

We could use `Executable.Name` in these cases instead of `Platform.Name`, but most of the plugin executable names have `CLI` as a suffix and we may not want this. BTW, there seems to be an issue displaying escape sequences for test build plugin names: 

image

In my opinion, continuing to display the `plugin.go` name as it is right now is probably fine, but I wanted to bring this up in case anyone disagrees.

AndyTitu: I think the current behaviour of `op plugin list` is the desired one

AndyTitu: opened https://github.com/1Password/shell-plugins/issues/236 to track the protocol upgrade separately

AndyTitu (approved):
Functionally tested for `op plugin run`, `op plugin init`, `op plugin ls`, `op plugin inspect`, and `op plugin clear`.

Code lgtm

plugins/aws/cdk.go
```diff
@@ -0,0 +1,25 @@
+package aws
```
SimonBarendse, line 1: Love how much simpler this now is with re-using the AWS credential definition 😍

SimonBarendse (approved):

PR #240, opened 2023-04-06
chore(docs): add `<p>` tag to force line break

mrjones2014:
If you go to the wide view of README.md https://github.com/1Password/shell-plugins/blob/main/README.md

The "Show all..." text is inline next to the image instead of below it. Surrounding with `<p>` fixes that.


SimonBarendse (approved):
Thanks Mat! ❤️

accraw (approved):

PR #244, opened 2023-04-13
Shell Plugin for Laravel Forge CLI

andresayej:
Shell Plugin for the Laravel Forge CLI

The [Laravel Forge CLI](https://forge.laravel.com/docs/1.0/cli.html) is the official command line tool for the [Laravel Forge Service](https://forge.laravel.com/) offered by Laravel LLC the company behind the popular open source PHP framework: [Laravel](https://laravel.com/)

The authentication mechanism is made via the API Token: `FORGE_API_TOKEN` based on the source code of the Forge CLI located here: https://github.com/laravel/forge-cli
This plugin also offers an importer for scanning the user's environment and file system for the needed credentials in the default config location allowing existing Forge CLI users to import their current token into 1Password.

Here are some screenshots with the following commands that need authentication:
```bash
forge site:list
```
```bash
forge server:list
```

<details><summary>Screenshots</summary>
<p>

CleanShot 2023-04-13 at 23 50 07@2x
CleanShot 2023-04-13 at 23 50 32@2x
CleanShot 2023-04-13 at 23 50 56@2x
CleanShot 2023-04-14 at 00 50 18@2x
CleanShot 2023-04-13 at 23 46 34@2x

</p>
</details> 

As stated in the [Build your own shell plugins (beta)](https://developer.1password.com/docs/cli/shell-plugins/contribute) guide the plugin was validated and built with the appropriate commands. A passing test is provided which tests both the provisioner and the importer.

CC:
@driesvints (Laravel Core team member) as a 1Password user, you might be interested in this 😎

PR #245, opened 2023-04-14
Shell Plugin for Laravel Vapor CLI

andresayej:
Shell Plugin for the Laravel Vapor CLI

The [Laravel Vapor CLI](https://docs.vapor.build/1.0/introduction.html#installing-the-vapor-cli) is the official command line tool for the [Laravel Vapor Service](https://vapor.laravel.com/) offered by Laravel LLC the company behind the popular open source PHP framework: [Laravel](https://laravel.com/) 

The Vapor CLI by default offers authentication via the `login` command using the credentials pair of an Email & Password which in turn makes an API call that fetches the needed API Token that gets stored in the default config location.
However due to the following limitation of 1Password Shell Plugins:
```
Plugins with multiple credential types are not supported yet
```
 
I've made the authentication mechanism for this plugin only via the API Token: `VAPOR_API_TOKEN` for now, based on the source code of the Vapor CLI located here: https://github.com/laravel/vapor-cli

Vapor users can create a Vapor API Token from the [Vapor API settings dashboard](https://vapor.laravel.com/app/account/api-tokens) directly or use the `login` command to authenticate with their Email & Password as usual and let the importer do its work later when the plugin is initialized. Like with other plugins the user should be advised to remove the local config file previously stored on the disk.

Hopefully, in the future, the above-mentioned limitation will be removed (keeping an eye on: https://github.com/1Password/shell-plugins/pull/191) which will allow this plugin to handle multiple authentication mechanisms:
- via an API Token (currently implemented)
- via credentials pair of Email & Password 
This mechanism might need a new dependency ([HTTP client?](https://pkg.go.dev/net/http)) to make an API call that fetches the needed API Token(somehow only once) and should handle 2FA for accounts that have it enabled. The [contributing instructions](https://github.com/1Password/shell-plugins/blob/main/CONTRIBUTING.md#%EF%B8%8F-third-party-dependencies) state that 3rd party dependencies should be avoided so any suggestion or examples on how one can handle this would be helpful.
P.S.: I'm a total beginner to GO 😃

This plugin also offers an importer for scanning the user's environment and file system for the needed credentials in the default config location allowing existing Vapor CLI users to import their current token into 1Password.

Here are some screenshots with the following commands that need authentication:
```
vapor project:list
```
```
vapor cert:list
```
```
vapor zone:list
```
```
vapor record:list <ZONE_ID>
```
<details><summary>Screenshots</summary>
<p>

CleanShot 2023-04-14 at 10 46 23@2x

CleanShot 2023-04-14 at 10 46 43@2x

CleanShot 2023-04-14 at 10 47 17@2x

CleanShot 2023-04-14 at 10 50 12@2x

CleanShot 2023-04-14 at 10 52 03@2x
</p>
</details> 

As stated in the [Build your own shell plugins (beta)](https://developer.1password.com/docs/cli/shell-plugins/contribute) guide the plugin was validated and built with the appropriate commands. A passing test is provided which tests both the provisioner and the importer.

CC (Laravel Team Members):
@driesvints
@timacdonald
@jbrooksuk


joedixon:
Very cool, thank you!

plugins/laravelvapor/vapor.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func LaravelVaporCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Laravel Vapor CLI",
+		Runs:    []string{"vapor"},
```
hculea, line 13: From what I could find in the docs, the calls to this CLI are usually made with 
```
php vendor/bin/vapor <command>
```
i.e. `vapor` is already a shell alias that has to be set by the user, is this correct?

In this situation, would it be worth looking into indicating that the command runs `php vendor/bin/vapor` instead? I think this would be nice, in order to cover all the different setups that users of this CLI might have.

hculea (commented):
Thank you for this contribution! Only a tiny comment 😄

plugins/laravelvapor/vapor.go
andresayej: Unlike the Forge CLI, the Vapor CLI can be installed both globally and on a per-project basis.

The global installation follows the same pattern as described here: https://github.com/1Password/shell-plugins/pull/244#discussion_r1166753444, which means the Vapor CLI is available as `vapor`

My best guess is that the globally installed method is more popular as it offers mostly a one-time setup. (in my opinion)

As for the per-project installation method:
`vapor` as a shell alias is only applicable if the user chooses to make it an alias (it's not mandatory) and refers specifically to the per-project installation method.
But even if the user chooses to make it a shell alias, would that conflict with the `.op/plugins.sh` aliases?
Maybe a note in this plugin's docs could advise the user to make the alias for a consistent experience?

Another approach that crosses my mind would be for the plugin to handle both if there is a way to specify it in the `vapor.go` file or somewhere:
```
Runs: []string{"vapor", "php vendor/bin/vapor "},
```
Not sure if that's possible or even what's the proper syntax to specify more than one item 😄 

Not sure what's the best option here to handle both cases, but I would appreciate any feedback.

andresayej:
@hculea 
> Thank you for this contribution! Only a tiny comment 😄

Answered as best I can and proposed some things.
Thanks again for taking the time to review the PR and would appreciate any feedback

plugins/laravelvapor/vapor.go
hculea: So let me make sure I understand:

`vapor` and `php vendor/bin/vapor` are mutually exclusive ways of calling the binary: the first relates to the global installation, the second, to the project-scoped installation?

hculea: In this case, then, I think we should indeed default to only supporting `vapor`, at this stage, to validate that only instances of global installation can use the plugin.

Shell Plugins currently don't have native support for project-scoped executables, but that is an interesting use-case!

hculea (approved):

plugins/laravelvapor/vapor.go
andresayej: > So let me make sure I understand:
> 
> 
> 
> `vapor` and `php vendor/bin/vapor` are mutually exclusive ways of calling the binary: the first relates to the global installation, the second, to the project-scoped installation?

Absolutely correct 👍

andresayej: > In this case, then, I think we should indeed default to only supporting `vapor`, at this stage, to validate that only instances of global installation can use the plugin.
> 
> 
> 
> Shell Plugins currently don't have native support for project-scoped executables, but that is an interesting use-case!

Sounds good to me 😊 

In the future if project-scoped executables become natively supported I'll probably try to add it as a feature enhancement to this plugin 🤝

SimonBarendse (approved):
Excellent. Welcome to the ecosystem Laravel! ❤️

PR #251, opened 2023-04-21
Support for importing AWS vault credentials from encrypted files

williamhpark:
## Summary

Currently, if you store AWS credentials using AWS vault with an encrypted file as the backend, then try to import credentials into 1Password using the AWS vault importer, you get a long `invalid memory address or nil pointer dereference` error in the terminal and no import candidates are shown. We want to support importing credentials from encrypted files.

Also, while importing using a build of the AWS plugin that **isn't a local build**, several log messages display in the terminal. These messages are invoked in the AWS vault code - we want to silence these. Example:

```
2023/04/21 13:19:32 Loading config file /Users/williampark/.aws/config
2023/04/21 13:19:32 Parsing config file /Users/williampark/.aws/config
2023/04/21 13:19:32 Unrecognised ini file section: DEFAULT
```

## How to Test

1. Add a profile through aws-vault using an encrypted file as the backend: `aws-vault add file-profile --backend file`
2. Run `op plugin init aws` (don't use a local build) > `Import into 1Password...`

## Expected Behaviour

No unnecessary  log messages should be displayed.

You should get a prompt in the terminal: `Enter passphrase to unlock "~/.awsvault/keys/": `. After inputting the correct password, `Encrypted file (file-profile)` should be displayed as an import candidate. You should be able to import the credential into 1Password successfully.

### DIFF
--- a/plugins/aws/importers.go
+++ b/plugins/aws/importers.go
@@ -2,10 +2,11 @@ package aws
 
 import (
 	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
 	"os"
 	"strings"
-
-	"gopkg.in/ini.v1"
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
@@ -14,6 +15,8 @@ import (
 	"github.com/99designs/aws-vault/v7/cli"
 	"github.com/99designs/aws-vault/v7/vault"
 	"github.com/99designs/keyring"
+	"golang.org/x/term"
+	"gopkg.in/ini.v1"
 )
 
 // TryCredentialsFile looks for the access key in the ~/.aws/credentials file.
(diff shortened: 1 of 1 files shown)


plugins/aws/importers.go
```diff
@@ -94,10 +97,27 @@ var backendNames = map[keyring.BackendType]string{
 	keyring.PassBackend:          "Pass",
 }
 
+func fileKeyringPassphrasePrompt(prompt string) (string, error) {
+	if password, ok := os.LookupEnv("AWS_VAULT_FILE_PASSPHRASE"); ok {
+		return password, nil
+	}
+
+	fmt.Fprintf(os.Stderr, "%s: ", prompt)
+	b, err := term.ReadPassword(int(os.Stdin.Fd()))
```
AndyTitu, line 106: I wasn't able to test this because I cannot get encrypted files to work correctly as a backend to my aws-vault.
Screenshot 2023-04-26 at 15 39 49
Is this working as expected for you @williamhpark ?

plugins/aws/importers.go
```diff
@@ -94,10 +97,27 @@ var backendNames = map[keyring.BackendType]string{
 	keyring.PassBackend:          "Pass",
 }
 
+func fileKeyringPassphrasePrompt(prompt string) (string, error) {
+	if password, ok := os.LookupEnv("AWS_VAULT_FILE_PASSPHRASE"); ok {
+		return password, nil
+	}
+
+	fmt.Fprintf(os.Stderr, "%s: ", prompt)
```
AndyTitu, line 105: This doesn't show up when running a local profile through RPC. However, the line underneath for the `Stdin` works as expected. @jpcoenen do you know why that is, and how could we overcome that? This is the only aspect that blocks this PR.

plugins/aws/importers.go
williamhpark: Hey @AndyTitu, the issue is that you need to specify `--backend file` when you run `aws exec` as well (I know kind of weird). So running `aws-vault exec file-profile --backend file -- aws sts get-caller-identity` should work for you. It works for me.

plugins/aws/importers.go
accraw: Why are we printing to Stderr?

jpcoenen: Apologies; I missed this comment originally. I think there is a way to read the output a plugin writes to `stderr`, but I am not sure whether that is the preferred way to go here. 

Might be interesting to see if there are any Terraform Providers that prompt the users. If so, we could see how they tackle that.

Issue #140, opened 2023-01-11
New plugin: Terraform


anthonyangel:
### Platform or tool

Terraform 

### Relevant links

https://developer.hashicorp.com/terraform/cli/commands/login

https://developer.hashicorp.com/terraform/cli/config/config-file#credentials-1

### Goal or desired behavior

Authenticate Terraform against a remote backend (e.g. Terraform Cloud or Scalr)

### Contribution

- [X] I'd be up for building the plugin myself.

### Additional information

Running `terraform login` generates a plaintext JSON credential file. Alternatively you can specify login credentials via env vars

anthonyangel:
Thanks Andy, I'm struggling to get these plugins to play nicely on my Chromebook (The op CLI and Linux app install and run just fine, but it looks like ChromeOS doesn't make the fingerprint reader available to Linux) so I may well not get round to this any time soon

kennedy:
I was browsing through the documentation for CLI configuration. Interally in terraform, there is a credential helper system. I believe we can take advantage of that, and incorporate 1pass to it.

https://developer.hashicorp.com/terraform/cli/config/config-file#credentials-helpers

There is also environment variables we can use as well, but only supported in Terraform v1.2.0 and later.

https://developer.hashicorp.com/terraform/cli/config/config-file#environment-variable-credentials.

For example, my use-case use the Terraform Cloud token: `TF_TOKEN_app_terraform_io`

PR #256, opened 2023-05-01
Add Terraform plugin

AndyTitu:
## Overview

This PR introduces the definition logic for enabling a Terraform shell plugin and similar plugins to be supported. 

It also introduces the Terraform shell plugin, as an example

## Thought process
Credential Usage SDK definition is to be modified to support a SelectFrom CredentialSelection field. This struct defines how a credential selection is to take place in op. This is mutually exclusive with the Name Plugin and Provisioner fields of CredentialUsage.

Additionally the Credential Usage struct will be specified with an Optional field to determine at init step whether the executable can be run without this credential or not.
NeedsAuth will also be added at the level of a credential usage to enable more granular selection of when to provision a credential or not based on the use case.
## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #140 

## How To Test
This can only be tested by internal contributors.

1. `op plugin init terraform` - configure with whatever credentials you want
2. `terraform init` in a new dir
3.  make sure the `main.tf` uses the providers for which you've configured the credentials
4. `terraform plan` - this should work without saying that no credentials were found, even though no credentials are on disk.

## Changelog
Authenticate the Terraform CLI using Touch ID and other unlock options with 1Password Shell Plugins.


sdk/schema/executable.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/1Password/shell-plugins/sdk"
 )
 
+type AllowedCredentials int
+
+const (
+	// DynamicNumber the executable annotated with this value can support a number of credentials which can't be statically determined.
+	DynamicNumber AllowedCredentials = iota
+)
+
```
hculea, line 17: I have a preference to taking this in a separate MR, as it is quite a fundamental change to the plugin schema that shouldn't IMO be hidden under the umbrella of adding a plugin requiring it.

plugins/terraform/terraform.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func TerraformCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Terraform CLI",
+		Runs:    []string{"terraform"},
+		DocsURL: sdk.URL("https://developer.hashicorp.com/terraform/cli"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
```
hculea, line 16: `terraform login` and `terraform logout`?

plugins/terraform/terraform.go
```diff
@@ -0,0 +1,20 @@
...
+		),
+		AllowedCredentials: schema.DynamicNumber,
```
hculea, line 18: This PR should also the specification of the terraform API token, within this plugin.

sdk/schema/executable.go
AndyTitu: Good point. I renamed this PR to reflect that it's mostly about adding some new functionality to the SDK, and Terraform is only a mere example.

plugins/terraform/terraform.go
AndyTitu: Since this PR is mostly about adding support for plugins that support multiple credential types, and Terraform is just an example, I will create separate issues for addressing Terraform specific improvements. Logging in and out of terraform is not absolutely required for using terraform with aws for example, as authentication is done at your AWS account level.

AndyTitu: Good point, will do!

sdk/schema/executable.go
florisvdg: If the only purpose of Terraform in this PR is to act as an example, then we should probably not introduce the actual plugin yet and instead turn it into a "usage snippet" in the PR description.

AndyTitu: It's also handy to couple these changes together for when pulling these dependencies into `op`. What is the main reason why we'd like to separate these? 

Given that these changes also kind of make sense together and their size is not too big, it would make the process easier for me to keep them tied together in the same commit.

sdk/schema/executable.go
```diff
@@ -8,6 +8,15 @@ import (
 	"github.com/1Password/shell-plugins/sdk"
 )
 
+type SupportedCredentialAmount int
+
+const (
+	// ExactlyOne signifies that a certain executable supports exactly one credential
+	ExactlyOne SupportedCredentialAmount = iota
```
AndyTitu, line 15: @florisvdg Does it make sense to introduce `ExactlyOne`? If we make `SupportedCredentialAmount ` an enumeration like we are doing here, for all plugins that don't specify this field, the value taken will be this default value. Should we make this `StaticAmount`, or `FixedAmount` instead? For now, we only have plugins with exactly one credential usage, but in the future this is bound to change. Do you see any future value in having this field be more specific like `ExactlyOne` or `ExactlyTwo`?

sdk/schema/executable.go
florisvdg: That's fine by me, but then we'll also have to review the Terraform plugin as part of this PR. And I'd then call this PR "Add Terraform plugin".

AndyTitu: Done, renamed it back to its original name

plugins/terraform/terraform.go
AndyTitu: https://github.com/1Password/shell-plugins/issues/260

plugins/terraform/terraform.go
AndyTitu: Done!

sdk/schema/plugin.go
```diff
@@ -68,19 +68,13 @@ func (p Plugin) Validate() (bool, ValidationReport) {
...
-		Assertion:   len(p.Credentials) == 1,
-		Severity:    ValidationSeverityError,
-	})
-
-	report.AddCheck(ValidationCheck{
-		Description: "Credentials referenced in executables are included in the same plugin definition",
-		Assertion:   CredentialReferencesInCredentialList(p),
+		Assertion:   len(p.Credentials) <= 1,
 		Severity:    ValidationSeverityError,
 	})
```
jpcoenen, line 79: For my understanding: is there a specific reason this needs to go in this PR? Or is it just a matter of cleaning up an outdated rule?

jpcoenen (commented):
Overall I really like how clean this has ended up becoming 👌 Great job, @AndyTitu. I do have a few remarks, but these are mostly

plugins/terraform/terraform.go
```diff
@@ -0,0 +1,35 @@
...
+				SelectFrom: schema.CredentialSelection{
+					ID:                    "project",
+					IncludeAllCredentials: true,
+					AllowMultiple:         true,
+				},
+				Optional: true,
+				NeedsAuth: needsauth.IfAny(
+					needsauth.ForCommand("refresh"),
+					needsauth.ForCommand("plan"),
+					needsauth.ForCommand("apply"),
```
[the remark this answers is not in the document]
AndyTitu: Added those as well, thanks for checking this!

sdk/schema/executable.go
```diff
@@ -37,6 +37,26 @@ type CredentialUsage struct {
 	// set in the credential schema, so should only be used if this executable requires a custom configuration, that deviates
 	// from the way the credential is usually provisioned.
 	Provisioner sdk.Provisioner
+
+	// (Optional) What this credential will be used for by the executable.
+	Description string
+
+	// (Optional) Instead of requiring a specific credential, have the user select from a list of compatible credentials.
+	// Mutually exclusive with: `Name` and `Plugin`.
+	SelectFrom CredentialSelection
```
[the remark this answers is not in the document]
AndyTitu: Done, good callout, thanks!

sdk/schema/executable.go
```diff
@@ -37,6 +37,26 @@ type CredentialUsage struct {
...
+	NeedsAuth sdk.NeedsAuthentication
+
+	// (Optional) Whether this credential is needed for the executable to run.
+	Optional bool
+}
+
+type CredentialSelection struct {
+	ID                    string
+	IncludeAllCredentials bool
+	AllowMultiple         bool
```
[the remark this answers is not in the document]
AndyTitu: Added comments, thanks!

sdk/schema/executable.go
```diff
@@ -37,6 +37,26 @@ type CredentialUsage struct {
...
+
+	// (Optional) Instead of requiring a specific credential, have the user select from a list of compatible credentials.
+	// Mutually exclusive with: `Name` and `Plugin`.
+	SelectFrom CredentialSelection
+
+	// (Optional) Whether the exectuable needs authentication for this credential. Works side by side with the executable's
+	// `NeedsAuth`, which can still be used for more generic authentications opt-outs, such as the help flag.
+	NeedsAuth sdk.NeedsAuthentication
+
+	// (Optional) Whether this credential is needed for the executable to run.
```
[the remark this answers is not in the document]
AndyTitu: Also added the extra `if true` explanation.

sdk/schema/validation.go
```diff
@@ -122,6 +104,28 @@ func NoDuplicateCredentials(plugin Plugin) bool {
...
+
+func CredentialUsagesSpecifyCredentials(executable Executable) bool {
+	for _, usage := range executable.Uses {
+		if usage.Name == "" && usage.SelectFrom.ID == "" {
+			return false
+		}
+
+		if usage.Name != "" && usage.SelectFrom.ID != "" {
+			return false
+		}
```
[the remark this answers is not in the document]
AndyTitu: Good point, improved this function, thanks!

sdk/schema/plugin.go
AndyTitu: This will be needed in the future but for now we don't need to have it. I removed all validation checks which were not absolutely neccessary for this PR

sdk/schema/validation.go
```diff
@@ -122,6 +104,28 @@ func NoDuplicateCredentials(plugin Plugin) bool {
 	return IsStringSliceASet(ids)
 }
 
+func NoDuplicateCredentialUsages(executable Executable) bool {
+	var usageIds []string
+	for _, credentialUsage := range executable.Uses {
+		usageIds = append(usageIds, credentialUsage.Name.ID().String()+"|"+credentialUsage.Plugin)
```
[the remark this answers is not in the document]
AndyTitu: Good point. I added an ID method for credential usages. Thanks!

sdk/schema/plugin.go
AndyTitu: Actually, after I introduced this check back in, it was still failing. I now made some adjustments to it so that it adheres to the new credential usage definition.

sdk/schema/executable.go
```diff
@@ -75,9 +104,45 @@ func (e Executable) Validate() (bool, ValidationReport) {
...
+
 	return report.IsValid(), report
 }
 
 func (e Executable) Command() string {
 	return strings.Join(e.Runs, " ")
 }
+
+func (c CredentialUsage) GetCredentialReference() *CredentialReference {
+	if c.Name != "" || c.Plugin != "" || c.Provisioner != nil {
```
jpcoenen, line 127: I think we end up with an easier-to-debug situation if we only look at `c.Name` here. We always expect that to be set, right?

AndyTitu: Yes, but we still need a nice way of telling the user "Oh-oh you have specified a `plugin` or a `provisioner` but those will be ignored if you haven't specified a `Name`  or if you otherwise specified a `SelectFrom`". It helps them setting the right expectations on what they are defining.

AndyTitu: In other words, if a user specifies `Provisioner` or `Plugin` there might be intent there, and we should not ignore that intent, but rather catch it and let the user know that something is off.

AndyTitu: I made this function return an error in case a provisioner or plugin is defined without a credential name

sdk/schema/executable.go
```diff
@@ -43,20 +43,29 @@ type CredentialUsage struct {
...
 	NeedsAuth sdk.NeedsAuthentication
 
-	// (Optional) Whether this credential is needed for the executable to run.
+	// Whether this credential is needed for the executable to run. If set to true, the executable cannot run without provisioning this credential.
 	Optional bool
 }
 
 type CredentialSelection struct {
-	ID                    string
+	// ID helps identify credentials chosen in this selection
```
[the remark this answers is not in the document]
AndyTitu: Done, thanks!

sdk/schema/executable.go
```diff
@@ -43,20 +43,29 @@ type CredentialUsage struct {
...
+	ID string
+	// IncludeAllCredentials specifies whether the entire list of credentials should be included in the selection prompt when configuring this credential use.
```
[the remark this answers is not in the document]
AndyTitu: Done, thanks!

sdk/schema/executable.go
```diff
@@ -123,26 +117,61 @@ func (e Executable) Command() string {
...
+		return strings.Join([]string{reference.Plugin, reference.Name.String()}, "|"), nil
+	}
+
+	if c.SelectFrom != nil {
+		return c.SelectFrom.ID, nil
 	}
 
-	return ""
+	return "", fmt.Errorf("credential usage cannot be identified")
 }
```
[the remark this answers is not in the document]
AndyTitu: Okay, removed the errors and only kept them in the `Validation` step.
The `ID` function now returns the empty string if no other id can be obtained from the definition.

sdk/schema/executable.go
```diff
@@ -123,26 +117,61 @@ func (e Executable) Command() string {
...
+	if c.Name != "" {
 		return &CredentialReference{
 			Name:        c.Name,
 			Plugin:      c.Plugin,
 			Provisioner: c.Provisioner,
-		}
+		}, nil
+	} else if c.Plugin != "" || c.Provisioner != nil {
+		return nil, fmt.Errorf("plugin or provisioner specified without a credential name")
 	}
```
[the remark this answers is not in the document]
AndyTitu: Done, as above.

sdk/schema/executable.go
```diff
@@ -21,12 +21,12 @@ type Executable struct {
...
+	// (Optional) Whether the executable needs authentication for certain args.
 	NeedsAuth sdk.NeedsAuthentication
 }
 
 type CredentialUsage struct {
-	// The name of the credential to use in the executable.
+	// (Optional) The name of the credential to use in the executable. Mutually exclusive with `SelectFrom`.
 	Name sdk.CredentialName
 
 	// (Optional) The plugin name that contains the credential. Defaults to the current package. This can be used to
```
accraw, line 32: Does "Mutually exclusive with `SelectFrom`" need to be added here too?

sdk/schema/executable.go
```diff
@@ -75,9 +98,57 @@ func (e Executable) Validate() (bool, ValidationReport) {
 		Severity:    ValidationSeverityError,
 	})
 
+	report.AddCheck(ValidationCheck{
+		Description: "Credential Usages are uniquely identifiable inside an executable",
+		Assertion:   CredentialUsagesUniquelyIdentifiable(e),
```
florisvdg, line 103: How about:
```suggestion
		Description: "Credential usage definitions uniquely identifiable inside an executable",
		Assertion:   IsCredentialUsageUniquelyIdentifiable(e),
```

sdk/schema/executable.go
```diff
@@ -37,6 +37,29 @@ type CredentialUsage struct {
...
+	NeedsAuth sdk.NeedsAuthentication
+
+	// Whether this credential is needed for the executable to run. If set to true, the executable cannot run without provisioning this credential.
+	Optional bool
+}
+
+type CredentialSelection struct {
+	// ID helps identify credentials chosen in this selection. This must be unique in relation to other selections specified in usages within its executable.
+	ID string
+	// IncludeAllCredentials specifies whether all credentials defined in all plugins should be included in the selection prompt when configuring this credential use.
```
florisvdg, line 59: We can shorten this, given the context that's already there:
```suggestion
	// IncludeAllCredentials specifies whether this selection should contain all credentials defined in all plugins.
```

plugins/terraform/terraform.go
```diff
@@ -19,7 +19,7 @@ func TerraformCLI() schema.Executable {
 			{
 				Description: "Credentials to use within the Terraform project",
 				SelectFrom: &schema.CredentialSelection{
-					ID:                    "project",
+					ID:                    "Project Credentials",
```
florisvdg, line 22: What's the reason for this change? The previous version looks more like an ID to me.

florisvdg (approved):

sdk/schema/executable.go
```diff
@@ -119,79 +113,42 @@ func (e Executable) Command() string {
...
+		Description: "If defined, a credential selection must have its `ID` and `IncludeAllCredentials` set",
+		Assertion:   c.SelectFrom == nil || (c.SelectFrom.ID != "" && c.SelectFrom.IncludeAllCredentials),
 		Severity:    ValidationSeverityError,
 	})
 
-	id, err := c.ID()
 	report.AddCheck(ValidationCheck{
 		Description: "Credential usage has either a credential reference or selection properly defined, but not both",
-		Assertion:   err == nil && !(credRef != nil && selection != nil),
+		Assertion:   c.ID() != "" && !(c.SelectFrom != nil && c.Name != ""),
```
jpcoenen, line 133: Can we write this check without using `c.ID()`? That function depends on `Validate()`. By also using it here, we have a circular dependency: `ID()`'s output is valid if `Validate()` succeeds, which success depends on the output of `ID()`.

plugins/terraform/terraform.go
AndyTitu: On `op` this ID is now used everywhere where the `Name` was used in the plugin commands, for messages like `Authenticate with __`. 
Screenshot 2023-05-19 at 13 21 13

AndyTitu: e.g. `Authenticate with Terraform Project Credentials` sounds better.

sdk/schema/executable.go
AndyTitu: Yes, makes sense! Done, thanks!

sdk/schema/executable.go
AndyTitu: Done, thanks!

sdk/schema/executable.go
AndyTitu: Done, thanks!

sdk/schema/executable.go
AndyTitu: Done, thanks!

plugins/terraform/terraform.go
```diff
@@ -0,0 +1,38 @@
...
+		Name:    "Terraform CLI",
+		Runs:    []string{"terraform"},
+		DocsURL: sdk.URL("https://developer.hashicorp.com/terraform/cli"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+		),
+		Uses: []schema.CredentialUsage{
+			{
+				Description: "Credentials to use within the Terraform project",
```
accraw, line 20:
```suggestion
				Description: "Credentials to use within terraform projects",
```

plugins/terraform/terraform.go
AndyTitu: Yep, makes sense. Changed back to initial version.

plugins/terraform/terraform.go
AndyTitu: Done, thanks!

florisvdg: Hmm, I think the previous version makes more sense, to be honest. This configuration is highly dependent on the Terraform project, so the more common use case would be to set up a different profile per project.

AndyTitu: Okay, changed this back to normal. We could revisit this if better wording comes to mind later.

jpcoenen (approved):
🚀

pattobrien:
@jpcoenen any reason the terraform plugin was not included with the official v2.19.0 release last week?

pattobrien:
@AndyTitu got it! Thanks for the info - any chance the feedback is publicly discussed? would be more than happy to contribute, just point me to where I can do so. 👍🏼

sfnelson:
We're using 1password to give terraform access to Github tokens. It's great, but it's getting painful to keep using the old beta build that includes access and we're worried it's going to break as 1password moves forwards. I know it's not perfect, but could we please get this plugin released so we're not relying on an old build?

SeanHood:
I've been using the `terraform` plugin in `op` for a few months now, but I'm now getting these errors:

```
$ op -v
2.23.0-beta.01

$ op whoami
[ERROR] 2024/04/24 23:48:28 connecting to desktop app: 1Password CLI couldn't connect to the 1Password desktop app. To fix this, update the desktop app to the latest version. If you're still having trouble connecting, restart the app. If you're using version 8.10.12 or earlier of the app, the 1Password CLI binary must be located in /usr/local/bin/op.
```

After moving `op` from `~/.local/bin` to `/usr/local/bin` I get:

```
$ op whoami
[ERROR] 2024/04/24 23:52:07 connecting to desktop app: connection reset, make sure 1Password CLI is installed correctly, then open the 1Password app, select 1Password > Settings > Developer and make sure the 'Integrate with 1Password CLI' setting is turned on. If you're still having trouble connecting, restart the app.
```

I don't get any of these errors when trying `v2.27.0` of `op` however this lacks the Terraform plugin.

Is anyone seeing similar on this old build?

sfnelson:
Yes, we've had the same issue. Since the latest 1Password release (April) we can't use the old `op` beta build anymore. I suspect there's a binary incompatibility or a deliberate version delta limit between the old beta CLI build and the new 1Password application.

As a workaround I've installed the current `op` build and compiled my own version of the terraform plugin, i.e.

```
git clone git@github.com:1Password/shell-plugins.git
cd shell-plugins
make terraform/build
op plugin init terraform
```

This isn't great as I get a giant warning every time I run terraform, and I'm going to have to keep re-building every time there's a release, but at least I can run my terraform builds.

rterbush:
I've run into this issue as well. I'm now on op version 2.28.0. Previous version seemed to still have support for the terraform plugin.

Issue #257, opened 2023-05-02
Add validation check that ensures that each plugin name is unique in the registry


AndyTitu:
### What does it concern?

The plugin SDK, design, or schema

### Goal or desired behavior

Currently, each plugin in our registry can be uniquely identified by its name. We rely on this invariant for multiple internal workflows, such as the plugin cache indexing. We should ensure this invariant remains true in the future by adding a check that ensures no duplicate plugin name is added to the registry.

### Current behavior

no such check is present.

PR #258, opened 2023-05-03
Add IDs to credentials based on their name

AndyTitu:
## Overview

We need a way to identify credentials in a stable long-term manner. The credential names could be subject to change at a certain point in the future, so we are introducing an additional layer meant for unique identification that is not vulnerable to sdk changes.

Also we need to enforce that plugins are uniquely identified by their name, and credentials inside each plugin definition are uniquely identifiable by their ID.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #257 

Relates other some internal issues.

## Changelog
  

Credentials now have an ID they can be identified after.

Plugins can be uniquely identified by their name, and credentials inside a plugin definition can be uniquely identified by their ID

### DIFF
--- a/plugins/registry_test.go
+++ b/plugins/registry_test.go
@@ -1,10 +1,10 @@
 package plugins
 
 import (
-	"github.com/stretchr/testify/assert"
 	"testing"
 
 	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestValidatePlugins(t *testing.T) {
--- a/plugins/registry_test.go
+++ b/plugins/registry_test.go
@@ -1,6 +1,7 @@
 package plugins
 
 import (
+	"github.com/stretchr/testify/assert"
 	"testing"
 
 	"github.com/1Password/shell-plugins/sdk/schema"
@@ -16,4 +17,13 @@ func TestValidatePlugins(t *testing.T) {
 			}
 		}
 	}
+}
+
+func TestAllPluginsHaveUniqueNames(t *testing.T) {
+	var pluginNames []string
+	for _, p := range registry {
+		pluginNames = append(pluginNames, p.Name)
+	}
+
+	assert.True(t, schema.IsStringSliceASet(pluginNames))
 }
--- a/sdk/names.go
+++ b/sdk/names.go
@@ -25,6 +25,8 @@ func (n CredentialName) ID() CredentialTypeID {
 func credentialNameToSnakeCase(name CredentialName) string {
 	str := name.String()
 	str = strings.ReplaceAll(str, " ", "_")
+	str = strings.ReplaceAll(str, "-", "_")
+	str = strings.ReplaceAll(str, "/", "_")
 	return strings.ToLower(str)
 }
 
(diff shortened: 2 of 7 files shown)


sdk/schema/credname/names_test.go
```diff
@@ -0,0 +1,54 @@
...
+		"cli_token",
+		"credential",
+		"credentials",
+		"database_credentials",
+		"login_details",
+		"personal_api_token",
+		"personal_access_token",
+		"registry_credentials",
+		"secret_key",
+	}
```
florisvdg, line 49: Are we expecting contributors to keep this list up to date? If not, I'd recommend making this slightly less realistic-looking and more like a unit test that is also testing more edge cases.

sdk/schema/validation.go
```diff
@@ -112,3 +112,27 @@ func CredentialReferencesInCredentialList(plugin Plugin) bool {
 	}
 	return true
 }
+
+func CredentialsAreUniquelyIdentifiableInsideAPlugin(plugin Plugin) bool {
```
florisvdg, line 116: How about something like:
```suggestion
func PluginHasNoDuplicateCredentials(plugin Plugin) bool {
```
Or:
```suggestion
func NoDuplicateCredentials(plugin Plugin) bool {
```

AndyTitu: Done, thanks!

sdk/schema/credname/names_test.go
AndyTitu: Added some additional edge case in addition to the list of the copy pasted existing credential names.

sdk/names.go
```diff
@@ -15,3 +17,19 @@ type CredentialName string
...
+
+func (n CredentialName) ID() CredentialTypeID {
+	return CredentialTypeID(credentialNameToSnakeCase(n))
+}
+
+func credentialNameToSnakeCase(name CredentialName) string {
+	str := name.String()
+	str = strings.ReplaceAll(str, " ", "_")
+	return strings.ToLower(str)
+}
```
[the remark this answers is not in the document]
AndyTitu: Done, thanks!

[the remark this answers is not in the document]
AndyTitu: Should we expect commas (,) as well?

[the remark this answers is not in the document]
florisvdg: `,` and `.` could make sense to add as well 👍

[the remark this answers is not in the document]
AndyTitu: Done, and added more tests. Thanks

jpcoenen (approved):

florisvdg (approved):

plugins/registry_test.go
```diff
@@ -17,3 +18,12 @@ func TestValidatePlugins(t *testing.T) {
...
 }
+
+func TestAllPluginsHaveUniqueNames(t *testing.T) {
+	var pluginNames []string
+	for _, p := range registry {
+		pluginNames = append(pluginNames, p.Name)
+	}
+
+	assert.True(t, schema.IsStringSliceASet(pluginNames))
+}
```
jpcoenen, line 29: We could add a similar test that also makes sure the credential ID's used within a plugin are unique.

hculea (approved):

PR #259, opened 2023-05-08
Don't write a new aws config file when importing

AndyTitu:
## Overview

It seems the aws-vault importer is creating the `.aws/config` file if it doesn't exist. We should make sure that importers only read.

## Type of change
 

- [ ] Created a new plugin

## How To Test

`op plugin init aws` and import credentials

## Changelog

AWS Shell Plugin importer never creates a `.aws/config` file if one does not exist already.

### DIFF
--- a/plugins/aws/importers.go
+++ b/plugins/aws/importers.go
@@ -133,24 +133,16 @@ func TryAWSVaultBackends() sdk.Importer {
 			// Use the CredentialKeyring struct from aws-vault to retrieve vaulting backend credentials
 			credentialKeyring := &vault.CredentialKeyring{Keyring: keyring}
 
-			// Load the AWS config file (default location at ~/.aws/config)
-			awsConfigFile, err := GetAWSConfigFile()
+			profilesInfo, err := GetProfilesInfo()
 			if err != nil {
 				attempt.AddError(err)
 				return
 			}
-			configLoader := &vault.ConfigLoader{File: awsConfigFile}
-
-			// Get the region specified for the "default" profile
-			var defaultRegion string
-			if defaultSection, ok := awsConfigFile.ProfileSection("default"); ok {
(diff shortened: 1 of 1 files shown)


accraw:
Checked that it fixes the bug, success! A little unsure as to how profiles are supposed to work here (I was unsuccessful with my test) and would love to chat Monday about that.

arunsathiya (approved):
The overall changes here look good to me -- both code-wise, and functional testing.

arunsathiya:
One question I have outside of the review is: if a `~/.aws/config` file does not exist, the aws-vault functionality (without shell plugin in its way) prompts only for access key ID and secret key, but the two of them alone cannot be used **without the region**. 

I believe that's expected behavior with aws-vault, but the larger question I have on my mind is, we completely skip the aws-vault importer when `~/.aws/config` does not exist -- is that expected?

Do you think we should explore a flow that imports credential from aws-vault, if the credential exists, along with region sourced from AWS_REGION envvar, again if it exists?

AndyTitu:
> Do you think we should explore a flow that imports credential from aws-vault, if the credential exists, along with region sourced from AWS_REGION envvar, again if it exists?

Good question! We are already checking for the `AWS_DEFAULT_REGION` envvar to import an existing region for the disk importer. Since importers are all called together, there is no need to duplicate checking for a region env var inside the aws-vault importer as well. 
I agree we could also check the `config` file for getting a region. We should only check the `default` profile though, as the information stored in 1Password acts as a `default` profile -- it is globally used in all other profiles. Each individual profile can then specify its own region inside its section in the `config`.
I have opened a starter issue for adding this logic in the importer: https://github.com/1Password/shell-plugins/issues/302

> access key ID and secret key, but the two of them alone cannot be used without the region.

They can be used without a region except for when using MFA, in which case we error out if we can't find a region.

Issue #164, opened 2023-01-25
Missing field name in the generated template when last word of the Credential Name is greater than 7 characters


arunsathiya:
### op CLI version

2.12.0

### Goal or desired behavior

Credential template file is generated correctly.

### Current behavior

Credential template file generated is **missing the field name** (not **credential name**) when the credential name is entered to be `Credentials` in the setup flow:

<img width="914" alt="image" src="https://user-images.githubusercontent.com/18581859/214452368-0c5f8e46-6064-42ea-85fd-9ef121c5365d.png">

```go
func Credentials() schema.CredentialType {
	return schema.CredentialType{
		Name:          credname.Credentials,
		DocsURL:       sdk.URL("https://ngrok.com/docs/credentials"), // TODO: Replace with actual URL
		ManagementURL: sdk.URL("https://console.ngrok.com/user/security/tokens"), // TODO: Replace with actual URL
		Fields: []schema.CredentialField{
			{
				Name:                fieldname.,
				MarkdownDescription: " used to authenticate to ngrok.",
				Secret:              true,
				Composition: &schema.ValueComposition{
					Length: 43,
					Charset: schema.Charset{
						Uppercase: true,
						Lowercase: true,
						Digits:    true,
					},
				},
			},
		},
		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
		Importer: importer.TryAll(
			importer.TryEnvVarPair(defaultEnvVarMapping),
			TryngrokConfigFile(),
		)}
}
```

Notice the `fieldname.` part in the above code.

arunsathiya:
I am happy to take a closer look at why this is happening only for `Credentials`. Will followup as soon as I do.

arunsathiya:
I was seeing incorrectly. It's not just `Credentials` that's affected. Anything Credential Name with the last word greater than seven characters fails to generate the Field Name. The seven character limit comes from this part:

https://github.com/1Password/shell-plugins/blob/5dae3fe19bf9640979198584f8cd17f67de4c408/cmd/contrib/main.go#L183-L199

I think a fallback Field Name should be decided, but I am not sure what that word should be. Since `Token` seems used the most, should we use that? Or just set the Field Name to be the same as the Credential Name, and show a Todo that the developer should change the Field Name?

Tagging @SimonBarendse and @hculea for their insights, and anyone else that has ideas are absolutely welcome to share!

PR #263, opened 2023-05-22
fix(new-plugin): Set only the last word of the credential name as the field name, if it's larger than seven characters

arunsathiya:
## Overview

Prior to this change, when the credential name's last word is larger than seven characters, the field name generated by the "make new-plugin" template is empty. 

With this change, the last word of the credential name is set as the field name. In the example below, field name `Credentials` is generated for the credential name `GitHub Access Credentials`.

image

If the last word is smaller than seven characters, the usual pattern of stitching together the last several words of the credential name (proposed originally on this PR https://github.com/1Password/shell-plugins/pull/72) will be retained. In the example below, field name `API Key` is generated for the credential name `GitHub API Key`.

image

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: https://github.com/1Password/shell-plugins/issues/164

## How To Test

- Clone this branch and switch to it.
- Run `make new-plugin`
- When prompted to enter a credential name, try two variants: one with the last word less than seven characters, and one with the last word having greater than seven characters. See that both work as expected, especially the latter variant should have only the last word as the field name.

## Changelog
  

Set only the last word of the credential name as the field name, if it's larger than seven characters

### DIFF
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -191,24 +191,25 @@ func newPlugin() error {
 
 	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting from the last word.
 	//
+	// When the last word of the credential name is greater than seven characters, the last word is used as the field name.
+	//
 	// For example:
 	// "Personal Access Token" => "Token"
 	// "Secret Key" => "Key"
 	// "API Key" => "API Key"
-	//
-	// When the last word of the credential name is greater than seven characters, the last word is used as the field name
-	var fieldNameSplit []string
+	// "GitHub API Key" => "API Key"
+	// "Credentials" => "Credentials"
 	lengthCutoff := 7
-	for i := range credNameSplit {
-		word := credNameSplit[len(credNameSplit)-1-i]
+	fieldNameSplit := []string{credNameSplit[len(credNameSplit)-1]}
+	for i := range credNameSplit[:len(credNameSplit)-1] {
+		// Skip credNameSplit's last word because it's already added to fieldNameSplit
 		if i == 0 {
-			fieldNameSplit = append([]string{word}, fieldNameSplit...)
 			continue
 		}
+		word := credNameSplit[len(credNameSplit)-1-i]
 		if len(strings.Join(append(fieldNameSplit, word), " ")) > lengthCutoff {
 			break
 		}
-
 		fieldNameSplit = append([]string{word}, fieldNameSplit...)
 	}
 	result.FieldName = strings.Join(fieldNameSplit, " ")
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -189,15 +189,22 @@ func newPlugin() error {
 		}
 	}
 
-	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting
-	// from the last word. For example:
+	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting from the last word.
+	//
+	// For example:
 	// "Personal Access Token" => "Token"
 	// "Secret Key" => "Key"
 	// "API Key" => "API Key"
+	//
+	// When the last word of the credential name is greater than seven characters, the last word is used as the field name
 	var fieldNameSplit []string
 	lengthCutoff := 7
 	for i := range credNameSplit {
 		word := credNameSplit[len(credNameSplit)-1-i]
+		if i == 0 {
+			fieldNameSplit = append([]string{word}, fieldNameSplit...)
+			continue
+		}
 		if len(strings.Join(append(fieldNameSplit, word), " ")) > lengthCutoff {
 			break
 		}
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -200,18 +200,7 @@ func newPlugin() error {
 	// "GitHub API Key" => "API Key"
 	// "Credentials" => "Credentials"
 	lengthCutoff := 7
-	fieldNameSplit := []string{credNameSplit[len(credNameSplit)-1]}
-	for i := range credNameSplit[:len(credNameSplit)-1] {
-		// Skip credNameSplit's last word because it's already added to fieldNameSplit
-		if i == 0 {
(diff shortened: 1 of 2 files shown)


cmd/contrib/main.go
```diff
@@ -180,15 +180,22 @@ func newPlugin() error {
...
-	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting
-	// from the last word. For example:
+	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting from the last word.
+	//
+	// For example:
 	// "Personal Access Token" => "Token"
 	// "Secret Key" => "Key"
 	// "API Key" => "API Key"
+	//
+	// When the last word of the credential name is greater than seven characters, the last word is used as the field name
```
AndyTitu, line 190: super nit: we could move this comment line above the examples so we could also introduce an example `"Credentials" => "Credentials"`

cmd/contrib/main.go
```diff
@@ -180,15 +180,22 @@ func newPlugin() error {
...
 	var fieldNameSplit []string
 	lengthCutoff := 7
 	for i := range credNameSplit {
 		word := credNameSplit[len(credNameSplit)-1-i]
+		if i == 0 {
```
AndyTitu, line 195: mega nit
```suggestion
                // always add the last word from the credential name
		if i == 0 {
```

AndyTitu (approved):
Thanks for this very well documented improvement! Lgtm

cmd/contrib/main.go
hculea, line 195: small nit, but I would prefer something like:
```suggestion
        fieldNameSplit := []string{credNameSplit[len(credNameSplit) - 1]}
        if len(credNameSplit) > 1 {
	  for i := range credNameSplit[:len(credNameSplit) - 1] {
		  word := credNameSplit[len(credNameSplit)-1-i]
		  if len(strings.Join(append(fieldNameSplit, word), " ")) > lengthCutoff {
			  break
		  }
	  }
	}
```
or something along these lines. given that we want the last element _always_ in the field name, I find it clearer to already init the slice with this token in it.

arunsathiya: Thanks for the tip, @hculea! This has a tiny problem that "GitHub API Key" will result only in "Key" while we have space for "API Key" (cutoff is 7 characters). So, I have taken a slightly updated approach in https://github.com/1Password/shell-plugins/pull/263/commits/cb4a91f5d0190b3634057608a53e24ada513a63f. This should satisfy all of these cases:

- "Personal Access Token" => "Token"
- "Secret Key" => "Key"
- "API Key" => "API Key"
- "GitHub API Key" => "API Key"
- "Credentials" => "Credentials"

arunsathiya:
@AndyTitu and @hculea, this is ready for another look now. Thank you!

cmd/contrib/main.go
```diff
@@ -180,19 +180,27 @@ func newPlugin() error {
...
-	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting
-	// from the last word. For example:
+	// As a placeholder, assume the field name is the short version (max 7 chars) of the credential name, starting from the last word.
+	//
+	// When the last word of the credential name is greater than seven characters, the last word is used as the field name.
+	//
+	// For example:
 	// "Personal Access Token" => "Token"
 	// "Secret Key" => "Key"
 	// "API Key" => "API Key"
```
jpcoenen, line 190: I am not sure that this case is still covered. 

To make sure that we are covering all cases, could we extract this functionality into a separate function and write a quick table-test for it? That would make it easy to see which cases are covered and which are not.

cmd/contrib/main.go
```diff
@@ -180,19 +180,27 @@ func newPlugin() error {
...
 	lengthCutoff := 7
-	for i := range credNameSplit {
+	fieldNameSplit := []string{credNameSplit[len(credNameSplit)-1]}
+	for i := range credNameSplit[:len(credNameSplit)-1] {
+		// Skip credNameSplit's last word because it's already added to fieldNameSplit
+		if i == 0 {
+			continue
+		}
 		word := credNameSplit[len(credNameSplit)-1-i]
 		if len(strings.Join(append(fieldNameSplit, word), " ")) > lengthCutoff {
```
jpcoenen, line 201: A suggestion to address that the `API Key` case is no longer covered: we could also use the original code and add an extra case to this condition:

```suggestion
		if len(strings.Join(append(fieldNameSplit, word), " ")) > lengthCutoff && len(fieldNameSplit) != 0 {
```

jpcoenen (changes requested):
Thank you for improving the contribution framework! Improvements like these make our work so much easier in the long-term 😀

cmd/contrib/main.go
arunsathiya: Sorry to surface this PR after a really long time! Just had some time to look at this.

You were right, @jpcoenen. `API Key` case wasn't covered. I have made some changes in https://github.com/1Password/shell-plugins/pull/263/commits/aecdd2dd526e04e5ed33ba0a29d810b0fe8ef78f that should cover this case, and it includes a test to ensure that all cases are successful too. Would like your feedback!

cmd/contrib/main.go
```diff
@@ -191,18 +191,7 @@ func newPlugin() error {
 	// "GitHub API Key" => "API Key"
 	// "Credentials" => "Credentials"
```
jpcoenen, line 192: This comment seems mostly duplicated from the godoc of the function. Do we need both?

jpcoenen (approved):
One tiny remark, but pretty much good to go 🚀

cmd/contrib/main.go
arunsathiya: I propose we maintain both examples (especially the second one) because we have a check in place to use the last word always, irrespective of whether that word is greater than seven characters or not.

arunsathiya:
Rebased with main to fix conflicts.

SimonBarendse (approved):

Issue #264, opened 2023-05-22
Add a validation check for usages's `Plugin` NOT to be specified if the credential is in the same plugin


AndyTitu:
### What does it concern?

The plugin SDK, design, or schema

### Goal or desired behavior

We wish to achieve as much explicit behaviour with our definitions. Currently a user can leave the `Plugin` field empty if the credential specified in the executable's credential usage references a credential from the plugin definition. We would like this pattern to not be breached.
<img width="628" alt="Screenshot 2023-05-22 at 16 33 22" src="https://github.com/1Password/shell-plugins/assets/45081667/24d12df6-ee5e-47a6-a24b-21f5dd93936f">

### Current behavior

BUT a user could just as well also specify a `Plugin` field in this case (with the name "aws" in the case shown above). This is an option that we would prefer a contributor not to use. 

To make this explicit, we should add a validation check that makes sure that the `Plugin` field is never set in cases where the plugin referenced in the executable's credential usage is any of the ones specified in the same plugin's definition, a.k.a. in the example above, don't allow the `Plugin` to be specified if its name is `aws`.

This validation check should happen in `sdk/schema/plugin.go`.

arunsathiya:
I am happy to take this on. 👍🏼

PR #265, opened 2023-05-23
Warn when plugin name set in credential usage is the same as where the credential is from

arunsathiya:
## Overview

When the plugin name defined in the credential usage block is the same as where the credential is from, that optional field needn't be defined. That field is designed to automatically default to the current plugin/package, so that field can be completely ignored in such cases.

The validation check proposed in this PR checks if it's from the same plugin and warns (does not appear as an error, just shows a warning) the contributor to remove it, when `make plugin-name/validate` command is run.

Example from the AWS plugin, both for AWS executable and CDK executables:

image

image

image

Without that plugin field defined:

image

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: https://github.com/1Password/shell-plugins/issues/264

## How To Test

- Clone this branch and switch to it.
- Edit AWS shell plugin's executables (either AWS, CDK or both) to add `Plugin: "aws",` to the CredentialUsage block.
- Run `make aws/validate` to see the warning.

## Changelog
  

Warn when plugin name set in credential usage is the same as where the credential is from

### DIFF
--- a/sdk/schema/plugin.go
+++ b/sdk/schema/plugin.go
@@ -90,6 +90,12 @@ func (p Plugin) Validate() (bool, ValidationReport) {
 		Severity:    ValidationSeverityError,
 	})
 
+	report.AddCheck(ValidationCheck{
+		Description: "Do not specify plugin name in credential usage because credential is from the same plugin",
+		Assertion:   !UnnecessaryPluginNameDefined(p),
+		Severity:    ValidationSeverityWarning,
+	})
+
 	return report.IsValid(), report
 }
 
--- a/sdk/schema/validation.go
+++ b/sdk/schema/validation.go
@@ -136,3 +136,14 @@ func IsStringSliceASet(slice []string) bool {
 
 	return true
 }
+
+func UnnecessaryPluginNameDefined(plugin Plugin) bool {
+	for _, executable := range plugin.Executables {
+		for _, execCredential := range executable.Uses {
+			if execCredential.Plugin == plugin.Name {
+				return true
+			}
+		}
+	}
+	return false
+}


arunsathiya:
@AndyTitu, I see at least two more improvements I could make to this flow. I'd like your thoughts/approval on these before I start working on them:

- https://github.com/1Password/shell-plugins/issues/266
- https://github.com/1Password/shell-plugins/issues/267

sdk/schema/validation.go
```diff
@@ -136,3 +136,14 @@ func IsStringSliceASet(slice []string) bool {
 
 	return true
 }
+
+func UnnecessaryPluginNameDefined(plugin Plugin) bool {
```
hculea, line 140: nit: let's use a more suggestive function name
```suggestion
func RedundantPluginNameDefined(plugin Plugin) bool {
```

sdk/schema/plugin.go
```diff
@@ -90,6 +90,12 @@ func (p Plugin) Validate() (bool, ValidationReport) {
 		Severity:    ValidationSeverityError,
 	})
 
+	report.AddCheck(ValidationCheck{
+		Description: "Do not specify plugin name in credential usage because credential is from the same plugin",
+		Assertion:   !UnnecessaryPluginNameDefined(p),
+		Severity:    ValidationSeverityWarning,
+	})
+
```
hculea, line 98: I am not really a fan of this, to be honest.

It imposes an artificial restriction on the contributors which does not really improve the code quality/readability (I would even argue that in the case of a common credential name - e.g. Personal Access Token, adding the plugin name too even improves readability - but this is a personal opinion).

I think there are two other problems related to the same thing that are more important to address: 
* the " Credentials referenced in executables are included in the same plugin definition" check, which would fail if any cross-referencing occurs
* enforcing that the "Plugin" value of a "CredentialUsage" actually refers to a plugin. Right now, any value could be in there and validation doesn't trigger anything.

Would love to hear your thoughts @AndyTitu @arunsathiya

hculea: I already see you captured my second concern in an issue @arunsathiya 😄 https://github.com/1Password/shell-plugins/issues/267

Thanks for the proactive action here!

AndyTitu: I see this comment is more related to the purpose of the issue we are trying to address itself than the actual implementation. 

This check originated from our Terraform changes in `op`. We are using a credential usage id now to identify credential usages inside an executable. This usage id is stored inside the json file. For old json configurations that don't have a usage id stored in them we need to come up with a default value ourselves, for backwards compatibility and we chose to use the `Name` of the credential usage (because that's what all plugins up to this point have defined). However, if someone were to add a `Plugin` to say the AWS credential usage, that rule would break. So we came up with the enforcement: if a credential usage references the credential in the same plugin, you can't specify the `Plugin`. Otherwise, you must specify the plugin. This explicit separation should help avoid issues/confusion in the future.

AndyTitu: > the " Credentials referenced in executables are included in the same plugin definition" check, which would fail if any cross-referencing occurs

I do agree we will run into this soon. But for the scope of this PR, there is no immediate need to remove that checks, as no definitions explicitly refer to other plugins.

hculea: > I see this comment is more related to the purpose of the issue we are trying to address itself than the actual implementation.

Indeed, I preferred leaving a comment here to block the merging of this PR before we address my concern, rather than leave it on the ticket where it could easily be missed. My apologies if this caused any confusion 😄 

> I added it now.

Thank you!

Despite your explanation, I still disagree this is the way to proceed. Such changes should not have collateral impact affecting the user (in this case, contributor) experience. Rather than making assumptions about plugin definitions only enforced by a validation check, we could simply use both keys to look up credentials:
* for credentials usages within the plugin where the credential itself is defined, check both for <credential name> and <plugin name + credential name> (if the `Plugin` field is set).
* for credentials usages within plugins different than the one where the credential is defined, only check for <plugin name + credential name>, enforcing that the plugin name is defined.

Now, I don't feel particularly strong about this, so in the interest of moving forward, I'm okay with using this approach, now that I have the context behind it. However, I would still recommend that we take on the effort of ensuring this is backwards compatible ourselves, without any visible side-effects.

PR #271, opened 2023-05-29
gh: do not require auth if running for shell completion

j178:
## Overview

Do not require auth if running `gh __complete` for shell completion. 

## Type of change

- [ ] Created a new plugin


florisvdg (approved):
Thanks for the contribution!

accraw (approved):

PR #273, opened 2023-05-29
feat: add vercel plugin

j178:
## Overview

Add `vercel` plugin.

## Type of change

- [x] Created a new plugin

## How To Test

```
op plugin init vercel
op plugin run -- vercel whoami
```

### DIFF
--- a/plugins/vercel/api_token.go
+++ b/plugins/vercel/api_token.go
@@ -1,0 +1,82 @@
+package vercel
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIToken,
+		DocsURL:       sdk.URL("https://vercel.com/docs/rest-api#authentication"),
+		ManagementURL: sdk.URL("https://vercel.com/account/tokens"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.Token,
+				MarkdownDescription: "Token used to authenticate to Vercel.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 24,
+					Charset: schema.Charset{
+						Lowercase: true,
+						Uppercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: vercelProvisioner{},
+		Importer: importer.TryAll(
+			importer.MacOnly(TryVercelConfigFile("~/Library/Application Support/com.vercel.cli/auth.json")),
+			importer.LinuxOnly(TryVercelConfigFile("~/.config/com.vercel.cli/auth.json")),
+		),
+	}
+}
+
+type vercelProvisioner struct{}
+
+func (v vercelProvisioner) Description() string {
+	return "Vercel cli token provisioner"
+}
+
+func (v vercelProvisioner) Provision(ctx context.Context, input sdk.ProvisionInput, output *sdk.ProvisionOutput) {
+	output.AddArgs("--token", input.ItemFields[fieldname.Token])
+}
+
+func (v vercelProvisioner) Deprovision(ctx context.Context, input sdk.DeprovisionInput, output *sdk.DeprovisionOutput) {
+	// No-op
+}
+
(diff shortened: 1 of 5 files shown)


plugins/vercel/api_token.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+type vercelProvisioner struct{}
+
+func (v vercelProvisioner) Description() string {
+	return "Vercel cli token provisioner"
+}
+
+func (v vercelProvisioner) Provision(ctx context.Context, input sdk.ProvisionInput, output *sdk.ProvisionOutput) {
```
hculea, line 52: Is there a reason you've opted for a file-based provisioner, instead of an env-var provisioner? 

Generally, I suggest using an env-var provisioner where possible, due to current limitations of the file-based approach when it comes to multiple reads within the same command.

plugins/vercel/api_token_test.go
```diff
@@ -0,0 +1,56 @@
...
+				ExpectedCandidates: []sdk.ImportCandidate{
+					{
+						Fields: map[sdk.FieldName]string{
+							fieldname.Token: "tZk79pLyPLGgUVlkHbnLeXgl",
+						},
+					},
+				},
+			},
+			"config file": {
+				OS: "darwin",
```
hculea, line 42: Let's also add a case for the Linux importer, to cover all the possible options here. 😄

plugins/vercel/vercel.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func VercelCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Vercel CLI",
+		Runs:    []string{"vercel"},
+		DocsURL: sdk.URL("https://vercel.com/docs/cli"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
```
hculea, line 17: should `vercel login` be part of these commands?

plugins/vercel/api_token.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIToken,
+		DocsURL:       sdk.URL("https://vercel.com/docs/cli/login"),
```
hculea, line 16: I think this page has no mention of programmatic authentication (using the API token). Do you have any knowledge of a different documentation page about that?

hculea (commented):
Thank you for your contribution! ❤️  I've only got a few questions and couple suggestions.

plugins/vercel/api_token.go
j178: How about this https://vercel.com/docs/rest-api ?

plugins/vercel/api_token.go
j178: Actually I used a arguments-based provisioner (providing `--token` argument). I checked the documentation and [source code](https://github.com/vercel/vercel/tree/main/packages/cli) for the vercel-cli, but I couldn't find any evidence that it reads tokens from environment variables.

While we could use a file-based provisioner instead, I think that using the `--token` argument is the simplest approach.

plugins/vercel/api_token.go
```diff
@@ -13,7 +13,7 @@ import (
 func APIToken() schema.CredentialType {
 	return schema.CredentialType{
 		Name:          credname.APIToken,
-		DocsURL:       sdk.URL("https://vercel.com/docs/cli/login"),
+		DocsURL:       sdk.URL("https://vercel.com/docs/rest-api"),
```
florisvdg, line 16: Given the lengthiness of the page, it may be nice to use the fragment:
```suggestion
		DocsURL:       sdk.URL("https://vercel.com/docs/rest-api#authentication"),
```

plugins/vercel/api_token.go
hculea: My apologies, slightly misread the code! I think the initial version of the code was also importing from `VERCEL_TOKEN`, reason why I assumed it would work. This is fine, in this situation. Thanks for clarifying!

hculea (approved):
Looks good! Thank you for your contribution! ❤️

j178:
A gentle ping to @florisvdg, would you be willing to give this  second approve and merge it?

AndyTitu:
Hey @j178 these commits seem to not have been verified and and therefore we cannot merge the PR yet. Could you amend those and sign them as you did with the rest of your commis?

j178:
@AndyTitu Hi, I rebased and signed all the commits, thanks.

PR #276, opened 2023-05-31
new(redis-cli): Add support for Redis CLI with password as environment variable and command-line flags

### LINKED ISSUE #275: New plugin: Redis CLI
### Platform or tool

Redis

### Relevant links

- Product: https://redis.io/
- CLI docs: https://redis.io/docs/ui/cli/

### Goal or desired behavior

Connecting to a redis database hosted by any redis hosting company.

### Contribution

### Additional information

_No response_

arunsathiya:
## Overview

This PR adds support for redis-cli. The password for connecting to the redis server can be set as an environment variable and it'll be used by the shell plugin.

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: https://github.com/1Password/shell-plugins/issues/275

## How To Test

- Sign up for a Redis hosting provider like [Upstash](https://upstash.com/).
- Create a Redis server.
- Make note of the password, hostname and port number shown by that provider.
- Clone this branch and run `make redis/build`
- Set up redis shell plugin by running `op plugin init redis-cli`. Set the password as a 1Password field item.
- Run `redis-cli -h host-address-here.com -p port-number-here PING`
- See response `PONG`

image

## Changelog
  

Add support for Redis CLI with environment variable-based provisioning.


arunsathiya:
A couple of quick notes:

- ManagementURL is not set for the credential because, depending on the redis hosting provider the link will be different.
- We provision using environment variables for now, because `REDISCLI_AUTH` is the only support variable, for the password. [There isn't one for the host, port or the database](https://github.com/redis/redis/blob/c2f1815bcbf5a02ae3920c2efa31099226a966bb/src/redis-cli.c#L77-L84).
- There isn't a config file provisioner for redis CLI. So, that's not an option.
- I was [exploring a new arguments provisioner for tunnelto](https://github.com/1Password/shell-plugins/pull/171#issuecomment-1523641351), but it's on hold at this time. I can take another stab at it soon and when it's ready, we can adopt the same concept here for redis, so that host, port and database can be set up as individual field items too.

Let me know what you think about this!

hculea:
Any reason not to use, in conjunction with the envvar provisioner, a command line provisioner (i.e. adding the port and host flags within the provisioner itself). The entire provisioner should be the equivalent of:
```
REDISCLI_AUTH=<value> redis-cli -p <port> -h <host>
```

plugins/redis/password.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func Password() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.Password,
```
florisvdg, line 14: Looking at the [Redis authentication docs](https://redis.io/docs/management/security/#authentication), they refer to this option as "legacy authentication". I do agree that we should offer this, but we may also want to offer the newer user-based [ACL authentication](https://redis.io/docs/management/security/acl/) option too.

Maybe we should distinguish those as:
1. `Database Password` (mentioned [here](https://redis.io/docs/management/security/#:~:text=and%20providing%20a-,database%20password,-using%20the%20requirepass)) or `Server Password` (loosely mentioned [here](https://redis.io/commands/auth/#:~:text=If%20the%20Redis%20server%20is%20password%20protected%20via%20the%20requirepass%20option.) and [here](https://redis.io/commands/auth/#:~:text=password%20in%20the%20configuration%20file%2C%20the%20server)).
4. `User Credentials` or `Database User`

I'd also be okay with leaving it as `Password` and adding support for ACL user auth later, as long as we make sure that if we print it in `op`, we prefix it with the platform name (so `Redis Password`).

arunsathiya: @florisvdg, thanks for the feedback! ACL authentication is definitely something on my mind and I discussed that with @AndyTitu ahead of this work as well. It just happened to slipped through when I started working/was reviewing [this particular Redis CLI documentation page](https://redis.io/docs/ui/cli/). At this stage, I am trying to tackle @hculea's [feedback](https://github.com/1Password/shell-plugins/pull/276#issuecomment-1569616959) first and once that's sorted, I'll look into ACL authentication as well.

>as long as we make sure that if we print it in op, we prefix it with the platform name

May I ask what you are referring to here? Is it the credential display whenever `op plugin init` or `op plugin inspect` is run? With that understanding, that's already the case: we show `Redis Password`.

Init phase:

image

image

Inspect:

image

arunsathiya:
@hculea, that's a great suggestion, thank you! 

So, today's the first time I properly dived into the provisioner SDK and I see that it's possible to provision two types of credentials (environment variables and command-line flags, in this case) in a single provisioner. The current shell plugin supports this behavior with a new `AddArgsImmediatelyAfterExecutableName` provisioner function. 

Examples to test with:

- `redis-cli ping` actually translates into `redis-cli -h host.com -p 1234 ping` behind the scenes, and outputs `pong`.
- `redis-cli incr counter_value` translates into `redis-cli -h host.com -p 1234 incr counter_value` behind the scenes, and outputs `(integer) 1`

While Redis CLI supports multiple command line flags, I have worked here with the assumption that host and port are minimum-required flags. If the user supplies additional flags, they'll be used properly. 

Example:

- `redis-cli -r 5 INCR counter_value` translates into `redis-cli -h host.com -p 1234 -r 5 INCR counter_value` behind the scenes, and outputs the following:

```
(integer) 1
(integer) 2
(integer) 3
(integer) 4
(integer) 5
```

Also, if the user decides to use a custom value for the host and port, as opposed to the value we have on the 1Password item, that shouldn't be a problem either. The flag value entered by the user will be used and the one from the 1Password item be skipped.

Let me know what you think of this setup!

arunsathiya:
>The current shell plugin supports this behavior with a new AddArgsImmediatelyAfterExecutableName provisioner function.

For some additional context on why this new function is needed, `AddArgs` function appends the command-line arguments to the end of the user-inputted command. That is, `redis-cli incr counter_value` becomes `redis-cli incr counter_value -h host.com -p 1234` but this is not supported by Redis CLI for some reason. It results in the following error:

```
AUTH failed: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
(error) ERR wrong number of arguments for 'incr' command
```

Redis CLI expects the flags to be immediately after the executable name, thus the new function `AddArgsImmediatelyAfterExecutableName`.

arunsathiya:
Couple of additional notes for all reviewers:

- 👍🏼 Since the last iteration, we now have username as a separate field, thus allowing scoped/ACL-based user accounts.
- 👍🏼  There isn't a way to set the "default" value for the username at this time: https://github.com/1Password/shell-plugins/issues/281 I am hoping redis users will be familiar with setting the standard "default" username as the default, when they don't have ACL-based accounts.
- 👍🏼  Tested with Upstash, Redis.com (enterprise Redis cloud, from the same makers of open source Redis software), works well.
- 👎🏼  Yet to test on AWS Elasticache. More significantly, yet to test/understand how AWS-hosted redis clusters work, because based on a quick glance, their access is locked down to EC2 instances.

For the near future:

- Investigate and fix optional setting, so that it can be used on the username field: https://github.com/1Password/shell-plugins/issues/280
- Decide on how we should approach redis-Terraform compatibility. I talked about this with @AndyTitu a bit already. The main blocker is that, for our Terraform shell plugin, we'll need to use environment variables instead of command-line arguments for authentication. But the environment variables for various redis providers are different:

| Name | Provider | Link | Environment variables |
|--------|--------|--------|--------|
| rediscloud | redislabs | [Cell](https://registry.terraform.io/providers/RedisLabs/rediscloud/latest/docs) | REDISCLOUD_URL, REDISCLOUD_ACCESS_KEY and REDISCLOUD_SECRET_KEY |
| redisdb | CruGlobal | [Cell](https://registry.terraform.io/providers/CruGlobal/redisdb/latest/docs) | REDISDB_DATABASE, REDISDB_HOSTNAME, REDISDB_PORT |
| rediscloud | DevopsHomeserve | [Cell](https://registry.terraform.io/providers/DevopsHomeserve/rediscloud/latest/docs) | REDISCLOUD_URL, REDISCLOUD_ACCESS_KEY and REDISCLOUD_SECRET_KEY |

As we can see here, these environment variables vary between providers, and most importantly, vary from the open source redis software, which offers only REDISCLI_AUTH.

So, we will need to think of how to best-approach Terraform compatibility. In the meantime, I propose we push forward the current version, as there's significant value for Upstash and redis.com cloud users.

plugins/redis/password.go
arunsathiya: @florisvdg, I have left some additional notes on ACL-scoped user support here: https://github.com/1Password/shell-plugins/pull/276#issuecomment-1573342063 Feedback welcome!

AndyTitu:
Screenshot 2023-06-02 at 19 14 29

This is what Arun and I discussed earlier today.

The approach is to use 2 different credential usages (for 2 different credentials), 1 for cloud and 1 for local, and to provision them using the above strategies. 

Here is each credential's structure:
Screenshot 2023-06-02 at 19 16 40

Good to note: the local credential uses ACL provisioning when `username` is specified, and uses "legacy" password provisioning when `username` is not specified (because `username` is optional)

accraw:
I did a functional test with local docker + ACL and it worked beautifully.

plugins/redis/redis-cli.go
```diff
@@ -0,0 +1,27 @@
...
+func RedisCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Redis CLI",
+		Runs:    []string{"redis-cli"},
+		DocsURL: sdk.URL("https://redis.io/docs/ui/cli"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotWhenContainsArgs("-u"),
+			needsauth.NotWhenContainsArgs("--user"),
+			needsauth.NotWhenContainsArgs("-a"),
+			needsauth.NotForVersion(),
```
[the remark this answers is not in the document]
arunsathiya: Good spot! This is done now: https://github.com/1Password/shell-plugins/pull/276/commits/893a4ffe3162f936991c7e90dcf0b0e686c6ec80

arunsathiya:
Those failing validation checks are tracked internally now. I think we'll need to discuss and conclude how we handle those as well before we can merge this PR.

plugins/redis/redis-cli.go
```diff
@@ -0,0 +1,34 @@
...
+		Runs:    []string{"redis-cli"},
+		DocsURL: sdk.URL("https://redis.io/docs/ui/cli"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotWhenContainsArgs("-u"),
+			needsauth.NotWhenContainsArgs("--user"),
+			needsauth.NotWhenContainsArgs("-a"),
+			needsauth.NotWhenContainsArgs("--help"),
+			needsauth.NotForVersion(),
+		),
+		Uses: []schema.CredentialUsage{
```
AndyTitu, line 21: No need to use `SelectFrom` here. Let's just add the credential name `User Credentials`.  The reason for this is that redis-cli can only work with the user credentials. No other credentials have any significance for the redis cli executable.

plugins/redis/secret_key.go
AndyTitu, line 1: Nice! I love that we introduced this credential. Since https://github.com/1Password/shell-plugins/pull/279, we can just define credentials in a plugin, without using them directly in code somewhere else. This credential can just exist in the credential list here, and it can be used with Terraform for example if someone wishes to do so.

plugins/redis/env_var_and_flags_provisioner.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+type EnvVarFlagsProvisioner struct {
+	sdk.Provisioner
+
+	Schema map[string]sdk.FieldName
+}
+
+func EnvVarFlags(schema map[string]sdk.FieldName) sdk.Provisioner {
```
AndyTitu, line 16: I'd prefer if this functionality existed in `sdk/provision/` and it would be shaped as:
1. A new util for flag provisioning similar to `EnvVars`. This can also take a `schema map[string]sdk.FieldName` as argument but the keys are the names of the flags.
2. A new util that chains a general number of provisioners together. This function can take a `[]Provisioner` and then it just loops through all of them and calls the `Provision` method.

sdk/provisioner.go
```diff
@@ -107,6 +107,15 @@ func (out *ProvisionOutput) AddArgs(args ...string) {
 	out.CommandLine = append(out.CommandLine, args...)
 }
 
+// AddArgsImmediatelyAfterExecutableName can be used to add additional arguments to the command line of the provision output, but immediately after the executable name. The rest of the command line appear after the additional arguments.
+func (out *ProvisionOutput) AddArgsImmediatelyAfterExecutableName(args ...string) {
```
AndyTitu, line 111: Can we call this function `PrependArgs` (or `AddArgsFirst`) and rename `AddArgs` to `AppendArgs` (or `AddArgsLast`)?

arunsathiya:
@AndyTitu, thanks for all the feedback! I trust all changes have been addressed but let me know if you spot anything off. 

I have also done a functional testing at the latest commit, https://github.com/1Password/shell-plugins/pull/276/commits/92af35eddb85552fd5aa3eafb3060c9fba4250dd, and all cases work well.

arunsathiya:
Regarding the merge conflict, I am going to rebase with `main` once the reviews are done. I could do that right now, but I have a feeling that pushing after the rebase will change the commit IDs? 🤔 So, I'll hold off now.

sdk/provision/chained_provisioner.go
```diff
@@ -0,0 +1,35 @@
+package provision
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+)
+
+// ChainedProvisioner chains multiple provisioners together for use at the same time.
+type ChainedProvisioner struct {
```
AndyTitu, line 10: Love this new provisioner!

sdk/provision/chained_provisioner.go
```diff
@@ -0,0 +1,35 @@
...
+	sdk.Provisioner
```
AndyTitu, line 11: Is there a need for the new chain provisioner to have a provisioner field? It implements the interface simply by implementing the required functions, so I feel this is not needed here.

sdk/provision/args_provisioner.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+)
+
+// EnvVarProvisioner provisions secrets as command-line arguments.
+// ArgsPosition field is to control where the arguments are placed in the command line. We check for only "first" and provision those arguments first, otherwise we provision them last.
+type ArgsProvisioner struct {
+	sdk.Provisioner
+
+	Schema       map[string]sdk.FieldName
+	ArgsPosition string
```
AndyTitu, line 15: We could make this a map instead: `ArgsPositions map[string]int` -> represents a mapping from the arg name to the position that arg should be present in. This ties in with my other comment on `AddArgsAtIndex`

Usually it's not a good idea to use strings like "first" or "last" in order to define logic, because it's hard to discover by contributors and it's prone to errors and unhandled cases, when someone fills that in with a random string. The alternative would be to use an enum, but I think in this case, the best approach is using a structure like above to define the positions.

AndyTitu (commented):
This is getting very close! Nice work @arunsathiya !

AndyTitu:
> Those failing validation checks are tracked internally now. I think we'll need to discuss and conclude how we handle those as well before we can merge this PR.

This is the failing check:
✘ Has no more than one credential type defined. Plugins with multiple credential types are not supported yet

We no longer need this. I have confirmed by testing the latest version of redis plugin with the latest cli.

arunsathiya:
@AndyTitu Appreciate the feedback! 

I built `AddArgsAtIndex(1, argsMap)` as suggested, but there's a sticky point with it. The only two valid indices would be: 1 and len(args), where 1 corresponds to immediately after the executable name and len(args) corresponds to the end of the command line. If one were to set a different value, including 0, provisioning breaks. This especially gets complicated when CLI programs stress importance on the order of flags. 

One example is, redis handles `redis-cli -h value -p value incr key` and `redis-cli incr hey -h value -p value` differently. And if I were to do something like, `AddArgsAtIndex(2, argsMap)`, that results in `redis-cli incr -h value -p value key`, which breaks entirely. 

So, in https://github.com/1Password/shell-plugins/pull/276/commits/fe969f8134371df21450f78d0c27235ebd3f19b6, I have taken a different approach: `AddArgs(provisionImmediatelyAfterExecutable bool, args ...string)`. The value set for `provisionImmediatelyAfterExecutable` controls where the provisioned arguments go.

Let me know what you think of this!

plugins/redis/secret_key.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func SecretKey() schema.CredentialType {
```
jpcoenen, line 12: If this is specific for Redis Cloud, can we add that to the name?

Something like:
```suggestion
func RedisCloudAPIKey() schema.CredentialType {
```

plugins/redis/secret_key.go
```diff
@@ -0,0 +1,55 @@
...
+	return schema.CredentialType{
+		Name:    credname.SecretKey,
```
jpcoenen, line 14: Redis seems to call these "API Keys". Is there a specific reason we're not using `credname.APIKey` here?

plugins/redis/secret_key.go
```diff
@@ -0,0 +1,55 @@
...
+		DocsURL: sdk.URL("https://docs.redis.com/latest/rc/api/get-started/manage-api-keys/"),
```
jpcoenen, line 15: Would it make sense to also add `https://app.redislabs.com/#/access-management/api-keys` as the management URL?

plugins/redis/secret_key.go
```diff
@@ -0,0 +1,55 @@
...
+				Optional:            false,
+				Composition: &schema.ValueComposition{
+					Length: 50,
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
```
jpcoenen, line 44: Is there a specific reason we need 2 separate fields for this? It feels to me like the "User API Key" and "Account API Key" are just two variations of the exact same thing: an API key.

plugins/redis/secret_key.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func SecretKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.SecretKey,
+		DocsURL: sdk.URL("https://docs.redis.com/latest/rc/api/get-started/manage-api-keys/"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.AccessKey,
```
jpcoenen, line 18: Does Redis call these Access Keys anywhere?

plugins/redis/secret_key.go
```diff
@@ -0,0 +1,55 @@
...
+					},
+				},
+			},
+			{
+				Name:                fieldname.SecretKey,
+				MarkdownDescription: "API user key (also known as Secret Key) to authenticate to Redis Enterprise Cloud.",
+				Secret:              true,
+				Optional:            false,
+				Composition: &schema.ValueComposition{
+					Length: 50,
```
jpcoenen, line 37: Hmm. I am getting a key of length 51 when generating a User API Key. That makes me wonder whether this length is even fixed at all. I'd consider not setting this value to avoid over fitting this rule.

arunsathiya:
Thanks for the review/feedback, @jpcoenen! All suggestions have been addressed now.

sdk/provisioner.go
```diff
@@ -102,8 +102,17 @@ func (out *ProvisionOutput) AddEnvVar(name string, value string) {
 	out.Environment[name] = value
 }
 
-// AddArgs can be used to add additional arguments to the command line of the provision output.
-func (out *ProvisionOutput) AddArgs(args ...string) {
+// AddArgs can be used to add additional arguments to the command line of the provision output, immediately after the executable name or at the end, depending on the value of provisionImmediatelyAfterExecutable.
+func (out *ProvisionOutput) AddArgs(provisionImmediatelyAfterExecutable bool, args ...string) {
```
jpcoenen, line 106: Two things I'd consider here:
- The change to this signature is breaking. This means that anyone in the process of developing a plugin will have to rewrite any usage of this. This is probably not too big of a deal, but something we should try to avoid as much as possible.
- A boolean as the argument makes it a bit difficult to see what a call to this function actually does. If you write `AddArgs(true, "argument")`, it's not clear to the reader what this `true` value means without looking at the function definition. If we instead use an enum for this, we can make it very easy to read. Consider something like this:

```go

type ArgumentPosition int

const (
       ArgumentPositionFirst ArgumentPosition = iota
       ArgumentPositionLast
)

output.AddArgs(ArgumentPositionFirst, "argument")
```
Possibly there are some better variants than "First" and "Last", but the idea here is that forcing the caller to use a constant, makes the code more readable.

jpcoenen (commented):
@arunsathiya You're quick! I had not even gotten around to finishing up the second part of my review before you addressed all of my previous comments (submitted them early to get you unblocked). 

Overall, I really like what I am seeing! Especially the _extensive_ work log in this PR, that's absolutely amazing. Nice work 🙌

plugins/redis/secret_key.go
AndyTitu: This is the credential for redis cloud. The official redis cloud terraform provider uses a secret access key: https://github.com/RedisLabs/terraform-provider-rediscloud

sdk/provision/args_provisioner.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// ArgsProvisioner provisions secrets as command-line arguments.
+// ArgsPosition field is to control where the arguments are placed in the command line. A value of "true" provisions the arguments immediately after the executable name, or at the end of the command.
+type ArgsProvisioner struct {
+	sdk.Provisioner
+
+	Schema       map[string]sdk.FieldName
+	ArgsPosition map[string]bool
+}
```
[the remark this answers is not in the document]
jpcoenen: I see you've opened up a separate PR for this; thanks 🙌 

Would it be possible to completely decouple this from that other PR by using a custom provisioner for Redis for now that sets both the environment variable and the arguments? Or would that add a lot of overhead?

plugins/redis/secret_key.go
jpcoenen: Aha. Looking a bit closer at Redis' documentation, I think "Access Key" and "Secret Key" are the legacy names:

image

I think it'd be best to stick to the most up-to-date names.

arunsathiya:
>Would it be possible to completely decouple this from that other PR by using a custom provisioner for Redis for now that sets both the environment variable and the arguments? Or would that add a lot of overhead?

@jpcoenen I chatted about it with @AndyTitu too, and we agree on decoupling this from the other PR with SDK-level arguments provisioner. So, I have removed it from the redis work and introduced a custom redis provisioner for now. I trust this is ready for one final review now.

jpcoenen (commented):
Last few comments. They're mostly about removing some potentially redundant stuff. But overall I am happy with there this is now 😀 

Also gave it a test with both a regular Redis server, and using the API through Terraform. Both worked correctly.

plugins/rediscloud/plugin.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "rediscloud",
+		Platform: schema.PlatformInfo{
+			Name:     "Redis Cloud",
+			Homepage: sdk.URL("https://redis.com/"),
```
florisvdg, line 13: I'd opt for:
```suggestion
			Homepage: sdk.URL("https://redis.com/redis-enterprise-cloud/overview/"),
```

AndyTitu (approved):
This looks good to me overall! I left a few final thoughts, but approving with the confidence that they'll be resolved soon.

plugins/redis/user_credentials.go
```diff
@@ -0,0 +1,84 @@
...
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"REDISCLI_AUTH": fieldname.Password,
+}
+
+var argsToProvision = []string{
+	fmt.Sprintf("--user %s", fieldname.Username),
+	fmt.Sprintf("-h %s", fieldname.Host),
+	fmt.Sprintf("-p %s", fieldname.Port),
+}
```
florisvdg, line 84: We're looking to provision the _value_ here, right? Not the name. 🤔

florisvdg: Also, the flag name and the value should be separate args.

florisvdg: Probably simpler to set the args directly in the provisioner, doesn't make much sense anyway to split it out until we merge your args provisioner PR.

plugins/redis/user_credentials_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func TestPasswordProvisioner(t *testing.T) {
+	plugintest.TestProvisioner(t, UserCredentials().DefaultProvisioner, map[string]plugintest.ProvisionCase{
+		"default": {
+			ItemFields: map[sdk.FieldName]string{
+				fieldname.Password: "pjtxpc2gaddifapjvalggspojexample",
+			},
+			ExpectedOutput: sdk.ProvisionOutput{
+				Environment: map[string]string{
```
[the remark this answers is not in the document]
AndyTitu: Test was added above

plugins/redis/redis_cli.go
```diff
@@ -14,8 +14,10 @@ func RedisCLI() schema.Executable {
 		DocsURL: sdk.URL("https://redis.io/docs/ui/cli"),
 		NeedsAuth: needsauth.IfAll(
 			needsauth.NotWhenContainsArgs("-u"),
-			needsauth.NotWhenContainsArgs("--user"),
-			needsauth.NotWhenContainsArgs("-a"),
+			needsauth.NotWhenContainsArgs("--user"), // skip when a custom username is provided
+			needsauth.NotWhenContainsArgs("-a"),     // skip when a custom password is provided
+			needsauth.NotWhenContainsArgs("-h"),     // skip when a custom host is provided
+			needsauth.NotWhenContainsArgs("-p"),     // skip when a custom port is provided
```
jpcoenen, line 20: Can you help me understand why we need to do this? Is it to allow a user to connect to a redis instance without a password?

What if the users wants to use credentials in 1Password to connect to a redis instance on a custom host/port? (or if the 1Password item does not have the host/port specified)

plugins/redis/provisioner.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	for i, arg := range argsToProvision {
+		if i%2 == 0 {
+			argName := arg
+			fieldName := sdk.FieldName(argsToProvision[i+1])
+			if fieldValue, ok := in.ItemFields[fieldName]; ok {
+				out.AddArgsAtIndex(1, argName, fieldValue)
+			}
+		}
+	}
```
AndyTitu, line 35:
```suggestion
	
	if fieldValue, ok := in.ItemFields[fieldname.Username]; ok {
		out.AddArgsAtIndex(1, "--user", fieldValue)
	}

	if fieldValue, ok := in.ItemFields[fieldname.Host]; ok {
		out.AddArgsAtIndex(1, "-h", fieldValue)
	}

	if fieldValue, ok := in.ItemFields[fieldname.Port]; ok {
		out.AddArgsAtIndex(1, "-p", fieldValue)
	}
```

plugins/redis/redis_cli.go
AndyTitu: @arunsathiya Correct me if I'm wrong but I think the desired behaviour is overwriting the host and port a user might have set up in 1Password with the ones that the user is specifying per command basis. All while still using the password stored in 1Password.

In this case, this behaviour shouldn't be handled here in the needsAuth, because this means we will skip even the password provisioning if the user just wants to override a host/port. This should be enforced at the time of provisioning the `-h` and `-p` flags to the command line. We need to detect if these are set already, and if yes, don't override them. If they are not set already in the command line, then provisioned the ones from 1Password.

plugins/redis/provisioner.go
```diff
@@ -0,0 +1,44 @@
...
+	if value, ok := in.ItemFields[fieldname.Password]; ok {
+		out.AddEnvVar("REDISCLI_AUTH", value)
+	}
+	for i, arg := range argsToProvision {
+		if i%2 == 0 {
+			argName := arg
+			fieldName := sdk.FieldName(argsToProvision[i+1])
+			if fieldValue, ok := in.ItemFields[fieldName]; ok {
+				out.AddArgsAtIndex(1, argName, fieldValue)
+			}
```
AndyTitu, line 33: I'm still convinced that this is the easier way to write this, as I need to spend some time figuring out what this does. 

Why don't we just use the following approach? We don't need to use an array because we only have 3 values we are interested in.
```suggestion
	if fieldValue, ok := in.ItemFields[fieldname.Username]; ok && CommandLineDoesNotContainUser {
		out.AddArgsAtIndex(1, "--user", fieldValue)
	}
	if fieldValue, ok := in.ItemFields[fieldname.Host]; ok && CommandDoesNotContainHost{
		out.AddArgsAtIndex(1, "-h", fieldValue)
	}
	if fieldValue, ok := in.ItemFields[fieldname.Port]; ok && CommandDoesNotContainPort {
		out.AddArgsAtIndex(1, "-p", fieldValue)
	}
```

AndyTitu (dismissed):
@arunsathiya  I just took another look and there are 2 other comments we should address

plugins/redis/provisioner.go
florisvdg: Or maybe just set the command line directly? Something like:

```go
commandLine := []string{out.CommandLine[0], "--user", user, "--host", host, "--port", port}
commandLine = append(commandLine, out.CommandLine[1:]...)
out.CommandLine = commandLine
```

arunsathiya:
Been a while since I last worked on this. Some progress notes and testing instructions below!

**New changes**:

- Changes made in https://github.com/1Password/shell-plugins/pull/276/commits/47d5292eb521fc6b9d8ea544393e8991950673fc ensure that secrets are sourced from 1Password, but if an user provisions them manually as command line arguments, they are prioritized. 
- Authentication is not skipped by using these arguments.

**Testing instructions**:

- Set up redis shell plugin with `op plugin init redis-cli`
- Import host, port, username and password as part of the setup flow
- Run a test command like `redis-cli incr key`. What happens behind the scenes is `redis-cli -h <value> -p <value> --user <value> incr key` (redis CLI requires that these flags are immediately after `redis-cli` and before the executing command. That's why we provision these flags at index 1), and the password is provisioned as a `REDISCLI_AUTH` environment variable.
- Run a test command like `redis-cli incr key -p 19261` or `redis-cli -p 19261 incr key`. These cases should be handled as well. Behind the scenes, the provisioner does `redis-cli -h <value> -p 19261 --user <value> incr key` and `REDISCLI_AUTH` for password.

plugins/redis/redis_cli.go
scottisloud: I removed the -`h/-p/-a/--user` `NotWhenContainsArgs` rules. Auth always runs (except -`-help/version`). The provisioner only skips injecting fields for flags already on the command line, so e.g. `redis-cli -h customhost` still gets the 1Password password and any missing host/port/user fields.

resolving unless more discussion required!

plugins/redis/provisioner.go
scottisloud: Should largely have addressed this in a preliminary way. Resolving unless subsequent review leads to further discussion.

scottisloud:
Revitalizing this PR, did some substantial rebasing and improved certain aspects or updated them to align with current behaviours. 

Re-requested review from Andy since Andy was an original reviewer but @AndyTitu don't feel obligated if this now falls way outside your responsibilities (or memory).

JillRegan (approved):

scottisloud:
Thanks everyone for the work one this. This will be merged and be part of a future release of the 1Password CLI!

PR #279, opened 2023-05-31
Don't require plugins to have executables

hculea:
## Overview

Don't require plugins to have executables.
Since the Terraform plugin can consume credentials of other plugins, there is now a use-case for plugins with no executables. 
If a service offers a Terraform provider, but no CLI, Shell Plugins should still offer a solution to allow for the provisioning of that credential.
In this situation, simply not providing an executable in the creation wizard does the trick.

## Type of change
 

- [ ] Created a new plugin
- [X] Improved contributor utilities or experience

## How To Test

`make new-plugin` and leave the "Executable" prompt empty. The wizard should still work.

### DIFF
--- a/cmd/contrib/main.go
+++ b/cmd/contrib/main.go
@@ -100,9 +100,8 @@ func newPlugin() error {
 			Validate: survey.Required,
 		},
 		{
-			Name:     "Executable",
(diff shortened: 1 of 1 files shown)


accraw (approved):

AndyTitu (approved):
lgtm

PR #283, opened 2023-06-09
new(mongosh): Support mongosh executable for connecting to a MongoDB database

arunsathiya:
## Overview

This PR introduces support for the `mongosh` executable within the `atlas` shell plugin. It facilitates connecting to a MongoDB database by fetching secrets from a 1Password item.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Relates: https://github.com/1Password/shell-plugins/pull/198/

## How To Test

- Set up a MongoDB database on a host like MongoDB Atlas or AWS DocumentDB.
- Create a 1Password API Credentials item with the following fields and corresponding values: `username`, `password`, `host`, `port` (at this time this is manual because `mongosh` doesn't seem to source from a configuration file or envvars, which we could import from)
- Run `op plugin init mongosh`
- Run `mongosh`, and you should be connected to the MongoDB Shell. If you don't have a host or port, like in the case of MongoDB Atlas, rather a [Connection String](https://www.mongodb.com/docs/manual/reference/connection-string/), you can run `mongosh ConnectionString` (example: `mongosh "mongodb+srv://cluster0.ygauyzj.mongodb.net"`)

## Couple of notes

- I have configured both the executable-specific importer and the default importer to nil, because there aren't any files/envvars to source from. Because of this, `op plugin init mongosh` kind of fails, that there's no importer configured. I am not sure how to handle this, besides asking the user to create the 1Password item manually. Is there a fallback importer flow that accepts the credential fields one by one? 🤔 
- DefaultProvisioner is set to `nil` for now.
- Injected arguments are provisioned immediately after the executable name, rather than at the end. That's because we need to set the index in the shell plugin code, and placing them at 1 seems to work pretty well.

## Changelog
  

Support for mongosh executable, allowing secure connection to a MongoDB Shell for a MongoDB database.


arunsathiya:
This PR is ready to review as well, but a quick couple of notes:

- There's some duplicate files that you'd already review first on [the new args PR](https://github.com/1Password/shell-plugins/pull/282). Please ignore those files. 
- There's a failing check "no more than one credential defined", but [that's being removed as well, on the redis PR](https://github.com/1Password/shell-plugins/pull/276/commits/0100c222bd8e3b00af20bf064ff054375187d2d2). If that redis PR is merged first, I can rebase mongosh branch with main and that should resolve it.

arunsathiya:
I'll mark this in progress while I branch out `atlas` and `mongosh` as separate shell plugins.

arunsathiya:
I think this is ready for a review now. A couple of more fleshed out notes for reviewers below.

## Testing with MongoDB Atlas-hosted MongoDB database

In this case, you wouldn't have a host and port, rather a "Connection String". To test, you can run `mongosh <value>` where <value> is the Connection String. When prompted to choose the 1Password item that has the credentials, you'd choose the item that contains the username and password fields.

image

Overall, Connection String is a replacement to Host and Port together.

The provisioned command line would be of the format below and will put you on a MongoDB shell where you can run queries.

["mongosh" "--password" "value" "--username" "value" "mongodb+srv://cluster0.redacted.mongodb.net"]

image

## Testing with a MongoDB host that offers the host and port values

[AWS DocumentDB](https://aws.amazon.com/documentdb/) is a great example, but doesn't offer a straightforward way to test. There's no way to connect to that database from the internet, [besides setting up SSH tunneling or SSH port forwarding](https://docs.aws.amazon.com/documentdb/latest/developerguide/connect-from-outside-a-vpc.html). So, if you find a different MongoDB host that offers host and port values, you can store those, and the username and password as a 1Password item.

And then, you can run just `mongosh`, which should provision the secrets in the format below and put you on the MongoDB shell:

["mongosh" "--password" "value" "--username" "value" "--port" "value" "--host" "value"]

A quicker option is to run a MongoDB database locally on your computer and connect to it. Default host is `127.0.0.1`, port is `27017` and no username and password are needed. 

Take care to start the MongoDB service after installing with Hombrew: `brew services start mongodb/brew/mongodb-community`

image

arunsathiya:
Provisioner test case is failing at the moment though and I can fix that once [I get some guidance on how to navigate that on the redis work](https://github.com/1Password/shell-plugins/pull/276#discussion_r1226914077).

plugins/mongodbshell/database_credentials.go
```diff
@@ -0,0 +1,78 @@
...
+						Digits:    true,
+						Specific:  []rune{'-', '_'},
+					},
+				},
+			},
+			{
+				Name:                fieldname.Password,
+				MarkdownDescription: "Password for authenticating to the MongoDB database.",
+				Secret:              true,
+				Optional:            true,
```
AndyTitu, line 58: Let's make this mandatory, as discussed

arunsathiya: This is done in https://github.com/1Password/shell-plugins/commit/deec05137acb09d60dc8e169198e3db1a8106edc. 👍🏼

sdk/schema/executable.go
```diff
@@ -38,6 +38,11 @@ type CredentialUsage struct {
 	// from the way the credential is usually provisioned. Mutually exclusive with `SelectFrom`.
 	Provisioner sdk.Provisioner
 
+	// (Optional) The importer to use to import the credential. Overrides the DefaultImporter set in the credential schema,
+	// so should only be used if the importer deviates from the general the default importer works.
+	// Mutually exclusive with `SelectFrom`.
+	Importer sdk.Importer
```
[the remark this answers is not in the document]
arunsathiya: I am trying to remember why I set a executable-level importer but yeah, I do see now that it's unnecessary and if needed at all, should be handled in a separate PR. I have removed it now in https://github.com/1Password/shell-plugins/commit/4f7a652c07ebedffae159d4fc4622ae6f90b2a7b.

plugins/mongodb/database_credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func DatabaseCredentials() schema.CredentialType {
+	return schema.CredentialType{
```
AndyTitu, line 13: can we support the connection string you mentioned before as an optional field name as well?

then at provisioning time, we provision the connection string if it's set, and otherwise we provision the host and port, if they are set

arunsathiya: Good point! This is now handled in https://github.com/1Password/shell-plugins/pull/283/commits/9a19cd3ae4ba66339fa28d0b8a72ab254270df92.

sdk/provisioner.go
```diff
@@ -107,6 +107,15 @@ func (out *ProvisionOutput) AddArgs(args ...string) {
 	out.CommandLine = append(out.CommandLine, args...)
 }
 
+// AddArgsAtIndex can be used to add additional arguments to the command line of the provision output, at a specific index.
+func (out *ProvisionOutput) AddArgsAtIndex(index uint, args ...string) {
```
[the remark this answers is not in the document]
arunsathiya: This is now handled in https://github.com/1Password/shell-plugins/pull/283/commits/aa27eee6cf63299c54a5ccb2bc73da9a69f01ce5 and https://github.com/1Password/shell-plugins/pull/283/commits/067183e32f3c04b7534768ab654b618bb6c91b62.

plugins/mongodb/database_credentials.go
```diff
@@ -0,0 +1,83 @@
...
+					Charset: schema.Charset{
+						Lowercase: true,
+						Uppercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		DefaultProvisioner: provision.NoOp(),
+		Importer:           nil,
```
AndyTitu, line 81: We can use here `importer.NoOp()` instead of leaving it nil. Have you tried this out? Maybe this will be unblock mongsh from the perspective of the importer not working as expected.

arunsathiya:
Been a while since I last worked on this. Some progress notes and testing instructions below!

**New changes**:

- Changes made in https://github.com/1Password/shell-plugins/pull/283/commits/08e27c6295a403882ad1e0e9f0bc73465f15c373 ensure that secrets are sourced from 1Password, but if an user provisions them manually as command line arguments, they are prioritized. 
- Authentication is not skipped by using these arguments.

**Existing changes retained**:

- When a `connection string` field exists on the 1Password item, we'd use that instead of host and port arguments.

**Testing instructions**:

- Set up mongoDB shell plugin with `op plugin init mongosh`
- Manually create a API credential item on 1Password with `username`, `password`, `host`, `port` and `connection string`. If you are testing with a local mongoDB database, you'd have the host but if you are using a cloud, managed mongoDB database (like mongodb.com), you'd have a connection string instead of the host. (see below on why this should be created manually)
- Run a test command like `mongosh`. What happens behind the scenes is `mongosh --host <value> --port <value> -u <value> -p <value>` in the case where the connection string field is not available on the 1Password item. Where connection string is available, the following is provisioned behind the scenes: `mongosh <connection-string-value> -u <value> -p <value>`
- Run a test command like `mongosh <connection-string-value>`, `mongosh -u <value> <connection-string-value>` or any combination of connection string (mutually exclusive with host and port), and username and password. The arguments input should take precedence over the ones stored on 1Password.

**Reason for manually creating the 1Password item**:

Covered earlier on this PR, but as a reminder: [mongoDB CLI does not support an environment variable for the password to source from, or a config file](https://www.mongodb.com/docs/mongodb-shell/connect/), so, there's no proper importer flow for this shell plugin.

PR #285, opened 2023-06-10
Add support for Spacelift (spacectl)

damianstasik:
## Overview
This PR introduces support for `spacectl` through API Keys. You need to pass your endpoint URL, API Key ID and Secret, that's it!

## Type of change
 

- [x] Created a new plugin

## How To Test

There's a handy command to test if the credentials are indeed provisioned:
```bash
spacectl whoami
```

The output should contain JSON with the ID and name of the API Key as well as the endpoint URL.


sdk/schema/fieldname/names.go
```diff
@@ -7,6 +7,7 @@ const (
 	APIHost         = sdk.FieldName("API Host")
 	APIKey          = sdk.FieldName("API Key")
 	APIKeyID        = sdk.FieldName("API Key ID")
+	APIKeySecret    = sdk.FieldName("API Key Secret")
```
arunsathiya, line 10: Could we reuse `API Secret` as the fieldname because that seems to be [referenced in the config file generated as well](https://docs.spacelift.io/integrations/api#authenticating-with-the-api)?

image

plugins/spacelift/spacectl.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func SpaceliftCLI() schema.Executable {
+	return schema.Executable{
+		Name:      "Spacelift CLI",
+		Runs:      []string{"spacectl"},
+		DocsURL:   sdk.URL("https://github.com/spacelift-io/spacectl"),
+		NeedsAuth: needsauth.NotForHelpOrVersion(),
```
arunsathiya, line 15: I noticed that `spacectl h` is a help command as well, so we can exclude 1Password authentication for that subcommand.

arunsathiya: Also, I think, we can exclude 1Password authentication for `spacectl profile` commands, [like how Fastly shell plugin does](https://github.com/1Password/shell-plugins/pull/169). Since those commands are for controlling the profile stored locally on the config file and not querying the Spacelift platform itself, disabling auth for that makes sense to me.

plugins/spacelift/api_key.go
```diff
@@ -0,0 +1,101 @@
...
+						Symbols:   true,
+					},
+					Prefix: "https://",
+				},
+			},
+			{
+				Name:                fieldname.APIKeyID,
+				MarkdownDescription: "The ID of your Spacelift API key. Available via the Spacelift application.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
```
arunsathiya, line 36: Could we set a `Length` here? My API Key seems to be 26 characters long.

plugins/spacelift/api_key.go
```diff
@@ -0,0 +1,101 @@
...
+						Uppercase: true,
+						Digits:    true,
+					},
+				},
+			},
+			{
+				Name:                fieldname.APIKeySecret,
+				MarkdownDescription: "The secret for your API key. Only available when the secret is created.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
```
arunsathiya, line 47: And 64 characters for the API Secret.

plugins/spacelift/api_key_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+	})
+}
+
+func TestAPIKeyImporter(t *testing.T) {
+	plugintest.TestImporter(t, APIKey().Importer, map[string]plugintest.ImportCase{
+		"environment": {
+			Environment: map[string]string{
+				"SPACELIFT_API_KEY_ENDPOINT": "xhdw",
+				"SPACELIFT_API_KEY_ID": "abc123",
```
arunsathiya, line 35: When the lengths are set for the field names, you can run `make spacelift/example-secrets` to generate example secrets and use those for the tests. Here's what I generated for now: API Key `KKJNW2KPSTGJYQIBYWBEXAMPLE` and API Secret `5mtg0fl6rbta1l40vvbll5q94edy0uum1bkj09yyjang4qoutyrrnxx0oexample`.

arunsathiya (commented):
Overall, this is a great start! Thank you for contribution, @damianstasik! I couldn't get the config file importer working so far, but I am going to take a closer look soon. I'll followup. :)

plugins/spacelift/api_key.go
```diff
@@ -0,0 +1,101 @@
...
+			out.AddError(err)
+			return
+		}
+
+		for name, profile := range config.Profiles {
+			if profile.Credentials.Endpoint == "" {
+				continue
+			}
+
+			out.AddCandidate(sdk.ImportCandidate{
```
arunsathiya, line 81: Is there a reason why we are not importing the KeyId and KeySecret? Ideally, we'd want them stored in the 1Password item as well without any credentials left on the disk.

arunsathiya: Let's be sure to update the corresponding tests too.

arunsathiya (commented):
Sorry about the long delay here, @damianstasik. I just had another look and I have one more comment regarding the importer. Please give it a try and once it's ready for another review, you can ping me!

AndyTitu:
Hey @damianstasik , could we help you move this forward in any way?

damianstasik:
@AndyTitu hey, sorry for the long delay, I'll be taking a look at it today!

plugins/spacelift/api_key.go
damianstasik: @arunsathiya Is it for validation when importing a secret? I've set the `Length`, rebuilt the plugin, tried to import a new key and provided a single character as value for `APIKeyID`, no errors/warnings 🤔

plugins/spacelift/api_key_test.go
damianstasik: @arunsathiya is there any way to exclude `Endpoint` when generating these secrets? I get this error:
```
panic: runtime error: makeslice: len out of range
```
...when trying to run `make spacelift/example-secrets` with missing `Length` on the endpoint field.

plugins/spacelift/api_key.go
damianstasik: @arunsathiya I honestly don't remember, but I've added support for importing api keys, thanks!

sdk/schema/fieldname/names.go
damianstasik: I went with `API Key Secret` as it's the full internal name for it, but I see that from user's point it may be not clear that those are the same. I'll change it.

damianstasik:
@arunsathiya thank you for your review, and again sorry for long delay! I've applied your suggestions and finalized the api key importer, works great on my end! Let me know if this is closer to being shippable :)

plugins/spacelift/api_key.go
AndyTitu: Yes, as well as potentially for auto-detecting secrets when importing this API credential into 1Password

AndyTitu (approved):
code lgtm

plugins/spacelift/api_key.go
```diff
@@ -0,0 +1,108 @@
...
+						Digits:    true,
+						Symbols:   true,
+					},
+					Prefix: "https://",
+				},
+			},
+			{
+				Name:                fieldname.APIKeyID,
+				MarkdownDescription: "The ID of your Spacelift API key. Available via the Spacelift application.",
+				Secret:              true,
```
SimonBarendse, line 35: I'd expect only the `API_KEY_SECRET` to be secret. Does Spacelift have any documentation on whether the API Key ID should be considered secret?

SimonBarendse (approved):

plugins/spacelift/api_key.go
AndyTitu: Screenshot 2024-02-27 at 12 36 29

From [their docs](https://docs.spacelift.io/concepts/spacectl#usage-on-github-actions) it seems their official GitHub action treats the key id as a secret as well.

AndyTitu:
Hey @damianstasik could you please sign your first few commits as well so we could merge this?

damianstasik:
@AndyTitu done, thanks for taking a look at this PR!

Issue #278, opened 2023-05-31
MySQL compatibility with Terraform


arunsathiya:
### Platform or tool

MySQL

### Desired or expected behavior

MySQL shell plugin currently provisions credentials using a temporary file. While this approach works great for the MySQL binary, for MySQL Terraform provider ([there are tons of unofficial providers](https://registry.terraform.io/search/providers?q=mysql), haven't spotted an official provider yet), we'll need to use [MySQL environment variables](https://dev.mysql.com/doc/refman/8.0/en/environment-variables.html).

The expected behavior would be:

- Temp file provisioner only for MySQL binary.
- Env var provisioner for Terraform.

### Current behavior

Credential provisioning happens as a temporary file. 

### op CLI version

2.19.0-beta.01

arunsathiya:
This was tracked in [a parent issue](https://github.com/1Password/shell-plugins/issues/262), but we are branching it out to track them more precisely.

PR #293, opened 2023-06-15
update(mysql): Move provisioner to the executable-level CredentialUsage configuration

arunsathiya:
## Overview

MySQL provisioner is moved to the executable-level CredentialUsage configuration, thus making it work only for `mysql` commands. MySQL Terraform provider supported, if any in the future, will use the DefaultProvisioner instead. For now, the default provisioner is set to nil.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Relates: https://github.com/1Password/shell-plugins/issues/278

## How To Test

- Clone this branch and run `make mysql/build`
- Run `op plugin init mysql` and set up your MySQL secrets as a 1Password item.
- Follow testing instructions as in [the original MySQL work](https://github.com/1Password/shell-plugins/pull/27). Does the command still succeed?

## Changelog
  

Move MySQL provisioner to the executable-level CredentialUsage configuration.

### DIFF
--- a/plugins/mysql/database_credentials.go
+++ b/plugins/mysql/database_credentials.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
-	"github.com/1Password/shell-plugins/sdk/provision"
 	"github.com/1Password/shell-plugins/sdk/schema"
 	"github.com/1Password/shell-plugins/sdk/schema/credname"
 	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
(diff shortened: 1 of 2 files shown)


arunsathiya:
I have marked this as ready for review because I spent a fair amount of time trying to set up a MySQL Terraform provider but none of the three I tried seemed to work. Two issues: 

- They seem to prompt for the username (despite there being a username field on the 1Password item), 
- If I enter the username manually and move forward, it seems to get stuck on the database creation stage. 

image

image

I am not sure why these are happening, but I attribute these to [being not-fairly-regularly-maintained community providers](https://registry.terraform.io/search/providers?q=mysql). [There was one maintained by Hashicorp](https://github.com/hashicorp/terraform-provider-mysql) but it seems to be archived now.

I can continue to take a closer look at these community-maintained providers and when one of these successfully works, and when the majority of the providers confirm to have the same environment variables, we can update the MySQL DefaultProvisioner. Until then, I propose/agree that we set it to nil.

Issue #254, opened 2023-04-28
Silence AWS plugin logs


williamhpark:
### Platform or tool

AWS

### Desired or expected behavior

1. Initialize the AWS plugin: `op plugin init aws`
2. Run an `aws` command using the plugin, e.g. `aws s3 ls`

The command should run successfully with no unnecessary log output.

### Current behavior

A few unnecessary lines of log output are outputted (shown in **Relevant log output**).

The source of these logs is `/plugins/aws/sts_provisioner.go` -> `getAWSAuthConfigurationForProfile`:

`configFile, err := confighelpers.LoadConfigFromEnv()`

A solution could be to silence the log output with `log.SetOutput(io.Discard)` then re-enable it, similar to [in this PR](https://github.com/1Password/shell-plugins/pull/251):

```
// Disable log output produced by AWS Vault code
log.SetOutput(io.Discard)

// Read config file from the location set in AWS_CONFIG_FILE env var or from  ~/.aws/config
configFile, err := confighelpers.LoadConfigFromEnv()
if err != nil {
return nil, err
}

// Re-enable log output
log.SetOutput(os.Stdout)
```

### Relevant log output

```shell
2023/04/28 10:40:34 Loading config file /Users/williampark/.aws/config
2023/04/28 10:40:34 Parsing config file /Users/williampark/.aws/config
```

Issue #248, opened 2023-04-17
AWS Vault importer failing when using encrypted file as backend


williamhpark:
### Platform or tool

AWS

### Desired or expected behavior

1. Add a profile through `aws-vault` using an encrypted file as the backend: `aws-vault add file-profile --backend file`
2. Run `op plugin init aws` > `Import into 1Password...`

You should see `Encrypted file (file-profile)` as one of the import candidates, alongside any other valid import candidates that should show based on where you stored your AWS credentials.

### Current behavior

You get a long error log in the terminal, and no import candidates (AWS Vault, file importer) are shown.

The error is being triggered when calling `keyring.AvailableBackends()`. The cause is that in the `keyringConfigDefaults` var, the `FilePasswordFunc` property is not currently specified. When `keyring.AvailableBackends()` is called, an attempt is made to access `k.passwordFunc` even though it doesn't exist.

The solution would be to specify it similar to how it's done in the AWS Vault codebase:

```
func fileKeyringPassphrasePrompt(prompt string) (string, error) {
	if password, ok := os.LookupEnv("AWS_VAULT_FILE_PASSPHRASE"); ok {
		return password, nil
	}

	fmt.Fprintf(os.Stderr, "%s: ", prompt)
	b, err := term.ReadPassword(int(os.Stdin.Fd()))
	if err != nil {
		return "", err
	}
	fmt.Println()
	return string(b), nil
}

// Default keyring config values, based on the default values used by AWS Vault
// Refer to the aws-vault codebase for more context: https://github.com/99designs/aws-vault/blob/master/cli/global.go
var keyringConfigDefaults = keyring.Config{
	ServiceName:              "aws-vault",
	FilePasswordFunc:         fileKeyringPassphrasePrompt,
	LibSecretCollectionName:  "awsvault",
	KWalletAppID:             "aws-vault",
	KWalletFolder:            "aws-vault",
	KeychainTrustApplication: true,
	WinCredPrefix:            "aws-vault",
	KeychainName:             "aws-vault",
	FileDir:                  "~/.awsvault/keys/",
}
```

### Relevant log output

```shell
encountered error while importing credentials automatically: {locally built plugin panicked: runtime error: invalid memory address or nil pointer dereference
stack trace:
goroutine 55 [running]:
runtime/debug.Stack()
        /usr/local/Cellar/go/1.19.5/libexec/src/runtime/debug/stack.go:24 +0x65
[...]

PR #297, opened 2023-06-16
Silence logs generated by aws-vault on the AWS shell plugin

arunsathiya:
## Overview

When any `aws` command is run, the shell plugin currently outputs some additional logs related to the config file reader/importer. With the changes proposed on this PR, we'd hide those logs for now, but a long-term effort would be to identify why those logs are appearing and not printing them in the first place.

I have used the same solution that @williamhpark [outlined on the issue](https://github.com/1Password/shell-plugins/issues/254).

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: https://github.com/1Password/shell-plugins/issues/254

## How To Test

- Clone this branch
- Run `make aws/build`
- Set up AWS shell plugin with `op plugin init aws`
- Run any AWS command, like `aws s3 ls`
- Make sure that no additional logs appear before the `aws s3 ls`'s output appears.

Here's how it looks for me with the proposed changes:

image

## Changelog
  

Silence AWS config file import logs while running any aws commands

### DIFF
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -3,8 +3,6 @@ package aws
 import (
 	"context"
 	"fmt"
-	"io"
-	"log"
 	"os"
 	"time"
 
@@ -185,11 +183,8 @@ func (m CacheProviderFactory) NewAccessKeysProvider() aws.CredentialsProvider {
 
 // getAWSAuthConfigurationForProfile loads specified configurations from both config file and environment
 func getAWSAuthConfigurationForProfile(profile string) (*confighelpers.Config, error) {
-	// Disable log output produced by AWS Vault code
-	log.SetOutput(io.Discard)
-
 	// Read config file from the location set in AWS_CONFIG_FILE env var or from  ~/.aws/config
-	configFile, err := confighelpers.LoadConfigFromEnv()
+	configFile, err := ExecuteSilently(confighelpers.LoadConfigFromEnv)()
 	if err != nil {
 		return nil, err
 	}
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -3,6 +3,8 @@ package aws
 import (
 	"context"
 	"fmt"
+	"io"
+	"log"
 	"os"
 	"time"
 
@@ -179,6 +181,14 @@ func (m CacheProviderFactory) NewMFASessionTokenProvider(awsConfig *confighelper
 
 func (m CacheProviderFactory) NewAccessKeysProvider() aws.CredentialsProvider {
 	return accessKeysProvider{itemFields: m.ItemFields}
+}
+
+func ExecuteSilently[G interface{}, e error](f func() (G, e)) func() (G, e) {
+	return func() (G, e) {
+		log.SetOutput(io.Discard)
+		defer log.SetOutput(os.Stderr)
+		return f()
+	}
 }
 
 // getAWSAuthConfigurationForProfile loads specified configurations from both config file and environment
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -69,7 +69,7 @@ func (p STSProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, ou
 		return
 	}
 
-	tempCredentials, err := tempCredentialsProvider.Retrieve(ctx)
+	tempCredentials, err := ExecuteSilently(tempCredentialsProvider.Retrieve)(ctx)
 	if err != nil {
 		out.AddError(err)
 		return
@@ -266,7 +266,7 @@ type assumeRoleProvider struct {
 }
 
 func (p assumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
-	credentials, err := ExecuteSilently(p.AssumeRoleProvider.Retrieve)(ctx)
+	credentials, err := p.AssumeRoleProvider.Retrieve(ctx)
 	if err != nil {
 		return aws.Credentials{}, err
 	}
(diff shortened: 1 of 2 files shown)


AndyTitu:
There is another place where you could see the logs, in the aws-vault importer. Let's handle that as well.

plugins/aws/sts_provisioner.go
```diff
@@ -180,12 +182,18 @@ func (m CacheProviderFactory) NewAccessKeysProvider() aws.CredentialsProvider {
...
+
 	// Read config file from the location set in AWS_CONFIG_FILE env var or from  ~/.aws/config
 	configFile, err := confighelpers.LoadConfigFromEnv()
 	if err != nil {
 		return nil, err
 	}
 
+	// Re-enable log output
+	log.SetOutput(os.Stderr)
+
```
AndyTitu, line 196: If we choose to go this way about silencing logs, i.e. wrapping some functionality in `log.SetOutput(io.Discard)` and `log.SetOutput(os.Stderr)`, I would rather have a nice generic wrapper function for that:

```
func ExecuteSilently[G interface{}, e error](functionalityToSilence func() (G, e)) (G, e) {
	// Disable log output
	log.SetOutput(io.Discard)
	// Re-enable log output
	defer log.SetOutput(os.Stderr)

	// Execute the functionality in a silent way
	return functionalityToSilence()
}
```

And then you could call that in our aws logic as such:

```
// Read config file from the location set in AWS_CONFIG_FILE env var or from  ~/.aws/config
	configFile, err := ExecuteSilently(confighelpers.LoadConfigFromEnv)
	if err != nil {
		return nil, err
	}
```

Give it a try!

Learn more about generics here: https://gobyexample.com/generics

arunsathiya: Thanks for that suggestion, I agree this is a better approach! 👍🏼 This is updated in https://github.com/1Password/shell-plugins/pull/297/commits/b179903cef133cc290d1283f3a21cc206013297d. Do you feel that `plugins/aws/utils.go` is a okay place for that generic function to live?

plugins/aws/utils.go
```diff
@@ -20,3 +24,12 @@ func getConfigSectionByProfile(configFile *ini.File, profileName string) *ini.Se
 
 	return nil
 }
+
+func ExecuteSilently(f func() (*vault.ConfigFile, error)) func() (*vault.ConfigFile, error) {
+	return func() (*vault.ConfigFile, error) {
+		log.SetOutput(io.Discard)
+		vaultConfig, err := f()
+		defer log.SetOutput(os.Stderr)
+		return vaultConfig, err
```
AndyTitu, line 33: nit:
```suggestion
		defer log.SetOutput(os.Stderr)
		return f()
```

plugins/aws/utils.go
AndyTitu, line 1: No need to create another file, let's just include this function in `sts_provisioner`

plugins/aws/utils.go
```diff
@@ -20,3 +24,12 @@ func getConfigSectionByProfile(configFile *ini.File, profileName string) *ini.Se
...
+	}
+}
```
AndyTitu, line 35: Let's please use the example I wrote in  https://github.com/1Password/shell-plugins/pull/297#discussion_r1232187265 which uses go generics, and is not constrained by the `(*vault.ConfigFile, error)` return types.

The reason I'm insisting for a more generic function is because we need to apply it in multiple other places as per my next other comment.

AndyTitu (commented):
After some testing I've spotted 3 places where we need to add `ExecuteSilently`
Screenshot 2023-06-21 at 18 21 22

1. The first log which appears in that screenshot comes from function `configLoader.LoadFromProfile(profile)`. This means that silencing `confighelpers.LoadConfigFromEnv` is not enough. We need to silence our whole `getAWSAuthConfigurationForProfile` (run cmd + F in `sts_provisioner.go` with the function name to find where it's called)  function.

2. The second log comes from `p.AssumeRoleProvider.Retrieve(ctx)` (run cmd + F in `sts_provisioner.go` with `p.AssumeRoleProvider.Retrieve(ctx)`  to find where it's called). We need to silence this as well using the generic function.

3. Similar to 2 we have `p.SessionTokenProvider.Retrieve(ctx)` (run cmd + F in `sts_provisioner.go` with `p.SessionTokenProvider.Retrieve(ctx)`  to find where it's called) which could output a log, so let's silence that as well with the generic silencing function

arunsathiya:
I have tested both the `aws sts get-caller-identity --profile value` and importer flows, and I can't see those logs on either anymore. 

image

image

The error in second image `encountered error while importing credentials automatically` occurs even when the aws-vault importer flow is not dismissed, plus it seems to be tracked already here:

- https://github.com/1Password/shell-plugins/issues/248

I haven't taken a look at that issue in detail yet, but all the logs outlined on this PR are handled now.

plugins/aws/sts_provisioner.go
```diff
@@ -254,7 +256,7 @@ type assumeRoleProvider struct {
 }
 
 func (p assumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
-	credentials, err := p.AssumeRoleProvider.Retrieve(ctx)
+	credentials, err := ExecuteSilently(p.AssumeRoleProvider.Retrieve)(ctx)
```
AndyTitu, line 259: Let's add the execute silently in the provision method instead here: `tempCredentials, err := tempCredentialsProvider.Retrieve(ctx)` and remove it from assume role and session providers levels. This is because of an edge case mentioned in the other comment's updated section

AndyTitu (approved):
Tested latest changes locally and no logs are shown. 

Also code LGTM

hculea (approved):
Love the approach with generics!

Code looks good to me. (I did *not* functionally test, let me know if you'd like more validation there as well).

hculea (approved):

PR #289, opened 2023-06-12
Add GitHub Actions workflow to check signed commits

florisvdg:
## Overview

Adds a GitHub Actions workflow on PRs to ensure all commits are signed, using the [check-signed-commits-action](https://github.com/1Password/check-signed-commits-action).

## Type of change
 

- [ ] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [x] Improved contributor utilities or experience

## How To Test

Fork the repo and create a PR with an unsigned commit. The check should fail and you should see a comment on the PR.

## Changelog

N/A

PR #307, opened 2023-06-22
Allow commit signing PR comment on fork PRs

florisvdg:
## Overview

Follow-up to https://github.com/1Password/shell-plugins/pull/289

Currently, the commit signing check on PRs doesn't show the PR comment if the PR author is external.

This is because the passed in `GITHUB_TOKEN` does not have write access to the PR.

For more info about the problem and the solution, see [this write-up on the `actions/labeler` repo](https://github.com/actions/labeler#permissions) which has the same problem.

## Type of change
 

- [ ] Created a new plugin

## How To Test

1. Create shell-plugins fork on personal namespace
2. Create a PR from a different account that does not have write access to the fork

## Changelog

N/A

### DIFF
--- a/.github/workflows/pr-check-signed-commits.yml
+++ b/.github/workflows/pr-check-signed-commits.yml
@@ -1,9 +1,12 @@
 name: Check signed commits in PR 
-on: pull_request
+on: pull_request_target
 
 jobs:
   build:
     name: Check signed commits in PR 
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Check out code


.github/workflows/pr-check-signed-commits.yml
```diff
@@ -1,9 +1,12 @@
 name: Check signed commits in PR 
-on: pull_request
+on: pull_request_target
 
 jobs:
   build:
     name: Check signed commits in PR 
+    permissions:
+      contents: read
+      pull-requests: write
```
jpcoenen, line 9: Are you sure we can grant this? [This table](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) seems to suggest it is capped at `read` for public forks.

florisvdg: According to the `actions/labeler` docs, it should be supported when using `on: pull_request_target`:

> To resolve this issue, you can modify the on: section of your workflow to use [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) instead of pull_request (see example [above](https://github.com/actions/labeler#create-workflow)). This change allows the action to have write access, because pull_request_target alters the [context of the action](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) and safely grants additional permissions

florisvdg:
For some reason the workflow didn't run on this PR. Will have to do more investigation.

.github/workflows/pr-check-signed-commits.yml
jpcoenen: Ah, I read over that part. Thanks for clarifying.

AndyTitu (approved):

PR #336, opened 2023-06-30
Implementing the appwrite plugin

ra-jeev:
## Overview
Added Appwrite CLI shell plugin.

Since Appwrite supports an interactive CLI login using Email & Password, we can't use that approach with OP yet. Appwrite also supports a CI mode with a limitation that it only works with one **project** at a time, and we need to generate a project specific `API_KEY` with necessary permissions to do so.

Commands requiring access to more than one project (`appwrite projects ...`), which is supported by the normal CLI won't work in the CI mode.

The best approach is to configure different OP items for each project and using these items from the specific project dir.

## Type of change
- [x] Created a new plugin

## Related Issue(s)
* Resolves: #
* Relates: #

## How To Test
After setting the plugin run the following command:
`appwrite client --debug`

We can also run any of the other commands without a subcommand (e.g. `appwrite client`, `appwrite functions` etc. as it only displays a list of available subcommands).

For running project specific commands we need to run the appwrite commands from the project folder containing the `appwrite.json` file.

## Changelog

Adds the appwrite cli shell plugin to 1password CLI
  
## Additional information

### Running the appwrite commands from the HOME dir
Screen Shot 2023-06-30 at 6 07 00 PM

### Running the appwrite commands from the project dir
Screen Shot 2023-06-30 at 6 08 16 PM


plugins/appwrite/api_key.go
```diff
@@ -0,0 +1,103 @@
...
+
+func TryAppwriteConfigFile() sdk.Importer {
+	return importer.TryFile(ConfigPath(), func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		var config Config
+		if err := contents.ToJSON(&config); err != nil {
+			out.AddError(err)
+			return
+		}
+
+		fmt.Println("Printing the imported file")
```
arunsathiya, line 80: Is there a specific reason why we need these comments, or could we remove them?

plugins/appwrite/appwrite.go
```diff
@@ -0,0 +1,39 @@
...
+		Runs:    []string{"appwrite"},
+		DocsURL: sdk.URL("https://appwrite.io/docs/command-line-ci"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
+			needsauth.NotWhenContainsArgs("client"),
+			needsauth.NotWhenContainsArgs("login"),
+			needsauth.NotWhenContainsArgs("logout"),
+			needsauth.NotForExactArgs("deploy"),
+			needsauth.NotForExactArgs("projects"),
```
arunsathiya, line 22: Do I understand correctly that both `projects` and `projects <subcommand>` don't require authentication of any type? If so, we can switch to `NotWhenContainsArgs`

plugins/appwrite/test-fixtures/import_prefs.json
arunsathiya, line 1: I am a bit fuzzy on the types of files Appwrite uses:

- appwrite.json
- prefs.json
- import_prefs.json

Could you help me get a better understanding of these files by differentiating them? I think I understand the first, but the last two are not clear. 

Also about the first, are the project key and project ID stored in that file?

plugins/appwrite/test-fixtures/import_prefs.json
```diff
@@ -0,0 +1,4 @@
+{
+  "key": "zsaugacpwq6k54nnbdbmh1cys98u2a32qqkacma2ioxn1e2j6eyrk9urom0vzcvm6qbbm8s6l4xbm86n37foauiqba9tlcvohuoz87j7nwvpob5wr71k58i105fn39a10vj7ob84opwf1vrfat3m8konch7xxy2z2dh1ykohdbef7xgmvtn82lebe4mzmfzoylqy4jslrok11zbjtmd6xs84ukd7b1k9ofyuanvinmlhkgua32p5x0gqbexample",
```
arunsathiya, line 2: Is this key the same as what's stored in API_KEY?

arunsathiya (commented):
This looks like a good start, thank you for your contribution @ra-jeev!

A couple of things to move forward:

- I'd like some clarity on the three files that Appwrite supports: appwrite.json, prefs.json, import_prefs.json
- Is there a reason why we are not storing the project ID as a 1Password field? Or is that because project ID is stored in appwrite.json file of that project folder? In other words, do I understand correctly that **a combination of appwrite.json file and the API key stored in envvar API_KEY** provide authentication? 🤔

plugins/appwrite/test-fixtures/import_prefs.json
ra-jeev: @arunsathiya The last one is just a name used in the test. We can call it anything (or call it prefs.json itself). The second one is the global config which resides in the user's home directory. In normal cases it stores the logged in user's (the CLI login) JWT cookie. But in CI mode it stores the API Key (which is project specific). And appwrite.json is project specific config file which resides in the project folder.

I've explained the rationale in details in PART 1 of the accompanying article [here](https://rajeev.dev/creating-1password-plugin-and-use-it-to-build-an-app-with-nuxt3-passage-appwrite)

Will be happy to explain more if I've not been able to clarify.

plugins/appwrite/test-fixtures/import_prefs.json
ra-jeev: Yes, the format is same.

plugins/appwrite/appwrite.go
ra-jeev: No, `projects` command will give you the available subcommands without requiring a login. But anything you want to do with any of the projects, you need to be signed in. (same goes for almost all the commands above)

ra-jeev:
@arunsathiya I've replied to all of the comments. And you've correctly guessed the reason for not storing the project ID in the prefs.json file. The project ID is stored in the project specific folder. The global config only stores the server endpoint, api key and cookie (in case of normal email / password login scenario).

plugins/appwrite/api_key.go
ra-jeev: Removed the debug prints.

plugins/appwrite/test-fixtures/import_prefs.json
ra-jeev: Renamed the import_prefs.json to prefs.json.

plugins/appwrite/appwrite.go
arunsathiya: Thanks for clarifying. In the current setup, `appwrite projects list` requires authentication. Could we switch to `NotWhenContainsArgs` so that we can not require authentication that command? And other similar subcommands? I say this because `projects list` wouldn't work with the project-specific API key that we are using for authentication.

Let me know if something about my question/concern seems unclear.

arunsathiya: >And other similar subcommands?

I think it's just projects that needs to be switched to `NotWhenContainsArgs`. `storage` and `databases` subcommands need auth.

arunsathiya (approved):
I am going to approve this with the understanding that `projects` will be switched to `NotWhenContainsArgs`. Besides that, everything looks good to me and functional testing is done too. Thank you for your contribution, @ra-jeev!

@hculea / @AndyTitu - may I ask one of you to take a look at the code once, before we switch this PR over to `waiting-on-sec-review`? I have done a functional testing already, so don't feel compelled to do it again:

```
➜  ~ source ~/.op/plugins.sh
➜  ~ which appwrite
appwrite: aliased to op plugin run -- appwrite
➜  ~ appwrite storage listBuckets
#########################################################################
# WARNING: 'appwrite' is not from the official registry.                #
# Only proceed if you are the developer of 'appwrite'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/appwrite. #
#########################################################################
total : 0
buckets
[]
✓ Success
➜  ~ appwrite functions list
#########################################################################
# WARNING: 'appwrite' is not from the official registry.                #
# Only proceed if you are the developer of 'appwrite'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/appwrite. #
#########################################################################
total : 0
functions
[]
✓ Success
➜  ~ appwrite databases list
#########################################################################
# WARNING: 'appwrite' is not from the official registry.                #
# Only proceed if you are the developer of 'appwrite'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/appwrite. #
#########################################################################
total : 0
databases
[]
✓ Success
```

plugins/appwrite/appwrite.go
ra-jeev: Thanks @arunsathiya. I've pushed the change.

ra-jeev:
I've pushed the requested change. Please let me know if any other information is needed.

Thanks

plugins/appwrite/appwrite.go
```diff
@@ -0,0 +1,39 @@
...
+			needsauth.NotWhenContainsArgs("client"),
+			needsauth.NotWhenContainsArgs("login"),
+			needsauth.NotWhenContainsArgs("logout"),
+			needsauth.NotForExactArgs("deploy"),
+			needsauth.NotWhenContainsArgs("projects"),
+			needsauth.NotForExactArgs("storage"),
+			needsauth.NotForExactArgs("teams"),
+			needsauth.NotForExactArgs("users"),
+			needsauth.NotForExactArgs("account"),
+			needsauth.NotForExactArgs("avatars"),
```
hculea, line 27: These looks like commands that would be scoped to a specific account. 

Is there any documentation you could point me to about these commands? I wonder why they would not require auth.

hculea (commented):
Thanks for the tag @arunsathiya .

Only have a couple remarks.

plugins/appwrite/api_key.go
```diff
@@ -0,0 +1,99 @@
...
+				Name:                fieldname.Endpoint,
+				MarkdownDescription: "Appwrite server endpoint.",
+				Secret:              false,
+				Optional:            false,
+			},
+		},
+		DefaultProvisioner: provision.TempFile(appwriteConfig, provision.AtFixedPath(ConfigPath())),
+		Importer: importer.TryAll(
+			TryAppwriteConfigFile(),
+		)}
```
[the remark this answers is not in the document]
ra-jeev: Done

plugins/appwrite/api_key.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func appwriteConfig(in sdk.ProvisionInput) ([]byte, error) {
+	config := Config{
+		APIKey:   in.ItemFields[fieldname.APIKey],
+		Endpoint: in.ItemFields[fieldname.Endpoint],
+	}
+
+	contents, err := json.MarshalIndent(&config, "", "  ")
```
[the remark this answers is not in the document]
ra-jeev: Done

plugins/appwrite/api_key.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+func ConfigPath() string {
+	configDir, err := os.UserHomeDir()
+	if err != nil {
+		return "~/.appwrite/prefs.json"
+	}
+
+	return configDir + "/.appwrite/prefs.json"
+}
```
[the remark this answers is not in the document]
ra-jeev: Changed the function to return hardcoded string

plugins/appwrite/appwrite.go
ra-jeev: @hculea without any argument these commands behave as if they were called with --help argument. So they just give you the list of available subcommands and options. Also, these are project specific commands. So if you run `appwrite users list` from the project folder (can only be run from project folder which contains the `appwrite.json` config file), you'll get the list of all users of that project.

Here is the [CLI documentation](https://appwrite.io/docs/command-line-commands) (they don't say the above in the docs, but all the examples given there are with arguments). I think @arunsathiya has verified this.

arunsathiya: From my functional testing: The project token that we provision seems to work for `users list` and `teams list`. I'd imagine that it works for the rest of the subcommands for `users` and `teams`.

However, `account get` (one of the many subcommands under `account`) does not work with the project token, rather requires account scope. I think that's expecting a combination of username and password?

```
➜  ~ appwrite account get
#########################################################################
# WARNING: 'appwrite' is not from the official registry.                #
# Only proceed if you are the developer of 'appwrite'.                  #
# Otherwise, delete the file at /Users/arun/.op/plugins/local/appwrite. #
#########################################################################
✗ Error app.6495df658da7c618cab3@service.cloud.appwrite.io (role: applications) missing scope (account)
```

Not too sure how `avatars`' subcommands work though. 

>without any argument these commands behave as if they were called with --help argument

This matches my testing too. 👍🏼

ra-jeev: `account` commands work with currently logged in user (requires a session). `users` is a more appropriate command from server's perspective. `account` is for impersonating a user and acting in their behalf using their JWT token.

`avatars` offers below subcommands:
>getBrowser
getCreditCard
getFavicon
getFlag
getImage
getInitials
getQR

These are some utility functions for getting images/icons for stuffs. e.g. getBrowser is for getting browser icon, getFlag for getting a country's flag icon etc.

AndyTitu: So if account doesn't work with the API KEY we provision anyway (it only works with a JWT), we should make sure we also exclude this command from out auth scheme as well: `needsauth.NotWhenContainsArgs("account"),`.

As a bit of a background, where does that JWT come from? We should make sure this doesn't get placed somewhere on disk in plain-text.

ra-jeev: @AndyTitu The JWT comes when you use a Login/Password to login to the CLI (which we can't use with 1Passsword as this is an interactive process). Also, in this flow the JWT actually gets written to the same config file (~/.appwrite/prefs.json) on disk, again there won't be any JWT if we're using the CI mode (which we're doing with this integration).

You're right that the `account` commands don't work with an API KEY. Please let me know if I need to make this change `needsauth.NotWhenContainsArgs("account")`.

Thanks,
Rajeev

accraw: Yes, I think that change makes sense.

Sorry for the delayed reply!

ra-jeev: No problem @accraw. Don't think this will be merged so it is okay. :-)

PR #338, opened 2023-07-01
Add support for Pipedream CLI

rajapri28613:
## Overview

Add support for Pipedream CLI. The shell plugin provisions a configuration file at `~/.config/pipedream/config`. The shell plugin also imports the API Key and the Org ID from the same file at the same path, if they exist.

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

- Install Pipedream CLI and set up the shell plugin.
- If the config file exists at `~/.config/pipedream/config`, the first time `op plugin init pd` is run, the API Key and Org ID can be imported into 1Password.
- Test with `pd login status`. It should show that the login is successful.

## Changelog

Add support for Pipedream CLI
  
## Additional information

### DIFF
--- a/plugins/pipedream/api_key.go
+++ b/plugins/pipedream/api_key.go
@@ -14,8 +14,8 @@ import (
 func APIKey() schema.CredentialType {
 	return schema.CredentialType{
 		Name:          credname.APIKey,
-		DocsURL:       sdk.URL("https://pipedream.com/docs/api_key"),                 // TODO: Replace with actual URL
-		ManagementURL: sdk.URL("https://console.pipedream.com/user/security/tokens"), // TODO: Replace with actual URL
+		DocsURL:       sdk.URL("https://pipedream.com/docs/api_key"),
+		ManagementURL: sdk.URL("https://console.pipedream.com/user/security/tokens"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.APIKey,
--- a/plugins/pipedream/api_key.go
+++ b/plugins/pipedream/api_key.go
@@ -32,7 +32,7 @@ func APIKey() schema.CredentialType {
 			{
 				Name:                fieldname.OrgID,
 				MarkdownDescription: "OrgId for the Pipedream organization.",
-				Secret:              true,
+				Secret:              false,
 				Optional:            true,
 				Composition: &schema.ValueComposition{
 					Length: 9,
--- a/plugins/pipedream/api_key.go
+++ b/plugins/pipedream/api_key.go
@@ -33,6 +33,7 @@ func APIKey() schema.CredentialType {
 				Name:                fieldname.OrgID,
 				MarkdownDescription: "OrgId for the Pipedream organization.",
 				Secret:              true,
+				Optional:            true,
 				Composition: &schema.ValueComposition{
 					Length: 9,
 					Charset: schema.Charset{
@@ -71,9 +72,10 @@ func TryPipedreamConfigFile() sdk.Importer {
 				fields[fieldname.OrgID] = section.Key("org_id").Value()
 			}
 
-			if fields[fieldname.APIKey] != "" && fields[fieldname.OrgID] != "" {
+			if fields[fieldname.APIKey] != "" {
 				out.AddCandidate(sdk.ImportCandidate{
-					Fields: fields,
+					Fields:   fields,
+					NameHint: importer.SanitizeNameHint(section.Name()),
 				})
 			}
 		}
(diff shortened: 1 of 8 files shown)


rajapri28613:
Corresponding blog post: https://rajapri28613.hashnode.dev/1password-shell-plugin-for-pipedream

plugins/pipedream/api_key.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://pipedream.com/docs/api_key"),                 // TODO: Replace with actual URL
```
arunsathiya, line 17: These TODO can be removed.

plugins/pipedream/api_key.go
```diff
@@ -0,0 +1,100 @@
...
+		DefaultProvisioner: provision.TempFile(
+			pipedreamConfig,
+			provision.AtFixedPath("~/.config/pipedream/config"),
+		),
+		Importer: importer.TryAll(
+			TryPipedreamConfigFile(),
+		)}
+}
+
+func TryPipedreamConfigFile() sdk.Importer {
```
arunsathiya, line 56: Pipedream supports the concepts of profiles:

https://pipedream.com/docs/cli/login/#using-the-cli-to-manage-multiple-accounts

Could we import credentials from all the profiles? Fastly shell plugin has an example of it using [Namehint](https://pkg.go.dev/github.com/1Password/shell-plugins/sdk#ImportCandidate.NameHint):

https://github.com/1Password/shell-plugins/blob/f57be66f86be90cf47450b1b0c4226c8c669c2a5/plugins/fastly/api_token.go#L51-L68

plugins/pipedream/api_key.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	})
+}
+
+type Config struct {
+	APIKey string
+	OrgId  string
+}
+
+func pipedreamConfig(in sdk.ProvisionInput) ([]byte, error) {
```
arunsathiya, line 88: Even though my proposal above is to import all profiles into 1Password, we will still provision only one set of profile/credentials at any time, so this provision function can remain as is.

plugins/pipedream/pd.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func PipedreamCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Pipedream CLI",
+		Runs:    []string{"pd"},
+		DocsURL: sdk.URL("https://pipedream.com/docs/cli/reference/"),
+		NeedsAuth: needsauth.IfAll(
```
arunsathiya, line 15: Let's skip authentication if the command contains "-p" or "--profile" flag. @rajapri28613, are you aware of any other arguments that don't need 1Password authentication?

arunsathiya (changes requested):
This looks like a good start, thank you for your contribution @rajapri28613! I have left a few comments to address before we can move forward, including building support for importing multiple profile credentials.

plugins/pipedream/pd.go
rajapri28613: I am not aware of any others.

arunsathiya (approved):
Both importer and provisioner work fine. Thanks for the contribution!

plugins/pipedream/api_key.go
```diff
@@ -0,0 +1,102 @@
...
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+			{
+				Name:                fieldname.OrgID,
+				MarkdownDescription: "OrgId for the Pipedream organization.",
+				Secret:              true,
```
SimonBarendse, line 35: The org ID isn't the authentication secret, and we usually do not conceal these by default.

Fields that aren't marked secret are still end-to-end encrypted in 1Password and only available to folks you've shared the item with. The difference is in how we handle them in the client, like showing the plaintext by default on screen.

```suggestion
				Secret:              false,
```

SimonBarendse (approved):
Appreciate your contribution!

I have left a tiny suggestion for your consideration.

accraw (approved):

PR #340, opened 2023-07-01
Add support for Todoist CLI

rajapri28613:
## Overview

Add support for Todoist CLI. The shell plugin provisions a configuration file at `~/.config/todoist/config.json`. The shell plugin also imports the Todoist token from the same file at the same path, if it exists.

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

- Install Todoist CLI and set up shell plugin.
- Run `op plugin init todoist` to import the token from `~/.config/todoist/config.json`
- The next time `todoist` commands are run, the shell plugin provisions the token to the same file path.

## Changelog
  
Add support for Todoist CLI.

## Additional information

### DIFF
--- a/plugins/todoist/api_token.go
+++ b/plugins/todoist/api_token.go
@@ -68,7 +68,7 @@ func todoistConfig(in sdk.ProvisionInput) ([]byte, error) {
 	config := Config{
 		Token: in.ItemFields[fieldname.Token],
 	}
-	contents, err := json.MarshalIndent(&config, "", "  ")
+	contents, err := json.Marshal(&config)
 	if err != nil {
 		return nil, err
 	}
--- a/plugins/todoist/api_token.go
+++ b/plugins/todoist/api_token.go
@@ -1,0 +1,76 @@
+package todoist
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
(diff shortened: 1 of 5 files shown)


rajapri28613:
Corresponding blog post: https://rajapri28613.hashnode.dev/1password-shell-plugin-for-todoist

plugins/todoist/todoist.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TodoistCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Todoist CLI",
+		Runs:    []string{"todoist"},
+		DocsURL: sdk.URL("https://github.com/sachaos/todoist"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
```
accraw, line 17: I think we should remove this - as it stands, if you run `todoist` without any arguments it just asks for your API token and doesn't give you the help text with the list of commands. What do you think?

accraw:
The functional test worked ✅ The only thing that wasn't as expected was the command with no args.

plugins/todoist/api_token.go
```diff
@@ -0,0 +1,76 @@
...
+
+type Config struct {
+	Token string `json:"token"`
+}
+
+func todoistConfig(in sdk.ProvisionInput) ([]byte, error) {
+	config := Config{
+		Token: in.ItemFields[fieldname.Token],
+	}
+	contents, err := json.MarshalIndent(&config, "", "  ")
```
arunsathiya, line 71: Minor comment for your consideration: could we use `json.Marshal` and update the test file to be without formatting?

arunsathiya (approved):
I have tested this functionally and code proposal looks good to me. There's only one minor comment about the test file format for your consideration. Thanks for the contribution!

Approving with the assumption that it'll be updated.

accraw (approved):

arunsathiya (approved):

SimonBarendse (approved):

PR #346, opened 2023-07-07
feat: add plugin for Backblaze B2 CLI

joneskoo:
## Overview

Add plugin for Backblaze B2 CLI

## Type of change

- [x] Created a new plugin

## How To Test

1. Sign up for Backblaze Cloud Storage (free tier available)
2. Create application key in Backblaze Cloud Storage portal image
3. `op plugin init b2`
4. `b2 list-buckets` etc.

## Changelog

Authenticate the Backblaze B2 CLI using Touch ID and other unlock options with 1Password Shell Plugins.


joneskoo:
At last minute (and after having created this to explore 1Password CLI plugins), I realise this plugin will not make sense. It does work, but the b2 cli always loads the credentials into a local sqlite file so credentials will stay on disk.

https://b2-command-line-tool.readthedocs.io/en/master/

> This program caches authentication-related and other data in a local SQLite database. The location of this database is determined in the following way:
> 
> If profile arg is provided: * XDG_CONFIG_HOME/b2/db-<profile>.sqlite, if XDG_CONFIG_HOME env var is set * ~/.b2db-{profile}.sqlite



I wanted to open this pull request anyway, as I have the code, in case they change that behaviour later, or if someone else looks for existing issues why B2 CLI is not included.

I don't think this should be merged.

I'll leave it open in case you have experience with other applications with similar behaviour, otherwise best close this and leave it for others to find later.

arunsathiya (commented):
Thanks for contributing this plugin, @joneskoo! 

I am a Backblaze user and definitely see value in this. I see the same behavior that you described: the credentials are stored in plaintext, even if I delete the file at `~/.b2_account_info`. The moment I run a b2 command with environment variables, that file seems to be recreated. 

That behavior isn't ideal and I'd love to see it change one day in the future -- I see that you created an issue upstream too, thanks for that! 

I'll leave this PR open for a bit to see if anyone else has ideas on what's possible.

joneskoo:
I tested setting environment variable where the account info is written to /dev/null and it didn’t work. Also setting to `:memory:` didn’t work.

I think options are to set it to a non-disk backed storage such as tmpfs or to implement non-persistent credentials handling upstream. I’m not sure if tmpfs approach is portable or the performance overhead, e.g. unnecessary API calls cost.

accraw:
Thank you for taking the time to do so much research on this! We have 2 main goals for shell-plugins: increase security for CLIs, and make it more convenient. You make a valid point that with the credentials getting written to disk, we're not adding to the security, but I would argue that we are adding convenience, so this might not be as much of a write-off as you think. That being said, we are still having internal discussions regarding how to approach this use case, do you think the added convenience here would still make it a worthwhile plugin for you, personally? (purely an opinion question)

AndyTitu:
> In my opinion, the convenience is minor and the surprise is large.

I agree with @joneskoo, this is not a road we wish to take.

> I tested setting environment variable where the account info is written to /dev/null and it didn’t work. Also setting to :memory: didn’t work.

These sound promising in this specific case. Are these officially documented somewhere? If these don't work because of a bug it could be worth keeping this PR around until those get solved and then taking this approach.

> I think options are to set it to a non-disk backed storage such as tmpfs 

Indeed, theoretically this could work pretty nicely for our use case. We'll need to do some research on this, but thanks for the hint!

joneskoo:
> > I tested setting environment variable where the account info is written to /dev/null and it didn’t work. Also setting to :memory: didn’t work.
> 
> These sound promising in this specific case. Are these officially documented somewhere? 

No, not as fas as I know.

accraw:
Closing for now - if the referenced bug gets fixed we can look at re-opening it in the future.

PR #347, opened 2023-07-10
Add `make registry.json` command 

hculea:
## Overview

This PR adds a `make registry.json` command, generating a json-formatted registry.

For the time being, this will be used as configuration for the Shell Plugins banners that OPH renders for Shell Plugins items. However, this can also serve as boilerplate for more complex configuration that can be built on top of this repo.

## Type of change
 

- [ ] Created a new plugin
- [X] Improved contributor utilities or experience

## How To Test

Run `make registry.json`

### DIFF
--- a/sdk/schema/plugin.go
+++ b/sdk/schema/plugin.go
@@ -125,7 +125,7 @@ func (p Plugin) MarshalJSON() ([]byte, error) {
 	}
 	plugin := struct {
 		PlatformName  string `json:"platform_name"`
-		ManagementURL string `json:"management_url"`
+		ManagementURL string `json:"management_url,omitempty"`
 	}{
 		PlatformName: p.Platform.Name,
 		ManagementURL: func() string {
--- a/sdk/schema/plugin.go
+++ b/sdk/schema/plugin.go
@@ -1,6 +1,7 @@
 package schema
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/url"
 )
@@ -116,4 +117,24 @@ func (p Plugin) DeepValidate() []ValidationReport {
 	}
 
 	return reports
+}
+
+func (p Plugin) MarshalJSON() ([]byte, error) {
+	if len(p.Credentials) == 0 {
+		return nil, nil
+	}
+	plugin := struct {
+		PlatformName  string `json:"platform_name"`
+		ManagementURL string `json:"management_url"`
+	}{
+		PlatformName: p.Platform.Name,
+		ManagementURL: func() string {
+			if p.Credentials[0].ManagementURL != nil {
+				return p.Credentials[0].ManagementURL.String()
+			} else {
+				return ""
+			}
+		}(),
+	}
+	return json.Marshal(plugin)
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Auto-generated with `make registry`
 plugins/plugins.go
 
+# Auto-generated with `make registry.json`
+plugins/registry.json
+
 # GoLand
 .idea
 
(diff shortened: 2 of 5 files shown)


sdk/schema/plugin.go
```diff
@@ -117,3 +118,23 @@ func (p Plugin) DeepValidate() []ValidationReport {
...
 	return reports
 }
+
+func (p Plugin) MarshalJSON() ([]byte, error) {
+	if len(p.Credentials) == 0 {
+		return nil, nil
+	}
+	plugin := struct {
+		PlatformName  string `json:"platform_name"`
+		ManagementURL string `json:"management_url"`
```
hculea, line 128: We can consider making this `omitempty`. 

However, this would make things a bit harder in the closed-source code, so I propose we leave it like it is for now.

accraw:
Functional test, check!

Should we add plugins/registry.json to .gitignore?

accraw:
Do you think it might make sense to include the name of the executable? For example, I would be more likely to name my cred for Oh Dear `ohdear` and not have the management url.

arunsathiya (commented):
This looks good to me! A couple of questions for consideration:

- Should we support multiple credentials in the registry? When in the future multiple credentials are supported, we'd want to record separate management URLs for them. One example is ngrok: right now the authtoken and API Key are bundled together as one credential, but it'd be ideal to have them as individual credentials.
- For shell plugins without a management URL, are we directing to the docs page?

arunsathiya (commented):
**Not necessarily questions but some drive-by thought processes**: 

Multi-executable shell plugins, and shell plugins that depend on other company secrets. Homebrew is an example of the latter, which uses GitHub Personal Access Tokens for authentication. 

If we are doing pattern matching based on the platform name, for example, should we show only the GitHub shell plugin (on the banner) on a GitHub secret, or show the Homebrew shell plugin as well?

If we are pattern matching with the URL in the secret item, should we show both GitHub and Homebrew shell plugins in that banner? If we choose to show more than one, will that constraint what we can show within that banner?

sdk/schema/plugin.go
```diff
@@ -117,3 +118,23 @@ func (p Plugin) DeepValidate() []ValidationReport {
...
+		return nil, nil
+	}
+	plugin := struct {
+		PlatformName  string `json:"platform_name"`
+		ManagementURL string `json:"management_url,omitempty"`
+	}{
+		PlatformName: p.Platform.Name,
+		ManagementURL: func() string {
+			if p.Credentials[0].ManagementURL != nil {
+				return p.Credentials[0].ManagementURL.String()
```
arunsathiya, line 133: I noticed that this results in escaping of the & character into its unicode representation. So, the Homebrew management URL becomes https://github.com/settings/tokens/new?scopes=gist,repo,workflow\u0026description=Homebrew and this fails to work on GitHub (missing token name):

image

Depending on how this management URL is used on the CLI side of the banner implementation, we should make sure to use the raw unescaped string.

hculea:
Added to `.gitignore`, thanks @accraw !

> Do you think it might make sense to include the name of the executable? For example, I would be more likely to name my cred for Oh Dear ohdear and not have the management url.

I suggest we carry out this discussion on the internal MR where this json gets used, as the PR here lacks context about the current intended consumption mechanism for this json.

hculea:
> Should we support multiple credentials in the registry? 

good point! I'll make sure to open an internal ticket about this. 😄 

> For shell plugins without a management URL, are we directing to the docs page?

https://developer.1password.com/docs/cli/shell-plugins/

sdk/schema/plugin.go
hculea: These are great questions! But I recommend we take them up within the internal MR, as this PR lacks the context for them.

arunsathiya (approved):
Current changes look good to me!

accraw (approved):
Looks good to me for this iteration, can discuss improvements or other changes internally later.

PR #362, opened 2023-08-01
Add ansible-vault plugin

Sirz3chs:
## Overview

Add the possibility to use a stored password with [ansible-vault](https://docs.ansible.com/ansible/latest/vault_guide/index.html) to encrypt/decrypt strings and files.

## Type of change

- [x] Created a new plugin

## Related Issue(s)

* Resolves: #
* Relates: #

## How To Test

```sh
ansible-vault encrypt_string 'bar' --name 'foo' > test.yml
```

The output file `test.yml` should be something similar to:
```
foo: !vault |
          $ANSIBLE_VAULT;1.1;AES256
          35313266333337303865383838653937323735383737366439626361643035656164306232353938
          6363393533346134333362373466343063623661356537300a616266633538323933383139383361
          63613864356137613535356331353536653036656163363336306632653163323365313238666337
          3936613930626366320a356134393765626134346365633134383933633166643637356438356561
          6232
```
Then to confirm:
```sh
ansible localhost -m ansible.builtin.debug -a var="foo" -e "@test.yml"
```

## Changelog
  
Add the possibilty to use a stored password as password-file to use with `ansible-vault`.


plugins/ansiblevault/test-fixtures/.ansible-vault
```diff
@@ -0,0 +1 @@
+Sl9WhfECfUn43o2qEcN2GnAw2EXAMPLE
```
AndyTitu, line 1: nit: add extra line

plugins/ansiblevault/password.go
```diff
@@ -0,0 +1,61 @@
...
+		),
+		Importer: importer.TryAll(
+			TryPasswordFile(),
+		)}
+}
+
+func passwordFile(in sdk.ProvisionInput) ([]byte, error) {
+	contents := ""
+	if password, ok := in.ItemFields[fieldname.Password]; ok {
+		contents += password
```
AndyTitu, line 39: If the password is not present in the 1Password item (say it's bee deleted) we should return an error to the plugin user explaining this unexpected state.

AndyTitu (approved):
code lgtm to me mostly. I've left a couple of small comments which should be addressed before merging.

Sirz3chs:
@AndyTitu I did the changes

Issue #111, opened 2022-12-14
Add file importer for Cachix plugin


jpcoenen:
### Platform or tool

Cachix

### Desired or expected behavior

The Cachix CLI can store auth tokens to a file: `~/.config/cachix/cachix.dhall`. The shell plugin should be able to import a token in this file.

See the original [PR](https://github.com/1Password/shell-plugins/pull/97) that implemented the plugin for more details.

### Current behavior

Only tokens in the `CACHIX_AUTH_TOKEN` environment variable are imported.

abhishekd1996:
Is issue Resolved?

dethancosta:
I've been looking into this to take a crack at it. I'm not very familiar with dhall files, but the only package I could find that looks appropriate is https://github.com/philandstuff/dhall-golang/v6. I understand from reading the [contributing guidelines](https://github.com/1Password/shell-plugins/blob/main/CONTRIBUTING.md) that adding third-party dependencies is undesirable, but I'm unaware of another solution, other than hard-coding a line-by-line read of the file, which seems brittle. I've written an implementation on a local branch that uses this package and is ready for a PR, if that's acceptable.  Otherwise, I'd welcome any suggestions/pointers to other possible solutions.

PR #373, opened 2023-08-25
Add file importer for cachix plugin

dethancosta:
## Overview

Adds a file importer for the Cachix plugin. The file is stored at `~/.config/cachix/cachix.dhall` and is generated from a terminal command. To avoid bringing in a third-party dependency, the file contents are processed as a string, rather than using a dhall parser.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #111 

## How To Test

When running `op plugin init cachix`, you should now see the option to import credentials using the specified file.

## Changelog

The Cachix plugin now checks for the `~/.config/cachix/cachix.dhall` file and attempts to import an auth token using the specified file.

### DIFF
--- a/plugins/cachix/auth_token.go
+++ b/plugins/cachix/auth_token.go
@@ -1,6 +1,9 @@
 package cachix
 
 import (
+	"context"
+	"strings"
+
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
 	"github.com/1Password/shell-plugins/sdk/provision"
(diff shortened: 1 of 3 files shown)


AndyTitu (approved):
code lgtm. Thanks for your contribution!

plugins/cachix/auth_token.go
```diff
@@ -33,9 +36,41 @@ func AuthToken() schema.CredentialType {
...
+	return importer.TryFile("~/.config/cachix/cachix.dhall", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		var token string
+
+		fileString := contents.ToString()
+		// The dhall config file contains a single block with variables inside
+		fileString = strings.Trim(fileString, "{}")
+		// Remove spaces to make parsing simpler
+		fileString = strings.Join(strings.Fields(fileString), "")
+		// There should be 3 fields: authToken, hostname, binaryCaches
+		keyVals := strings.SplitN(fileString, ",", 3)
```
SimonBarendse, line 57: I propose we remove the assumption that the file will have 3 fields, so that if Cachix chooses to extend the config file in the future with an extra field, this doesn't break the importer.

```suggestion
		keyVals := strings.Split(fileString, ",")
```

SimonBarendse (commented):
Security review approved.

I just have a small suggestion to make the implementation robust for potential future extensions by Cachix.

Appreciate the contribution, @dethancosta! ❤️

plugins/cachix/auth_token.go
dethancosta: Thanks for the suggestion, I've made the change. 😊

SimonBarendse (approved):
That was quick, thank you! 🙌

PR #379, opened 2023-09-11
Remove an existing Pulumi credentials file

ringods:
## Overview

Remove an existing Pulumi credentials file before injecting the environment variables

Context: https://github.com/pulumi/pulumi-kubernetes-operator/issues/483#issuecomment-1712557930

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

Have a different Pulumi access token configured for two different directory trees. Each should login with the correct token.

### DIFF
--- a/plugins/pulumi/pulumi_access_token.go
+++ b/plugins/pulumi/pulumi_access_token.go
@@ -2,7 +2,9 @@ package pulumi
 
 import (
 	"context"
+	"fmt"
 	"net/url"
+	"os"
 	"time"
 
 	"github.com/1Password/shell-plugins/sdk"
@@ -38,7 +40,7 @@ func PulumiAccessToken() schema.CredentialType {
 				Optional:            true,
 			},
 		},
-		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		DefaultProvisioner: DeleteSecretsFile(provision.EnvVars(defaultEnvVarMapping)),
 		Importer: importer.TryAll(
 			importer.TryEnvVarPair(defaultEnvVarMapping),
 			TryPulumiConfigFile(),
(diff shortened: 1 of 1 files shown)


plugins/pulumi/pulumi_access_token.go
```diff
@@ -109,3 +111,31 @@ func TryPulumiConfigFile() sdk.Importer {
...
+
+func DeleteSecretsFile(provisioner sdk.Provisioner) sdk.Provisioner {
+	return DeleteSecretsFileProvisioner{
+		RealProvisioner: provisioner,
+	}
+}
+
+func (p DeleteSecretsFileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
+	// delete the current credentials file.
+	os.Remove(fmt.Sprintf("%s/%s", in.HomeDir, ".pulumi/credentials.json"))
```
EronWright, line 130: Perhaps it would be better to run `pulumi logout` rather than unceremoniously deleting the creds file.

AndyTitu, line 130: As discussed over our call, this won't work nicely with the security guarantees 1Password is trying to provide through the shell plugins. We are aiming to not allow any secrets on disk, and with the current approach secrets will remain on disk for a long time. So the plugin doesn't bring any security value. 

A better approach is to use the `Deprovision` function to remove the file, but we'll first need to make the deprovisioning logic available in the CLI. This still wouldn't be acceptable the the plaintext secrets will still touch the disk.

We are researching ways of better supporting temp file provisioning.

AndyTitu (changes requested):
As discussed with @ringods , this approach is not suitable for merging.

PR #332, opened 2023-06-29
Add support for InfluxDB shell Plugin

bala-ceg:
## Overview

Added support for influx cli to store the influx environmental variables

Closes https://github.com/1Password/shell-plugins/issues/311

## Type of change

- [x] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [ ] Improved contributor utilities or experience

## How To Test

```
bseetharaman@EXT-C02D27NHMD6M ~ % influx bucket
########################################################################################
# WARNING: 'influxdb' is not from the official registry.                               #
# Only proceed if you are the developer of 'influxdb'.                                 #
# Otherwise, delete the file at /Users/bseetharaman/.config/op/plugins/local/influxdb. #
########################################################################################
? Locate your InfluxDB Database Credentials: InfluxDB Database Credentials (Personal)

NAME:
   influx bucket - Bucket management commands

USAGE:
   influx bucket command [command options] [arguments...]

COMMANDS:
   create          Create bucket
   delete          Delete bucket
   list, find, ls  List buckets
   update, set     Update bucket

OPTIONS:
   --help, -h  show help
   
bseetharaman@EXT-C02D27NHMD6M ~ % influx bucket ls
########################################################################################
# WARNING: 'influxdb' is not from the official registry.                               #
# Only proceed if you are the developer of 'influxdb'.                                 #
# Otherwise, delete the file at /Users/bseetharaman/.config/op/plugins/local/influxdb. #
########################################################################################
ID			Name		Retention	Shard group duration	Organization ID		Schema Type
70c0800c8694cf95	_monitoring	168h0m0s	n/a			716490b51c18261c	implicit
1118b01e3b00234e	_tasks		72h0m0s		n/a			716490b51c18261c	implicit
bafeabd992325253	mindsdb		720h0m0s	n/a			716490b51c18261c	implicit
bseetharaman@EXT-C02D27NHMD6M ~ % 


```

## Additional information

- [x] Check this box if this is a [Hashnode Hackathon](https://hashnode.com/hackathons/1password) submission

PR #381, opened 2023-09-22
Fix signed commits CI job

SimonBarendse:
The job was previously ran on "pull_request_target". This runs the job on the base branch (i.e. the `main` branch of the
 1password/shell-plugins repo). Since we want to know if commits are signed on all the commits to be merged (i.e. also on the fork), I have changed this job to run on the merge commit instead.

### Relevant documentation
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

Note that the usage example of https://github.com/1Password/check-signed-commits-action also recommends running on the "pull_reqeusts" event.

### Use Cases
See https://github.com/1Password/shell-plugins/pull/332 where we run into this in practice. The job is passed, but commits aren't signed.

### DIFF
--- a/.github/workflows/pr-check-signed-commits.yml
+++ b/.github/workflows/pr-check-signed-commits.yml
@@ -1,5 +1,5 @@
 name: Check signed commits in PR 
-on: pull_request
+on: pull_request_target
 
 jobs:
   build:
--- a/.github/workflows/pr-check-signed-commits.yml
+++ b/.github/workflows/pr-check-signed-commits.yml
@@ -9,8 +9,5 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - name: Check out code
-        uses: actions/checkout@v3
-
       - name: Check signed commits in PR
         uses: 1Password/check-signed-commits-action@main
--- a/.github/workflows/pr-check-signed-commits.yml
+++ b/.github/workflows/pr-check-signed-commits.yml
@@ -1,5 +1,5 @@
 name: Check signed commits in PR 
-on: pull_request_target
+on: pull_request
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


florisvdg:
The [change from `pull_request` to `pull_request_target`](https://github.com/1Password/shell-plugins/pull/307) was needed to support PR commenting on PRs from forks. Otherwise the GitHub token mounted in the fork `pull_request` workflow does not have permissions to comment in the upstream repo.

However, that change alone didn't fix the commenting, because `pull_request_target` breaks the PR number check, which results in the `No PR found to scan for commits.` you're seeing.

There's a [PR open](https://github.com/1Password/check-signed-commits-action/pull/4) in the action repo that in turn fixes that. I'll make sure to also update the examples/docs in the action repo.

SimonBarendse:
> However, that change alone didn't fix the commenting, because pull_request_target breaks the PR number check, which results in the No PR found to scan for commits. log message you're seeing.
>
> There's a https://github.com/1Password/check-signed-commits-action/pull/4 in the action repo that in turn fixes that. I'll make sure to also update the examples/docs in the action repo.

Looks like this didn't work. The [job](https://github.com/1Password/shell-plugins/actions/runs/6273690974/job/17037676950) still exists with the same log message.

florisvdg (approved):
Usually not a fan of using `@main` in actions, but given that we own both repos I think it's acceptable in this case.

hculea (approved):

florisvdg:
> Looks like this didn't work. The [job](https://github.com/1Password/shell-plugins/actions/runs/6273690974/job/17037676950) still exists with the same log message.

That's because it's using `pull_request_target` which is `main` and thus not this branch. So in practice, any changes made to this particular workflow will be ignored until they're merged to `main`.

.github/workflows/pr-check-signed-commits.yml
```diff
@@ -13,4 +13,4 @@ jobs:
         uses: actions/checkout@v3
```
florisvdg, line 13: And related to this change: we should remove this step.

PR #382, opened 2023-09-22
Add Ansible vault plugin

SimonBarendse:
Expanded on https://github.com/1Password/shell-plugins/pull/362, just added tests and fixed bugs that uncovered.


accraw (approved):
Looks good to me - I ran the tests I could without having a working sensible setup

plugins/ansiblevault/password.go
```diff
@@ -0,0 +1,63 @@
...
+	if password, ok := in.ItemFields[fieldname.Password]; ok {
+		return []byte(password + "\n"), nil
+	}
+	return nil, fmt.Errorf("unable to find password field")
+}
+
+func TryPasswordFile() sdk.Importer {
+	file := os.Getenv("ANSIBLE_VAULT_PASSWORD_FILE")
+	if file == "" {
+		file = "~/.ansible-vault"
```
AndyTitu, line 49: Is this location the same for both Linux and Mac? If not, we can use the custom importer logic we're using in ngrok:

```
Importer: importer.TryAll(
			importer.MacOnly(
				TryngrokConfigFile("~/Library/Application Support/ngrok/ngrok.yml"),
			),
			importer.LinuxOnly(
				TryngrokConfigFile("~/.config/ngrok/ngrok.yml"),
			),
```

plugins/ansiblevault/ansible_vault.go
```diff
@@ -0,0 +1,25 @@
+package ansiblevault
```
florisvdg, line 1: I'd call the plugin `ansible`. The `ansible-vault` CLI is shipped as part of the `ansible` package.

PR #384, opened 2023-09-27
update(postgresql): add support for 'pgcli'

szymon:
## Overview

Add support support for `pgcli` as alternative cli client for PostgreSQL.
Pgcli is a rich, postgres client with syntax highlighting, smart completion and multiline support.

More info about pgcli: https://www.pgcli.com/

## Type of change
 

- [ ] Created a new plugin

## How To Test

Since pgcli can use the same environment variables as psql it can be used as a psql replacement:
```shell
op plugin run -- pgcli
```
or we can configure it if psql is not already configured:
```shell
op plugin init pgcli
```

## Changelog
  
PostgreSQL plugin now also supports `pgcli` as an alternative to psql.

### DIFF
--- a/plugins/postgresql/pgcli.go
+++ b/plugins/postgresql/pgcli.go
@@ -1,0 +1,22 @@
+package postgresql
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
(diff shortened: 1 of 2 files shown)


cesarferradas:
Interested in this too, so following this PR. Would be really useful to have this as a plugin!

accraw (approved):
Looks good to me, and works as expected.

florisvdg (approved):

Issue #114, opened 2022-12-15
New plugin:


martijnarts:
### Platform or tool

Docker

### Relevant links

* https://docs.docker.com/engine/reference/commandline/login/

### Goal or desired behavior

Probably implement a credential helper for the Docker CLI, since logging in using username/password is probably not ideal.

The link above also references the "very simple protocol": https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol

### Contribution

- [ ] I'd be up for building the plugin myself.

SimonBarendse:
> Probably implement a credential helper for the Docker CLI

That's a great idea! In case anyone is reading along who wants to tackle, that would be added here https://github.com/docker/docker-credential-helpers/ and could leverage https://github.com/1Password/onepassword-sdk-go to load the credentials from 1Password.

reitzig:
I thought about this a bit, since I also need _something_ connecting 1Password to the `docker` CLI. (Docker Desktop seems to be taken care of?)

If my understanding is correct, then an `op` plugin is essentially a wrapper around the target command injecting fixed credentials via env var(s) and/or file(s) which it cleans up after the target command finishes. The problem with `docker` is that there is no fixed set of credentials; the user may want to log into many different registries. What can we do about that?
 
- Require the user to add `op://` references to `.docker/config.json` in place of the credentials; the wrapper would amount to `op run` resp. `op inject` for file provisioning.
  - This leaves the main config file invalid in a certain sense, rendering `docker` unusable without the wrapper.
- Require the user to use some convention on how they store registry credentials in 1Password; provision all matching credentials.
  - Limits 1Password use; may lead to duplicate credentials just to make the plugin work.
  - Given the lack of search/query features in the 1Password SDK (if I didn't miss anything), this would mean retrieving _all_ items to the client and filtering there.
- Either way: Since we'd mess with the single one config file that serves more purposes than just storing secrets, we'd ~either~ have to point `docker` to a temporary file ~or store the original~.

Neither option satisfies me, so I've built [reitzig/docker-credential-1password](https://github.com/reitzig/docker-credential-1password) for myself, for now. This implements (part of) the helper interface the _Docker_ CLI puts forth, that is it can retrieve credentials when and as needed.

PR #386, opened 2023-10-05
Add Docker plugin

dethancosta:
## Overview

Adds a plugin to authenticate the Docker CLI using a username and secret (a password or an access token).

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #114 

## How To Test

docker login

## Changelog
  

Authenticate to a Docker registry using Touch ID and other unlock options with 1Password Shell Plugins.

### DIFF
--- a/plugins/docker/credentials.go
+++ b/plugins/docker/credentials.go
@@ -1,0 +1,89 @@
+package docker
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func Credentials() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.Credentials,
+		DocsURL:       sdk.URL("https://docs.docker.com/engine/reference/commandline/login"),
+		ManagementURL: sdk.URL("https://hub.docker.com/settings/security"),
+		Fields: []schema.CredentialField{
+			{
+				Name: fieldname.Username,
+				MarkdownDescription: "Username used in Docker registries.",
+				Secret: false,
+			},
+			{
+				Name:                fieldname.Secret,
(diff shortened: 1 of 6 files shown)


dethancosta:
Note that from the Docker binary's perspective, the password is being entered in plaintext on the command line, so it will emit a warning.
The Docker binary provides the [--password-stdin](https://docs.docker.com/engine/reference/commandline/login/#password-stdin) flag for passing in the password to stdin, but I'm not sure that this is possible from within the provisioner code.
Another method that Docker provides is through a [credential helper](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol), which would require a rewrite of op-cli to add the necessary subcommands.

accraw:
Thank you very much for your contribution! 

It doesn't look like this approach will work consistently. We like the third option you suggested, and have even sketched out a solution in this comment: https://github.com/1Password/shell-plugins/pull/301#issuecomment-1702406762

I don't _think_ this will require work on the cli side, it should be possible with just shell-plugins, but that's something that would be figured out once someone really starts to dig into this. If you don't want to take this on, it's on our radar to add but not in the immediate future, in case that influences your decision.

dethancosta:
Ah, apologies, I didn't realize there was a previous PR with the same approach. I'll take a crack at the sketched out solution possibly after the weekend, if that's alright.

accraw:
Absolutely, that would be awesome!

dethancosta:
Alright, I *think* I've implemented all the credential helper functionality, but with a few issues/caveats that I'm not sure how to resolve:
- I had to use manual creation of the config file and executable rather than the `provision` sub-package's Add*File, otherwise the docker binary hangs. Because of this, I couldn't include the files in the expected sections of the test file. This also means that the config file needs to be deleted or temporarily removed prior to the provisioner running.
- If os.MkDirAll isn't called, an error is returned that `~/.docker/config.json` is a non-existent file or directory.
- For some reason, the provisioner isn't triggered when the plugin condition `needsAuth.ForCommand("login","push","pull")` is present, even though 'docker login' is being called
- the executable emits `error unsupported operation` as docker tries to save the credentials; I'm unsure of what the desired behaviour would be for this, it could either be changed to not emit anything and return exit code 0, or to emit a message to the user that the credentials don't need to be saved (or whatever message is appropriate)

This PR likely isn't mergeable, but I figured I'd at least document the issues I ran into attempting this approach.

PR #390, opened 2023-10-20
MonogoDB CLI (mongosh) provider

tresni:
## Overview
This provider takes a bunch of optional arguments/fields. The most important one is "database". That can be used for either specifying just a database name in conjuction with the other arguments/fields, or to supply a MongoDB connection string
(e.g. `mongodb+srv://username:password@example.org:27017/database?options`).

I generally use a username, password, and a connection string without username and password in it.

I modeled after the mysql and postgres providers.

## Type of change

- [X] Created a new plugin

## How To Test
```
op init plugin mongosh
op plugin run -- mongosh
```

## Changelog
Authenticate the MongoDB CLI using Touch ID and other unlock options with 1Password Shell Plugins.

### DIFF
--- a/plugins/mongosh/database_credentials.go
+++ b/plugins/mongosh/database_credentials.go
@@ -1,0 +1,77 @@
+package mongosh
+
+import (
+	"context"
+	"github.com/1Password/shell-plugins/sdk"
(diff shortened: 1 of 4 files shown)


techcraver:
Thank you for your contribution!

PR #313, opened 2023-06-24
Added CrateDB Shell Plugin

parthiv11:
## Overview
- Added basic implementation of CrateDB using CRATEPW env Variable

- [x] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [ ] Improved contributor utilities or experience

## Related Issue(s)
  

* Resolves: #
* Relates: #

PR #394, opened 2023-10-20
Amanda/cratedb

accraw:
Just https://github.com/1Password/shell-plugins/pull/313 with the test fixed and commits signed


scottisloud (approved):
Approved (LGTM). I pushed a few commits to align the generated CLI flags with the current CrateDB docs (--user → --username, --host → --hosts). Everything else looked good on review.
Thanks for sticking with this PR and for the contribution, @accraw.

JillRegan (approved):

PR #398, opened 2023-10-31
feat: add zsh plugin

iloveitaly:
## Overview
Users can load all of the required shell config with:

`zinit load 1Password/shell-plugins`

## Type of change

- [ ] Created a new plugin

## Related Issue(s)
NA

## How To Test
Use the load command above

## Changelog
Add zsh plugin for completion and plugin aliases

### DIFF
--- a/1password.plugin.zsh
+++ b/1password.plugin.zsh
@@ -1,0 +1,10 @@
+# make sure you execute this *after* asdf or other version managers are loaded
+if (( $+commands[op] )); then
+  eval "$(op completion zsh)"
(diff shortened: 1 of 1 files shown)


Marton6 (commented):
Thank you for the contribution.

This seems to be a zsh plugin, so it might be best to contribute it to a repository centralizing zsh plugins, such as https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/1password

This repository is meant for plugins for the 1Password CLI that can be used to authenticate other CLIs.

PR #400, opened 2023-11-08
Add Shodan Command-Line Interface Plugin

kusshi94:
## Overview

Add Plugin for Shodan Command-Line Interface.

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

- Install Shodan CLI with pip
- Run `op plugin init shodan`
- Run `shodan info`

## Changelog
  

Authenticate the Shodan Command-Line Interface using Touch ID and other unlock options with 1Password Shell Plugins.

### DIFF
--- a/plugins/shodan/api_key.go
+++ b/plugins/shodan/api_key.go
@@ -14,8 +14,8 @@ import (
 func APIKey() schema.CredentialType {
 	return schema.CredentialType{
 		Name:          credname.APIKey,
-		DocsURL:       sdk.URL("https://shodan.com/docs/api_key"), // TODO: Replace with actual URL
-		ManagementURL: sdk.URL("https://console.shodan.com/user/security/tokens"), // TODO: Replace with actual URL
+		DocsURL:       sdk.URL("https://developer.shodan.io/api/requirements"),
+		ManagementURL: sdk.URL("https://account.shodan.io"),
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.APIKey,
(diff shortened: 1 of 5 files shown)


plugins/shodan/api_key.go
```diff
@@ -0,0 +1,57 @@
...
+			provision.FieldAsFile(fieldname.APIKey),
+			provision.AtFixedPath("~/.config/shodan/api_key"),
+		),
+		Importer: importer.TryAll(
+			TryShodanConfigFile(),
+		)}
+}
+
+func TryShodanConfigFile() sdk.Importer {
+	return importer.TryFile("~/.config/shodan/api_key", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
```
hculea, line 44: For many tools, the equivalent of ~/.config from Linux on MacOS is ~/Library/Application Support.

Is this the case here too?

plugins/shodan/api_key.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://developer.shodan.io/api/requirements"),
```
hculea, line 17: This looks to refer to the API, could we replace this by https://cli.shodan.io/?

plugins/shodan/shodan.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func ShodanCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Shodan Command-Line Interface",
+		Runs:    []string{"shodan"},
+		DocsURL: sdk.URL("https://cli.shodan.io"),
+		NeedsAuth: needsauth.IfAll(
+			needsauth.NotForHelpOrVersion(),
+			needsauth.NotWithoutArgs(),
```
hculea, line 17: shall we also exclude `shodan init` from here?

hculea (commented):
Thank you for your contribution! I've left a few questions.

plugins/shodan/api_key.go
kusshi94: After checking the source code of the CLI tool (https://github.com/achillean/shodan-python/blob/master/shodan/cli/settings.py), it seems that the OS does not matter and the configuration file is stored in ~/.shodan/config if the ~/.shodan/ directory exists. Otherwise, it is stored in ~/.config/shodan/.
I modified the importers to try reading from both locations, and changed the provisioner to use the ~/.shodan/ directory.

plugins/shodan/api_key.go
kusshi94: I replaced the link with the one you provided.

plugins/shodan/shodan.go
kusshi94: I appended `needsauth.NotForExactArgs("init")` to the arguments of `needsauth.IfAll()`.

kusshi94:
@hculea Thank you for your review and suggestions! I have addressed them. Please let me know if the changes are satisfactory or if there are any other modifications I should make.

Issue #132, opened 2022-12-27
New plugin: oci cli (Oracle Cloud 


alon-hearter-sage:
### Platform or tool

OCI

### Relevant links

https://github.com/oracle/oci-cli

### Goal or desired behavior

Similar to the AWS plugin.
Authenticate the oci cli tool with a token or key.

### Contribution

- [X] I'd be up for building the plugin myself.

SimonBarendse:
Hi Alon, welcome to the community! :wave: A plugin for OCI would be a great addition! :raised_hands: Thank you for offering to build the plugin. We're here to help in case you have any questions or run into any issues. You can ask questions here on GitHub and via [all of the other ways to reach us](https://github.com/1Password/shell-plugins#-contributing).

Welcome! :green_heart:

alon-hearter-sage:
Oops. I opened this issue with the wrong user :)
Could you assign my personal account? @barlevalon

michaelAbon1p:
> Oops. I opened this issue with the wrong user :) Could you assign my personal account? @barlevalon

We can't assign that particular account, since it hasn't participated in this repo yet. But, consider it assigned. :)

SimonBarendse:
@barlevalon have you had the chance to give writing this plugin a try? We're here to help in case you run into any issues and if you have questions

barlevalon:
Unfortunately no, not yet.

giovannicandido:
Hi I started the work in https://github.com/1Password/shell-plugins/pull/401 it probably works but I comment in the pr way I could not test it (I use windows with WSL). As son as I figureout how to test it I can increment the PR.

PR #401, opened 2023-11-15
experimentation oci cli

giovannicandido:
## Overview

This is a work in progress not tested yet. I need feedback

Add support for oracle cloud infrastructure command line.

## Type of change

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #132 

## How To Test

I need feedback on how to test this. I use linux in WSL and my 1password desktop app is on windows. Could not find a way to develop plugins in Windows, and couldn't find a way to connect the 1password cli on WSL with the app running in windows.

The instructions is as follows:

Generate a keyfile adding a new api signing file https://docs.oracle.com/en-us/iaas/Content/Identity/Tasks/managingcredentials.htm

Get the fingerprint and private key contents, your tenancy ocid and user ocid

Add the plugin  as instructions in development and import the values. (not sure if private key needs to be in base64, it is unformated in the import because it gets only first line of certificate)

## Changelog
  

Authenticate the OCI CLI using 1password


hculea:
Hey @giovannicandido, thank you for opening this MR! 

Unfortunately, at the time, there is no straight-forward way to test shell plugins via WSL. Is using a Linux VM a viable option for you? We've recently made changes to the prompting experience that imply that system authentication/Windows Hello is no longer required, so in a VM it should be quite easy to test.

Do let us know if we can help with that, and feel free to reach out directly in the [Developer Slack workspace](https://developer.1password.com/joinslack)!

giovannicandido:
> Hey @giovannicandido, thank you for opening this MR!
> 
> Unfortunately, at the time, there is no straight-forward way to test shell plugins via WSL. Is using a Linux VM a viable option for you? We've recently made changes to the prompting experience that imply that system authentication/Windows Hello is no longer required, so in a VM it should be quite easy to test.
> 
> Do let us know if we can help with that, and feel free to reach out directly in the [Developer Slack workspace](https://developer.1password.com/joinslack)!

Right, I will try in a Linux VM.

giovannicandido:
> I tested on OSX and it works. There seems to be some nuance on the format of the secret access key if creating manually, but the import generates the field correctly.

Thanks for the test, good to know it works. I will try this one in a Linux VM and polish the code to publish it. The import from local file fallback is not implemented. Maybe it is useful, but I think it deceives the point of using 1Password because I use this to not have key files unencrypted in the file system and have a single place to authenticate.

PR #404, opened 2023-11-24
Add Exercism plugin

### LINKED ISSUE #142: New plugin: Exercism
### Platform or tool

Exercism

### Relevant links

- https://exercism.org/docs/using/solving-exercises/working-locally

### Goal or desired behavior

Authenticate the Exercism CLI with API token stored in 1Password.

### Contribution

### Additional information

_No response_

dethancosta:
## Overview
Adds a plugin for the Exercism CLI. A config file at `~/.config/exercism/user.json` is provisioned/imported. This file contains a URL to the Exercism API, an API key, and a directory in which the user's Exercism workspace is located.

## Type of change
- [x] Created a new plugin

## Related Issue(s)
* Resolves: #142 

## How To Test
Run any exercism command other than help or version.
E.g., `exercism download --track=cpp --exercise=armstrong-numbers`

## Changelog
The user can now use the Exercism plugin to authenticate to the Exercism CLI.

### DIFF
--- a/plugins/exercism/api_key.go
+++ b/plugins/exercism/api_key.go
@@ -1,0 +1,94 @@
+package exercism
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://exercism.org/docs/using/solving-exercises/working-locally"),
+		ManagementURL: sdk.URL("https://exercism.org/settings/api_cli"),
+		Fields: []schema.CredentialField{
+			{
+				Name: fieldname.URL,
+				MarkdownDescription: `The URL of the Exercism API.`,
+				Secret: false,
+			},
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "API Key used to authenticate to Exercism.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Length: 37,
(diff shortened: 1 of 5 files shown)


plugins/exercism/api_key.go
```diff
@@ -0,0 +1,94 @@
...
+			},
+			{
+				Name:                sdk.FieldName("Directory"),
+				MarkdownDescription: `The path to the workspace directory where the exercises are stored.`,
+				Secret:              false,
+			},
+		},
+		DefaultProvisioner: provision.TempFile(
+			tempFileConfig,
+			provision.AtFixedPath("~/.config/exercism/user.json"),
```
hculea, line 46: For many tools, the equivalent of `~/.config` from Linux on MacOS is `~/Library/Application Support`.

Is this the case here too?

plugins/exercism/exercism.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func ExercismCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Exercism CLI",
+		Runs:    []string{"exercism"},
+		DocsURL: sdk.URL("https://exercism.org/docs/using/solving-exercises/working-locally"),
+		NeedsAuth: needsauth.IfAll(
```
hculea, line 15: I think there may be some other situations as well when we don't want to prompt, for example for a command such as `exercism configure --token=<your-api-token>`. 

Is `--token` a flag available for all the commands? In that case, we may want to exclude this flag from prompting altogether.

hculea (commented):
Thank you for your contribution! I've left a few questions

plugins/exercism/api_key.go
dethancosta: I actually had to find the config file on my Mac since the documentation doesn't give the location, so it is indeed `~/.config` for MacOS.

plugins/exercism/exercism.go
scottisloud: fixed

scottisloud (approved):
Updated some values and made a few corrections in response to Horia's recommendations to the contributor. 

Tested clean but I have a quirky environment right now so will have a teammate give it another whirl as a gut check. Otherwise LTGM

PR #408, opened 2023-11-30
Prepend arg for defaults-file

### LINKED ISSUE #406: 1Password mysql shell plugin breaks additional CLI arguments
### Platform or tool

mysql

Invoking `mysql additional arguments` should work as if invoking native `mysql` binary

### Current behavior

Supplying additional arguments while using the 1Password mysql shell plugin integration returns error:

```
mysql: [ERROR] unknown variable 'defaults-file=/var/folders/hp/7hxtb9_968d3vj4q13z28vjc0000gn/T/1PasswordShellPlugins-413974828/my.cnf'.
```

### Relevant log output

```shell
I investigated a bit, and given an invocation like `mysql dbname`

The plugin prepares the temporary file and appends it to the CLI arguments like so:


/opt/homebrew/opt/mysql-client/bin/mysql dbname --defaults-file=/var/folders/hp/7hxtb9_968d3vj4q13z28vjc0000gn/T/1PasswordShellPlugins-413974828/my.cnf

mysql: [ERROR] unknown variable 'defaults-file=/var/folders/hp/7hxtb9_968d3vj4q13z28vjc0000gn/T/1PasswordShellPlugins-413974828/my.cnf'.
```

It appears however that the position matters to the underlying `mysql` binary, and arguments like `--defaults-file` need to come before additional arguments (like `dbname` or `-e foo.sql` ...)

This order works:
```
/opt/homebrew/opt/mysql-client/bin/mysql dbname --defaults-file=/var/folders/hp/7hxtb9_968d3vj4q13z28vjc0000gn/T/1PasswordShellPlugins-413974828/my.cnf
```
```


### op CLI version

2.23.0

### PR DESCRIPTION
**Pull request by nr-dbuckwalter**
## Overview
The `mysql` CLI depends on the `--defaults-file=foo` argument being the *very first* argument.  This change provides a new 'PrependArgs' method and uses it for the `mysql` plugin.

From the `mysql` CLI help text:
```
The following groups are read: mysql client
The following options may be given as the first argument:
--print-defaults        Print the program argument list and exit.
--no-defaults           Don't read default options from any option file,
                        except for login file.
--defaults-file=#       Only read default options from the given file #.
--defaults-extra-file=# Read this file after the global files are read.
--defaults-group-suffix=#
                        Also read groups with concat(group, suffix)
--login-path=#          Read this path from the login file.
```

## Type of change

- [ ] Created a new plugin

## Related Issue(s)

* Resolves: #406

## How To Test

Build the plugin locally and run `op plugin run mysql -- --host=foo` using a credential with no `host` field.  Before this PR, it results in the error: `mysql: [ERROR] unknown variable 'defaults-file=...` because the MySQL treats unknown arguments in the form `--variable-name=foo` as MySQL configuration variable. After the PR, additional command-line arguments are appended after `--defaults-file` that is injected by the `mysql` plugin.

## Changelog
- The `mysql` CLI depends on the `--defaults-file=foo` argument being the very first argument

### DIFF
--- a/sdk/provision/file_provisioner.go
+++ b/sdk/provision/file_provisioner.go
@@ -159,7 +159,7 @@ func (p FileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, o
 			argsResolved[i] = result.String()
 		}
 
-		out.AddArgs(argsResolved...)
+		out.PrependArgs(argsResolved...)
 	}
 }
 
--- a/sdk/provisioner.go
+++ b/sdk/provisioner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"path/filepath"
+	"slices"
 	"time"
 )
 
@@ -105,6 +106,11 @@ func (out *ProvisionOutput) AddEnvVar(name string, value string) {
 // AddArgs can be used to add additional arguments to the command line of the provision output.
 func (out *ProvisionOutput) AddArgs(args ...string) {
 	out.CommandLine = append(out.CommandLine, args...)
+}
+
+// PrependArgs can be used to insert additional arguments at the beginning of the command line of the provision output.
+func (out *ProvisionOutput) PrependArgs(args ...string) {
+	out.CommandLine = slices.Insert(out.CommandLine, 1, args...)
 }
 
 // AddSecretFile can be used to add a file containing secrets to the provision output.

### CODE REVIEW

**Comment by nr-dbuckwalter** [comment]
Added my GPG key to Github -- commits are signed

File: sdk/provisioner.go [hunk 1]
```diff
@@ -107,6 +108,11 @@ func (out *ProvisionOutput) AddArgs(args ...string) {
 	out.CommandLine = append(out.CommandLine, args...)
 }
 
+// PrependArgs can be used to insert additional arguments at the beginning of the command line of the provision output.
+func (out *ProvisionOutput) PrependArgs(args ...string) {
+	out.CommandLine = slices.Insert(out.CommandLine, 1, args...)
```
AndyTitu, line 113: In general it's a bit weird to place flags as very first arg, but for `mysql`  this seems like the standard: https://dev.mysql.com/doc/refman/8.2/en/mysql.html

AndyTitu: Probably there are other tools out there as well that do that so I'm also fine with having this as a re-usable method.

AndyTitu (approved):
lgtm

minaguib:
Any chance to get this approved and released ?  The plugin is otherwise broken for all but the simplest usecases.

nr-dbuckwalter:
@AndyTitu can you help with getting a second approver?  Any chance you could add a `CODEOWNERS` or some other mechanism to help find the right reviewers in the future?

francoisjacques (approved):
Prepending appears a reasonable and pragmatic approach.

minaguib:
Any maintainers able to merge to get this officially into next 1password-cli release ?

francoisjacques:
@SimonBarendse this seems ready to go. Can you provide guidance if there are any steps missing?

nr-dbuckwalter:
@mrjones2014 I see you are a recent committer/merger -- can you approve and merge this please?

sdk/provision/file_provisioner.go
```diff
@@ -159,7 +159,7 @@ func (p FileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, o
 			argsResolved[i] = result.String()
 		}
 
-		out.AddArgs(argsResolved...)
+		out.PrependArgs(argsResolved...)
```
mrjones2014, line 162: Do we want to do this unconditionally or just for `mysql`? We should verify this doesn't break any of the existing plugins first.

nr-dbuckwalter: Fair enough, though it's kind of unique to the `mysql` CLI that flag-type arguments be order-sensitive.

It's been a long while since I made this PR but as I recall, scoping the change to a single plugin was more involved.  I will do what I can but may not have any time in the near future to fix up my PR limit the argument ordering change to a single plugin.

mrjones2014: It may be fine as is. We just need to verify it doesn’t break anything.

HighMileage: I wanted to surface that [this PR](https://github.com/1Password/shell-plugins/pull/571) extends the work here and addresses @mrjones2014 concerns

PR #410, opened 2023-12-06
Fixes for github enterprise tokens

eedgar:


plugins/github/personal_access_token.go
```diff
@@ -64,7 +64,11 @@ func PersonalAccessToken() schema.CredentialType {
 }
 
 var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"GH_TOKEN": fieldname.Token,
 	"GITHUB_TOKEN": fieldname.Token,
+	"GH_ENTERPRISE_TOKEN": fieldname.Token,
+	"GITHUB_ENTERPRISE_TOKEN": fieldname.Token,
+        "GH_HOST": fieldname.Host,
```
AndyTitu, line 71: I wouldn't provision all fields like this. To determine which to provision between `GH_TOKEN` and  `GITHUB_TOKEN` I would find out which one takes priority if I have both set up, via testing with `gh` cli. 

Then, to avoid provisioning both enterprise and regular account at the same time, I would create another CredentialType and CredentialUsage and assign it in the plugin and executable respectively so that the user can configure which use case they'd like to use when installing the plugin.

eedgar: GH_TOKEN, GITHUB_TOKEN (in order of precedence): an authentication token for github.com
API requests. Setting this avoids being prompted to authenticate and takes precedence over
previously stored credentials.

GH_ENTERPRISE_TOKEN, GITHUB_ENTERPRISE_TOKEN (in order of precedence): an authentication
token for API requests to GitHub Enterprise. When setting this, also set GH_HOST.

scottisloud:
GH Enterprise support has been added since this PR was posted, so I will be closing it off. Thanks for the contribution!

PR #418, opened 2024-01-05
Create wireguard wg-quick plugin

drzombey:
## Overview

I've created a plugin for wg-quick to use the configurations out of 1password instead of the standard config path.

## Type of change
 

- [x] Created a new plugin

## How To Test
Example command: 

Without plugin: 

`wg quick up wg0`

With plugin:

`wg-quick up`

No Interface name is needed cause the plugin will handle it with the credential

## Changelog
Enable secure authentication for wg-quick using 1Password Shell Plugins, ensuring proper initialization with default credentials via op plugin init and support for importing credentials from the specified standard wireguard config file e.g. "/etc/wireguard/wg0.conf"

### DIFF
--- a/plugins/wireguard/access_config.go
+++ b/plugins/wireguard/access_config.go
@@ -1,0 +1,119 @@
+package wireguard
+
+import (
+	"context"
+	"fmt"
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AccessConfig() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.AccessConfig,
+		DocsURL: sdk.URL("https://www.wireguard.com/quickstart/"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.PrivateKey,
+				MarkdownDescription: "Private Key of your client.",
+				Secret:              true,
+			},
+			{
+				Name:                fieldname.Address,
+				MarkdownDescription: "VPN IP address of your client.",
+				Secret:              false,
+			},
+			{
+				Name:                fieldname.PublicKey,
+				MarkdownDescription: "Public Key of the Peer.",
+				Secret:              false,
+			},
+			{
+				Name:                fieldname.Endpoint,
+				MarkdownDescription: "Endpoint address of the Peer.",
+				Secret:              false,
+			},
+			{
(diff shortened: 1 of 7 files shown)


plugins/wireguard/access_config.go
```diff
@@ -0,0 +1,119 @@
...
+			return
+		}
+
+		fields := make(map[sdk.FieldName]string)
+		for _, section := range configFile.Sections() {
+			if section.HasKey("PrivateKey") && section.Key("PrivateKey").Value() != "" {
+				fields[fieldname.PrivateKey] = section.Key("PrivateKey").Value()
+			}
+			if section.HasKey("Address") && section.Key("Address").Value() != "" {
+				fields[fieldname.Address] = section.Key("Address").Value()
```
hculea, line 102: Which of these fields actually required for the working of the CLI? The provisioner suggests all fields are required, while the importer - that none are.

Let's make sure we add an error if a field that's required is not present.

plugins/wireguard/access_config.go
```diff
@@ -0,0 +1,119 @@
...
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.PrivateKey,
+				MarkdownDescription: "Private Key of your client.",
+				Secret:              true,
+			},
+			{
+				Name:                fieldname.Address,
+				MarkdownDescription: "VPN IP address of your client.",
+				Secret:              false,
```
hculea, line 27: nit: we can omit specifying `Secret: false`, since that's the default iirc.

Let's make sure we add mark the fields that are not required as `Optional`.

hculea (commented):
Thank you for your contribution! ❤️ The code looks great - my main concern is about the UX, based on the test commands outlined in the PR description.

With shell plugins, we aim to make the transition between the manual credential provisioning process to 1Password seamless - the user shouldn't have to change the command they run for that. However, the before/after commands look slightly different - can you elaborate a bit on that, so we can reason on the best way to move forward here?

drzombey:
> Thank you for your contribution! ❤️ The code looks great - my main concern is about the UX, based on the test commands outlined in the PR description.
> 
> With shell plugins, we aim to make the transition between the manual credential provisioning process to 1Password seamless - the user shouldn't have to change the command they run for that. However, the before/after commands look slightly different - can you elaborate a bit on that, so we can reason on the best way to move forward here?

Hey, yes of course I can. You're right the commands are different. Currently the main issue is that the wg0 interface is a hint for the wg-quick cli which config it should use in the /etc/wireguard folder. But you can also add as an argument the config file like the plugin will do it, for this you must replace the interface name with the config path. If we can find a way to replace the interface name during the command execution by the config file path the experience should be the same as before. Or if there is a way to get the correct secret in 1password by the interface name.

plugins/wireguard/access_config.go
drzombey: Will add a error during the import. So all fields are required to connect to the server

PR #371, opened 2023-08-17
Add LocalStack plugin

simonrw:
## Overview

LocalStack is ["a fully functional local cloud stack, to develop and test your cloud and serverless apps offline".](https://localstack.cloud/).

This PR adds support for setting the API key for use with LocalStack's pro offering.

*Disclaimer: I work for LocalStack*

## Type of change
 

- [x] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [ ] Improved contributor utilities or experience

## How To Test

```
# configure your shell for using the plugin
localstack start
# LocalStack reads the API key from 1Password, and enables pro mode
```

PR #428, opened 2024-02-14
Update localstack plugin to use LOCALSTACK_AUTH_TOKEN

simonrw:
## Overview

LocalStack is migrating to use auth tokens instead of api keys. This should be reflected in the 1Password LocalStack shell plugin.

See the first item in the [LocalStack v3.0.0 release notes miscellaneous section](https://blog.localstack.cloud/2023-11-16-announcing-localstack-30-general-availability/#miscellaneous).

Updates the plugin created in https://github.com/1Password/shell-plugins/pull/371

## Type of change
 

- [ ] Created a new plugin

## How To Test

```
localstack start
```

## Changelog
  

The LocalStack plugin now uses `LOCALSTACK_AUTH_TOKEN` instead of `LOCALSTACK_API_KEY`


scottisloud:
Thanks for updating the LocalStack plugin for auth tokens, and thank you for your patience. 

A few changes before merge:

**Env-var mapping**

`AuthToken()` and `APIKey()` currently share `defaultEnvVarMapping` with cross-field entries (`LOCALSTACK_API_KEY` → `fieldname.APIKey`, `LOCALSTACK_AUTH_TOKEN` → `fieldname.AuthToken`). Each credential type should have its own map — `AuthToken` should provision/import only `LOCALSTACK_AUTH_TOKEN`. Please also either remove the unused `APIKey()` path or update `api_key_test.go` if you keep it for backward compat.

**Auth gating**

Please add `needsauth.NotWhenContainsArgs("auth")` in `localstack.go` so setup commands like `localstack auth set-token`, `show-token`, and `clear-token` do not require a 1Password item first.

Optional: add `Prefix: "ls-"` to the token composition, and consider provisioning both `LOCALSTACK_AUTH_TOKEN` and `LOCALSTACK_API_KEY` during the migration window (LocalStack still accepts auth tokens via the legacy env var temporarily).

Happy to take another look once those are in!

simonrw:
Thanks @scottisloud , I am afk at the moment but I'll make these changes in a few weeks

simonrw:
Thanks @scottisloud for the review comments. I have implemented the changes you suggested. We have stopped supporting API keys entirely (they are no longer given out by our authentication system), so I have removed the api key path altogether.

scottisloud:
Hi @simonrw 

I'll merge this shortly. Thanks for the patience here, and for the careful follow-up.

scottisloud (approved):
LGTM! 

Thanks again, and keep your eyes out for this to show up in a future release of the 1Password CLI!

Issue #201, opened 2023-03-13
New plugin: Shopify


williamhpark:
### Platform or tool

Shopify

### Relevant links

* Platform homepage: https://www.shopify.com
* CLI docs: 
  * https://shopify.dev/docs/themes/tools/cli
  * https://shopify.dev/docs/apps/tools/cli

### Goal or desired behavior

Authenticate the Shopify CLI using an Authentication Token so users can authenticate into their Shopify account from the CLI and build apps/themes.

### Contribution

- [ ] I'd be up for building the plugin myself.

### Additional information

There are three main uses for the Shopify CLI, outlined here:
1. [Shopify CLI for apps](https://shopify.dev/docs/apps/tools/cli)
2. [Shopify CLI for themes](https://shopify.dev/docs/themes/tools/cli)
3. [Shopify CLI commands for Hydrogen storefronts](https://shopify.dev/docs/custom-storefronts/hydrogen/cli/commands)

It seems that authentication is only needed for the first two (apps and themes). Shopify uses OAuth to authenticate into your account. However, it seems that they also support authentication tokens. More information can be found at https://shopify.dev/docs/apps/tools/cli/ci-cd and https://shopify.dev/docs/themes/tools/cli/ci-cd for apps and themes respectively.

PR #397, opened 2023-10-29
Adding Shopify CLI plugin

ammiranda:
## Overview

Adding a shell-plugin for the Shopify CLI.

## Type of change
 

- [X] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [ ] Improved contributor utilities or experience

## Related Issue(s)
  

* Resolves: #201

## How To Test

`shopify theme dev`

## Changelog
  
Adds the Shopify CLI plugin.

PR #434, opened 2024-02-28
Added Shopify Themes CLI Plugin (The Second Attempt™️)

jerryzhou196:
## Overview
This PR fully implements the Shopify Theme CLI Shell Plugin. It builds off of the great work done by @ammiranda and @hculea on #397.  

https://github.com/1Password/shell-plugins/assets/46613983/51674685-198d-4acc-ad6f-4ac11aab88cd

The video shows the 1Password Shopify Shell Plugin integrating with the `shopify theme` CLI and a [theme access API key](https://shopify.dev/docs/themes/tools/theme-access). 

Incidentally, there were a few interesting things I stumbled upon. 

1. I couldn't implement the `importer` because the [Shopify CLI stores the environment file by project directory rather than a fixed path. ](https://shopify.dev/docs/themes/tools/cli/environments) 
(instead of something like `~/config.toml`, it's `{PROJECT_DIRECTORY}/shopify.theme.toml`). 

2. I noticed that `importer.NoOp()` would cause the `op plugin init` flow to fail with `[ERROR]... credential not found` for the `Import into 1Password...` option. The same issue was mentioned in #283. Without being able to see the underlying code causing the error, I provide a band-aid solution in this PR that fixes the `importer.NoOp()` implementation to no longer give this error.  

3. Shopify CLI also has a [lone Webhooks CLI integration](https://github.com/Shopify/cli/blob/main/packages/cli/README.md#commands) called `shopify webhook trigger` that requires a `--client-secret` passkey.  I'm very happy to implement this too. What are your thoughts? 

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #201 
* Relates: #397
* Relates: #283

## How To Test
1. [Download the Shopify CLI](https://shopify.dev/docs/themes/tools/cli/install). 
2. Create a Shopify Store on [Shopify](https://www.shopify.com/) and [configure the theme access app to your store](https://shopify.dev/docs/themes/tools/theme-access). 
3. Provision a key through the theme access portal. 
    (Side note: Not sure how to test the automatic chrome extension detection import here)
image
6. Build the plugin and run `op plugin init shopify`. Store the key through the import option. 
7. Try running the protected theme commands without the `--password` option or the Shopify built-in authentication. It should work. 

## Changelog
Added Shopify Themes Plugin. 
Added an empty attempt to importer.NoOp() to correctly function without error.


plugins/shopify/access_key.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func AccessKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:    credname.CLIToken,
+		DocsURL: sdk.URL("https://shopify.dev/docs/themes/tools/cli"),
+		//ManagementURL: sdk.URL("https://admin.shopify.com/store/{YOUR_STORE_ID}/apps/theme-kit-access"), // Can't support unless URL supports variable URLs
```
AndyTitu, line 16: Hmm, I see no reason why you can't just uncomment this part and leave the url as is: `https://admin.shopify.com/store/{YOUR_STORE_ID}/apps/theme-kit-access`. Obviously this isn't clickable but it will direct users towards where they could obtain the auth token, which is the goal of this field.

plugins/shopify/access_key.go
```diff
@@ -0,0 +1,52 @@
...
+					Prefix: "shptka_",
+					Charset: schema.Charset{
+						Lowercase: true,
+						Digits:    true,
+					},
+				},
+			},
+		},
+		Importer: importer.NoOp(),
+		// Can't Implement Shopify environment parsing as the files are stored by project directory instead of a fixed location
```
AndyTitu, line 33: We can move this comment on top of the `Importer:` field to point out that it refers to the importing logic,

AndyTitu (approved):
Looks good to me apart from a couple of minor doc suggestions. Will forward this for a security review.

jerryzhou196:
Yaay - impleented your comments. So sorry about the late reply - midterms are finals are always fun. Also, really excited to join the team this summer, haha. Really admire the open source DevOps + Rust projects you guys are building - so much to learn!

PR #435, opened 2024-03-05
feat: Add a Nix Flake with a devShell, NixOS module, and home-manager module to configure Shell Plugins

mrjones2014:
## Overview

The `flake.nix` added does two things:

1. Adds a `devShell` so Nix users can just run `nix develop` or use `direnv` to have a Go environment set up to build the plugins.
2. Add a NixOS module and a `home-manager` module (users will choose one or the other) to configure 1Password Shell Plugins via Nix; this only works if you're using Nix to manage your shell config, but if you're deep into Nix enough to be using Flakes, you probably are managing your shell config with Nix, too.

With this flake as an input to your own flake, you would then be able to configure 1Password Shell Plugins with the following:

```nix
programs._1password-shell-plugins = {
  enable = true;
  plugins = with pkgs; [ gh awscli2 cachix ];
};
```

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

Create a `home-manager` Nix flake that manages your shell configuration:

`flake.nix`
```nix
{
  description = "My dotfiles managed with nix as a flake";
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
    home-manager = {
      url = "github:nix-community/home-manager";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    # This should be the path to the repo checkout locally, with this branch checked out
    _1password-shell-plugins.url = "git+file:///$HOME/git/shell-plugins";
  };

  outputs = inputs@{ self, nixpkgs, home-manager, ... }: {
    homeConfigurations = {
      # note: this "my-config" name will be the value you use on the command line
      # to apply the configuration as well
      "my-config" = home-manager.lib.homeManagerConfiguration {
        extraSpecialArgs = { inherit inputs; };
        # or pkgs = nixpkgs.legacyPackages.x86_64-linux; if you're on linux
        pkgs = nixpkgs.legacyPackages.aarch64-darwin;
        modules = [
          inputs._1password-shell-plugins.hmModules.default
          ./home.nix
        ];
      };
    };
  };
}
```

`./home.nix`
```nix
{ inputs, config, pkgs, lib, ... }: {
  home = {
    username = "<your-os-username>";
    # or `/home/<your-linux-username>` if you're on Linux
    homeDirectory = "/Users/<your-os-username>";
    # this should be set up by home-manager automatically for you, leave it alone
    stateVersion = "22.11";
  };
  # allow 1Password CLI package; Nix doesn't allow any unfree software by default
  nixpkgs.config.allowUnfreePredicate = pkg:
    builtins.elem (lib.getName pkg) [
      "1password-cli"
    ];
  programs._1password-shell-plugins = {
    enable = true;
     # whatever shell plugins you want to use that have packages
     # available in nixpkgs
    plugins = with pkgs; [ gh ];
  };
  programs.fish = {
    enable = true;
    # the rest of your shell config goes here
  };
}
```

If you're on macOS, you will need to install `home-manager` manually as well: https://nix-community.github.io/home-manager/index.xhtml#sec-install-standalone

On NixOS, you can manage `home-manager` as a NixOS module instead.

For standalone `home-manage` installation, run `home-manager switch --flake ~/path/to/flake-directory/.#my-config` to apply the configuration. If you're on NixOS and using a flake to manage `home-manager` as a NixOS module, you can just run `sudo nixos-rebuild switch --flake ~/path/to/flake-directory/.#my-config`

## Changelog
  
Add a `flake.nix` that provides a `devShell` for contributors as well as a NixOS and `home-manager` module that can be used to configure 1Password Shell Plugins via Nix.


.github/workflows/update-flake-dependencies.yml
```diff
@@ -0,0 +1,30 @@
...
+        COMMIT_BRANCH: automation/update-flake-dependencies
+        COMMIT_MESSAGE: "chore(nix): Update Flake dependencies"
+      run: |
+          # commit via the GitHub API so we get automatic commit signing
+          gh api --method PUT /repos/1Password/shell-plugins/contents/$FILE_TO_COMMIT \
+            --field message="$COMMIT_MESSAGE" \
+            --field content=@<(base64 -i $FILE_TO_COMMIT) \
+            --field branch="$COMMIT_BRANCH" \
+            --field sha="$(git rev-parse $COMMIT_BRANCH:$FILE_TO_COMMIT)"
+          gh pr create --title "[automation]: Update Flake dependencies" --reviewer mrjones2014 --base main --head $COMMIT_BRANCH
```
mrjones2014, line 30: I got this from here: https://gist.github.com/swinton/03e84635b45c78353b1f71e41007fc7c

I'm not 100% sure if this will work, and I'm not sure how to test it other than merging it to `main` and running the job. Let me know if we'd rather punt on this and tackle this in a follow-up PR.

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
+{ pkgs, lib, config, is-home-manager, ... }:
+with lib;
```
mrjones2014, line 2: `with lib;` brings library functions into scope like `map`, `baseNameOf`, `getExe` etc.

They could also be referenced without this line as `lib.map`, `lib.baseNameOf`, `lib.getExe`, for example.

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+let cfg = config.programs._1password-shell-plugins;
```
mrjones2014, line 3: this line lets us reference our own configuration as set by the end user, e.g. it gives us the values the user sets via:

```nix
programs._1password-shell-plugins = {
  # it gives us these values
  enable = true;
  plugins = with pkgs; [ gh awscli2 cachix ];
};
```

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+in {
+  options = {
```
mrjones2014, line 5: the `options` block defines the options our module exposes

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+    programs._1password-shell-plugins = {
+      enable = mkEnableOption "1Password Shell Plugins";
```
mrjones2014, line 7: `mkEnableOption` is a shorthand to just create a `boolean` option with description `"Whether to enable ${opt}"`, where `opt` is the text we're passing here ("1Password Shell Plugins").

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+            cachix
+          ]
+        '';
+        description =
+          "CLI Packages to enable 1Password Shell Plugins for; ensure that a Shell Plugin exists by checking the docs: https://developer.1password.com/docs/cli/shell-plugins/";
+      };
+    };
+  };
+
+  config = let
```
mrjones2014, line 24: the `config` block is the Nix expression that gets generated based on what the user sets for the `options` we defined above.

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+    # Explanation:
+    # Map over `cfg.plugins` (the value of the `plugins` option provided by the user)
+    # and for each package specified, get the executable name, then create a shell alias
+    # of the form:
+    # `alias {pkg}="op plugin run -- {pkg}"`
+    # where `{pkg}` is the executable name of the package
+    aliases = listToAttrs (map (package: {
+      name = package;
+      value = "op plugin run -- ${package}";
+    }) pkg-exe-names);
```
mrjones2014, line 43: This results in an `attrset` in the form:

```nix
aliases = {
  gh = "op plugin run -- gh";
  aws = "op plugin run -- aws";
  cachix = "op plugin run -- cachix";
};
```

Using a Nix `attrset` for the option (see `bash.shellAliases = aliases;` below) allows us to not depend on the shell's specific alias definition syntax, so if we ever decided to support additional shells like `nushell` or something, it would only require an additional attribute below, we wouldn't have to modify the way the aliases are created (since Nix handles actually setting them in the shell config for us).

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+    packages = [ pkgs._1password ] ++ cfg.plugins;
+  in mkIf cfg.enable (mkMerge [
```
mrjones2014, line 45: `mkIf` returns `{}` if `cfg.enable == false`, i.e. set up these options only if the `enable` option is `true`, and `mkMerge` is pretty much just type coercion; this is a common pattern used in nixpkgs and home-manager packages.

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+    ({
+      programs = {
+        bash.shellAliases = aliases;
+        zsh.shellAliases = aliases;
+        fish.shellAliases = aliases;
+      };
+    } // optionalAttrs is-home-manager { home.packages = packages; }
```
mrjones2014, line 52: if it's the `home-manager` module, set `home.packages = packages;`

nix/shell-plugins.nix
```diff
@@ -0,0 +1,57 @@
...
+      // optionalAttrs (!is-home-manager) {
+        environment.systemPackages = packages;
```
mrjones2014, line 54: otherwise, if its the NixOS module, set `environment.systemPackages = packages;`

.github/workflows/update-flake-dependencies.yml
mrjones2014: I may be able to make something like that work. I'm not sure how to downgrade the flake.lock (other than manually getting all the versions and hashes and such) but I think it would also work if we outright delete flake.lock as a test. I'll try this out today.

AndyTitu:
Hmm, I wasn't able to test according to the `How to test` section. @mrjones2014 could you help me walk through the use case as someone who just installed nix on their Mac OS machine?

.github/workflows/update-flake-dependencies.yml
mrjones2014: Tested by deleting `flake.lock`. I've pushed a couple changes and it works now, you can see in #436 that the latest commit on that PR is the automated commit from the bot and it is indeed a singed commit! Yay for automated `flake.lock` updates!

mrjones2014:
> Hmm, I wasn't able to test according to the `How to test` section. @mrjones2014 could you help me walk through the use case as someone who just installed nix on their Mac OS machine?

Sure thing, I'll send you a message on slack 👍

mrjones2014:
Just pushed a cool commit that validates the `plugins` option and gives an error message if the user specified a package that isn't a supported Shell Plugin

CleanShot 2024-03-06 at 11 10 36

AndyTitu:
Functionally tested as a new Nix user on Mac:

1. dev-mode (after I uninstalled go):

```
bash-3.2$ nix develop
(nix:Shell-with-Go-toolchain-env) Andis-MacBook-Pro:shell-plugins andititu$go version
go version go1.21.7 darwin/arm64
(nix:Shell-with-Go-toolchain-env) Andis-MacBook-Pro:shell-plugins andititu$ which go
/nix/store/b9wbj651hcyfm63kkcxykpjcqc0qqwp5-go-1.21.7/bin/go
```

**Question:** could we only have go and goals installed at the level of the shell-plugin directory, and then when I cd out of this directory, I don't have go installed anymore? From my understanding this is how nix is working right?

2. Prod mode: 

After running `home-manager switch --flake .#mac`

```
bash-3.2$ which aws
/Users/andititu/.nix-profile/bin/aws
bash-3.2$ aws organizations list-accounts
###############################################################################
# WARNING: 'aws' is not from the official registry.                           #
# Only proceed if you are the developer of 'aws'.                             #
# Otherwise, delete the file at /Users/andititu/.config/op/plugins/local/aws. #
###############################################################################
[ERROR] 2024/03/07 18:40:58 RequestDelegatedSession: cannot connect to 1Password app, make sure it is running
```

AndyTitu (approved):

mrjones2014:
> **Question:** could we only have go and goals installed at the level of the shell-plugin directory, and then when I cd out of this directory, I don't have go installed anymore? From my understanding this is how nix is working right?

Yep, that's the idea of Nix. I don't have any toolchains installed globally (like `go`, `gopls`, `rustc`, `cargo`, or anything like that); they get installed to temporary environments when I `cd` into a project directory that has a `flake.nix` (because I have `nix-direnv` enabled, it automatically uses the `flake.nix` when I `cd`).

If you use [direnv](https://github.com/direnv/direnv) or [nix-direnv](https://github.com/nix-community/nix-direnv) (more optimized with extra caching for Nix stuff), you can create a `.envrc` file with contents just `use flake`, and then the Nix shell will be automatically set up when you `cd` into the directory, and exited when you `cd` out of the directory. Alternatively you can just manually run `nix develop` which will start a subshell set up by the flake, only this way you will have to manually `exit` the shell to get back to your main shell with the flake stuff gone. I have a component in my terminal prompt that tells me when I'm in a `nix-shell` subshell, and whether its because of a direnv `.envrc`: 

CleanShot 2024-03-07 at 11 48 15

SimonBarendse (approved):
I'm not a Nix user myself, so take this approval with a grain of salt from a functional review perspective.

That said, from a code review perspective using the documentation and this code, I don't see any issues.

.github/workflows/update-flake-dependencies.yml
```diff
@@ -0,0 +1,37 @@
+# CI job to periodically (once a week) update flake.lock
+name: Update flake dependencies
```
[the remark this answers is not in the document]
mrjones2014: I mostly copied it out of my own dotfiles repo 😛 

Just the bit to create the PR is a little different because we require signed commits in this repo.

joneshf:
Just popping in to say thanks for adding this!

After a little bit of figuring out (first time adding a module to Home Manager), it works great! Plugins are much easier to manage, and keep track of now!

Issue #433, opened 2024-02-23
Use shell functions instead of aliases to avoid interfering with shell completion


jimeh:
### op CLI version

2.24.0

### Goal or desired behavior

I want to setup the homebrew shell plugin (as an example), and have shell completion work correctly for `brew`, and also my custom shell alias (`alias br=brew`).

This can be achieved by using a shell function in `~/.config/op/plugins.sh` instead of an alias:

```bash
brew() { op plugin run -- brew "$@"; }
```

### Current behavior

Currently when you run `op plugin init brew`, a shell alias is written to `~/.config/op/plugins.sh`, for example:

```bash
alias brew="op plugin run -- brew"
```

This results in the `brew` command no longer having shell completions available on it, unless `setopt completealiases` is used. But that breaks completion for shorthand aliases like `alias br=brew`, because it will try to find completions for a command named `br`.

Hence the shell function approach is the only one that works correctly in both scenarios.

mrclrchtr:
Thank you so much for sharing the workaround!

mrjones2014:
Note for myself: we'll also need to make this change in `./nix/shell-plugins.nix`. In a Nix environment, the aliases and managed by Nix instead of by the 1Password CLI.

mrjones2014:
I started playing around with this idea and ran into a small snag -- it's easy to do in the Nix Flake way of configuring things, because we can just generate separate versions of the shell code.

However, for users not using Nix and having the CLI manage their `plugins.sh` file, we run into a bit of an issue with Fish shell users, because Fish uses a different function syntax.

Bash/Zsh:
```sh
my-func() {
  echo "stuff"
}
```

Fish:

```fish
function my-func
  echo "stuff
end
```

I'm not sure the best way to approach this with the CLI.

jimeh:
Simplest approach I guess would to also write out a `plugin.fish` file, and just update instructions for fish users to load that file instead.

Another approach could be to have a `plugin.bash` file for bash/zsh and `plugin.fish` for fish. Then have the existing `plugin.sh` load one of the files conditionally based on the value of `$SHELL`. Assuming such a conditional can be written in a bash and fish compatible way.

jimeh:
In the meantime, I've made my own hacky but automated solution in my dotfiles:

https://github.com/jimeh/dotfiles/blob/53052def08e931e10a37b1cc53092281b5731b51/zsh/1password.zsh

[...]

PR #464, opened 2024-05-13
fix(nix): Update flake.nix to configure Shell Plugins as functions vs. aliases

mrjones2014:
## Overview

Users of the Nix Flake will now have Shell Plugins wired up as shell functions instead of shell aliases. This results in a better tab completion experience.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Relates: #433  (Resolves for Nix Flake users but not others)

## How To Test

1. However you consume the flake (e.g. as an input to your own flake) update the URL to `git+file:///path/to/local/shell-plugins` (replacing the file path as appropriate for your own system)˜
2. Rebuild your Nix configuration
3. Restart your shell
4. Check that Shell Plugins still work

## Changelog
  

Update Nix Flake to configure Shell Plugins as shell functions instead of aliases for better tab completion.

### DIFF
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -54,28 +54,55 @@ in {
     pkg-exe-names = map getExeName cfg.plugins;
     # Explanation:
     # Map over `cfg.plugins` (the value of the `plugins` option provided by the user)
-    # and for each package specified, get the executable name, then create a shell alias
+    # and for each package specified, get the executable name, then create a shell function
     # of the form:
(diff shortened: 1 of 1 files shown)


SimonBarendse (approved):
Thanks for putting this together and removing this annoying friction point from the experience! 🙌 

Kudos to @jimeh for inspiring this work in https://github.com/1Password/shell-plugins/issues/433 🙌  ❤️

mrjones2014:
@edif2008 can we get this one approved and merged as well?

Issue #467, opened 2024-05-23
CI: `flake.lock` update job should update the existing PR if an unmerged one exists


mrjones2014:
That is, if we haven't merged a `flake.lock` PR by the time the job runs again, the job should push to the existing branch, updating the unmerged PR with the latest `flake.lock` changes.

Currently it fails because the branch already exists; instead of failing, it should push to the existing branch.

PR #468, opened 2024-05-24
fix(ci): Update existing flake.lock PR if it exists already

mrjones2014:
## Overview

If a `flake.lock` automated PR already exists and hasn't been merged by the time a new job runs to create a new PR, it updates the existing PR instead of trying to create a new PR and failing because the branch already exists.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #467 

## How To Test

Read the changed bash script and ensure it makes sense. You can also test out the logic locally on a test repo. The thinking in the first few lines (the first `if` block) is:

- If the branch already exists on remote, checkout and pull before trying to update `flake.lock` (this is to ensure we can push new changes, e.g. a commit exists on remote that we don't have locally)
- Otherwise, create the branch and push the empty branch to remote
- _Then_ run `nix flake update`, which now runs on the latest version of the file from the branch
- Push the changes to the branch
    - If the branch already existed and had a PR for it, pushing new commits to the branch will update the existing PR

## Changelog
  

Update `flake.lock` CI automation to update existing PR if there is one.

### DIFF
--- a/.github/workflows/update-flake-dependencies.yml
+++ b/.github/workflows/update-flake-dependencies.yml
@@ -13,20 +13,28 @@ jobs:
     - uses: actions/checkout@v4
     - uses: DeterminateSystems/nix-installer-action@main
     - uses: DeterminateSystems/magic-nix-cache-action@main
-    - name: update flake.lock
-      run: nix flake update
-    - name: Create signed commit with flake.lock changes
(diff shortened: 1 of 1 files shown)


edif2008 (approved):
Thank you for your contribution! 😄 

This definitely looks like a great improvement that we should have to keep things nice and clean.

PR #471, opened 2024-06-14
Optional package

Aigeruth:
## Overview

1Password CLI requires setting the group of the executable to `onepassword-cli`. This isn't possible when home-manager is installed in standalone mode (e.g. Ubuntu or other distributions). Opting out of installing the 1Password CLI
with nix allows using the system-wide installation (e.g. via apt).

This change also allows changing the package. It's beneficial when nixpkgs-unstable is mapped to `pkgs.unstable` via an overlay.

This is **not** breaking, the current behavior remains as is.

## Type of change
Improved experience for nix users.

## How To Test
Opting out of installing the package:
```nix
{
    imports = [ inputs._1password-shell-plugins.hmModules.default ];
    programs._1password-shell-plugins = {
      enable = true;
      package = null;
      plugins = with pkgs; [ gh awscli2 cachix ];
    };
}
```

Changing the package:
```nix
{
    imports = [ inputs._1password-shell-plugins.hmModules.default ];
    programs._1password-shell-plugins = {
      enable = true;
      package = pkgs.unstable._1password-cli;
      plugins = with pkgs; [ gh awscli2 cachix ];
    };
}
```

## Changelog

Allow nix users to change the `_1password-cli` package like in home-manager modules or allow opting out from installing the package by setting it to `null`.

### DIFF
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -21,6 +21,7 @@ in {
   options = {
     programs._1password-shell-plugins = {
       enable = mkEnableOption "1Password Shell Plugins";
+      package = mkPackageOption pkgs "_1password" { };
       plugins = mkOption {
         type = types.listOf types.package;
         default = [ ];
@@ -62,7 +63,7 @@ in {
       name = package;
       value = "op plugin run -- ${package}";
     }) pkg-exe-names);
-    packages = [ pkgs._1password ] ++ cfg.plugins;
+    packages = [ cfg.package ] ++ cfg.plugins;
   in mkIf cfg.enable (mkMerge [
     ({
       programs = {
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -21,7 +21,7 @@ in {
   options = {
     programs._1password-shell-plugins = {
       enable = mkEnableOption "1Password Shell Plugins";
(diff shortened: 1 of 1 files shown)


shymega:
I would love to see this merged. Mostly because of standalone home-manager.

SimonBarendse (approved):
> This is not breaking, the current behavior remains as is.

Excellent! Definitely appreciate you thinking about this. 💙

This all looks sensible to me. @hculea or @mrjones2014 could you give this a look for a second pair of eyes?

SimonBarendse:
How does this compare to alternative implementation in https://github.com/1Password/shell-plugins/pull/503 ?

mrjones2014:
#503 is technically solving a different issue. The upstream `nixpkgs` changed the package name from `_1password` to `_1password-cli` to closer match our 1Password nomenclature, so evaluation prints a warning about that.

However, this PR also changes the default package from `_1password` to `_1password-cli` which also fixes that issue.

Basically, this PR supersedes #503 (in a slightly more Nix-convetional way as well) and also fixes an additional problem with the home-manager integration.

So we can merge this PR and close #503.

mrjones2014 (approved):

mrjones2014:
The changes look good to me but it looks like we need another approval from someone with write access. @hculea ?

PR #474, opened 2024-07-10
[BUG FIX] Remove ngrok version check everytime a op plugin command is called

MOmarMiraj:
## Overview

This MR will fix a bug regarding `ngrok --version` being called on every `op plugin` command.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

The test pipelines should pass as usual and when running an `op plugin` command, if you check the activity monitor.. there should be no `ngrok` processes. (Note: Must have the ngrok CLI installed to test this)

Sometimes its hard to see the `ngrok` process on the activity monitor/task manager. An easier way to test is by changing L129 on `plugins/ngrok/provisioner.go` to open up any application you have on your PC and see if the app opens up when you run a `op plugin` command.

## Changelog
  
The op plugin commands do not call a ngrok version check even if the ngrok plugin is not initialized.


plugins/ngrok/provisioner.go
```diff
@@ -138,9 +146,9 @@ func getNgrokVersion() (string, error) {
 }
 
 func (f fileProvisioner) Deprovision(ctx context.Context, in sdk.DeprovisionInput, out *sdk.DeprovisionOutput) {
-	// nothing to do here: files get deleted automatically by 1Password CLI
+	// nothing to do here: files get deleted automatically by 1Password CLI and environment variables get wiped when process exits
 }
 
 func (f fileProvisioner) Description() string {
-	return "Config file aware provisioner. It will first check if an already existing config file is present."
+	return "If ngrok version is 3.2.1 or higher than provision ngrok credentials as environment variables NGROK_AUTH_TOKEN and NGROK_API_KEY otherwise config file aware provisioner. It will first check if an already existing config file is present."
```
AndyTitu, line 153:
```suggestion
	return "If ngrok version is 3.2.1 or higher then provision ngrok credentials as the environment variables NGROK_AUTH_TOKEN and NGROK_API_KEY, otherwise use config file aware provisioner. It will first check if an already existing config file is present."
```

AndyTitu (approved):
lgtm apart from the wording suggestion

plugins/ngrok/provisioner.go
```diff
@@ -28,12 +28,12 @@ const (
 type fileProvisioner struct {
 }
 
-func ngrokProvisioner() sdk.Provisioner {
+func IsNgrokAPIKeySupported() bool {
```
hculea, line 31: Let's call this function something clearer, such as `HasEnvVarSupport` or something like that.

plugins/ngrok/provisioner.go
```diff
@@ -42,14 +42,22 @@ func ngrokProvisioner() sdk.Provisioner {
...
 
 	// Otherwise use config file to provision credentials
-	return fileProvisioner{}
+	return false
 }
 
 func (f fileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
+	ngrokProvisioner := IsNgrokAPIKeySupported()
+
+	if ngrokProvisioner {
```
hculea, line 55: nit - we can call this function directly in the if statement, to avoid initializing a variable that's only used once.

plugins/ngrok/ngrok.go
```diff
@@ -23,8 +23,7 @@ func ngrokCLI() schema.Executable {
 		),
 		Uses: []schema.CredentialUsage{
 			{
-				Name:        credname.Credentials,
-				Provisioner: ngrokProvisioner(),
```
hculea, line 27: What is the reason and effect of removing this overridden provisioner here?

MOmarMiraj: When shell plugins get initialized, all the plugins New() function is called which calls on the ngrokCLI() which calls on Uses and in there, there is a optional provisioner field and currently there is a function that checks whether the version is greater than 3.2.1 and depending if its < or >=, then the type of provisioner is used (either file or using ENV variables).

So by doing any `op plugin` command, it will call `ngrok version`

plugins/ngrok/provisioner.go
```diff
@@ -42,14 +42,22 @@ func ngrokProvisioner() sdk.Provisioner {
...
-		return ngrokEnvVarProvisioner{}
+		return true
 	}
 
 	// Otherwise use config file to provision credentials
-	return fileProvisioner{}
+	return false
 }
 
 func (f fileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
```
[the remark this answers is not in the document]
MOmarMiraj: Done

plugins/ngrok/provisioner.go
MOmarMiraj: Done!

plugins/ngrok/provisioner.go
MOmarMiraj: Done never knew you could do this.. thanks!

plugins/ngrok/provisioner.go
```diff
@@ -42,14 +42,21 @@ func ngrokProvisioner() sdk.Provisioner {
...
 
 	// Otherwise use config file to provision credentials
-	return fileProvisioner{}
+	return false
 }
 
-func (f fileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
+func (p ngrokProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, out *sdk.ProvisionOutput) {
+
+	if ngrokProvisioner := HasEnvVarSupport(); ngrokProvisioner {
```
hculea, line 54: can't this simply be
```
if HasEnvVarSupport() 
```
?

hculea (approved):
Code looks good, and the changes make sense - calling the `ngrok --version` command only when actually provisioning and not in the Provisioner's instantiation.

Haven't functionally tested.

plugins/ngrok/provisioner.go
MOmarMiraj: Yes, Done!

PR #475, opened 2024-07-21
Add support for LLM CLI utility

dasevilla:
## Overview

Add a shell plugin to support [LLM](https://llm.datasette.io/):

> A CLI utility and Python library for interacting with Large Language Models, both via remote APIs and models that can be installed and run on your own machine.

Depending on your usage, LLM needs access to several API keys that should be best kept in 1Password. This plugin supports importing from the `keys.json` file and provisions keys using environment variables. Instead of a 1Password item per service, this uses a single entry with a field per LLM API.

Example entry:

2024-07-21-screenshot-nZHgduWx

## Type of change

- [x] Created a new plugin

## How To Test

Once you've installed the plugin, you can import your existing `keys.json` using `op plugins init llm` and follow the prompts.

Once setup, you can invoke `llm`:

```shell
llm "Why should developers use the 1Password cli?"
llm -m claude-3-haiku "Pitch my boss on adopting 1Password"
```

## Changelog

Add shell plugin for LLM CLI utility.

### DIFF
--- a/plugins/llm/api_key.go
+++ b/plugins/llm/api_key.go
@@ -1,0 +1,242 @@
+package llm
+
+import (
+	"context"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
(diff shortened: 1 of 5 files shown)


dasevilla:
Hmm, based on the `CONTRIBUTING.md`, this might be out of scope:

> CLIs that authenticate to multiple different platforms depending on the project are also out of scope at this stage

If so, feel free to close and I'll continue to use it locally.

Issue #496, opened 2024-11-02
Nix warning - package rename


JeffersonBledsoe:
`evaluation warning: _1password has been renamed to _1password-cli to better follow upstream name usage`:

https://github.com/1Password/shell-plugins/blob/36e4bf4d5b177dd059cafdafaed34bd3264e1e2e/nix/shell-plugins.nix#L65

reonokiy:
and here:
https://github.com/1Password/shell-plugins/blob/f03bbc885daf60554ed7b0d011efe0f3e1e48c7e/nix/shell-plugins.nix#L11

scottisloud:
This should be resolved by changes introduced in commit 4edb39e. If you still see this, feel free to re-open this issue with new reproduction steps.

PR #503, opened 2025-01-06
Make 1Password CLI package configurable and fix `pkgs._1password` warning

LukeChannings:
The `_1password` package was renamed to `_1password-cli` in nixpkgs.

Using the `_1password` package prints the following:

```
evaluation warning: _1password has been renamed to _1password-cli to better follow upstream name usage
```

This PR is similar to #498, but they missed the invocation on line 11 so it still prints the warning.

## Changes

- Add a `cliPackage` option, which defaults to `_1password-cli`. Users may want to use a custom version of the 1Password CLI without using an overlay - this provides that option.
- The evaluation message is no longer printed

### DIFF
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -8,7 +8,7 @@ let
       pkgs.runCommand "op-plugin-list" { }
       # 1Password CLI tries to create the config directory automatically, so set a temp XDG_CONFIG_HOME
       # since we don't actually need it for this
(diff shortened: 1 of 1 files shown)


dawidd6 (approved):

bjeanes:
Closes #496 #498

dudeofawesome (approved):

anaclumos:
Would love to hear if there are any updates on this!

RyanPrussin:
I just made #528, but realized after the fact that this exists and is probably a better and more flexible implementation.

This update has even been approved by multiple reviewers -- what is holding it back?

Issue #527, opened 2025-07-24
Remove


RyanPrussin:
### op CLI version

[Nix Support] Deprecated `pkgs._1password` References should be Updated to `pkgs._1password-cli`

### Goal or desired behavior

All references to `pkgs._1password` should be updated to `pkgs._1password-cli` to remove the warning from code that references this flake.

### Current behavior

When importing the `1Password/shell-plugins` from another flake to use in `home-manager`, on running `nix flake check` with the nix CLI tool, the following evaluation warning occurs:

```
evaluation warning: _1password has been renamed to _1password-cli to better follow upstream name usage
```

This relates to [this line](https://github.com/NixOS/nixpkgs/blame/c62f9b26b01b0c1fa162fa81d473e568be3f67a7/pkgs/top-level/aliases.nix#L64) in nixpkgs that was added close to a year ago.

### Relevant log output

```shell

```

PR #528, opened 2025-07-24
Remove warning when importing the `1Password/shell-plugins` module in a nix flake

RyanPrussin:
## Overview
Remove warning when importing the `1Password/shell-plugins` module in a nix flake.
```SHELL
evaluation warning: _1password has been renamed to _1password-cli to better follow upstream name usage
```

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
#527 

* Resolves: 
  - #527
  - #496
  - #498

* Relates: 
  - #527
  - #503 (this is probably a superior fix, but it is seemingly being ignored...?)

## How To Test
- Manual Tests: in order to replicate the warning and confirm that it is fixed, you would have to import the package into a separate flake using home-manager, and then run `nix flake check` in that flake. Unfortunately I think you have no choice but to test it this way right now.
- Automated Tests: you probably should be able to `nix flake check` in order to see these kind of warnings in your library directly, however this does not work with the current `flake.nix` configuration you have, and it would require more significant updates to your Nix code in order to support. Perhaps something worth considering.

## Changelog
Nix Support: Remove evaluation warning when importing this module in a nix flake.

### DIFF
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -8,7 +8,7 @@ let
       pkgs.runCommand "op-plugin-list" { }
       # 1Password CLI tries to create the config directory automatically, so set a temp XDG_CONFIG_HOME
       # since we don't actually need it for this
-      "mkdir $out && XDG_CONFIG_HOME=$out ${pkgs._1password}/bin/op plugin list | cut -d ' ' -f1 | tail -n +2 > $out/plugins.txt"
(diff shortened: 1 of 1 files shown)


scottisloud:
Thanks for putting this together, and sorry for the long wait on review.

Good catch. The rename `pkgs._1password` to `pkgs._1password-cli` looks right and #498 missed. On current nixpkgs, `_1password` is no longer just a warning; it can hard-fail evaluation, so this is worth landing.

Before merge, could you rebase onto `main`? GitHub currently reports merge conflicts on this branch.

If you'd like to go further, #503's optional `cliPackage` setting is a nice way to let consumers pin a custom CLI package, but the two-line rename here is enough to fix the warning/throw on its own.

Happy to take another look once you've rebased.

PR #531, opened 2025-07-29
Add output to commands that need authentication in Terraform

fearoffish:
## Overview
Terraform needs authentication for output, this adds it.

## Type of change
- [ ] Created a new plugin

## How To Test
`terraform output`

## Changelog
The terraform plugin now authenticates when running `terraform output`


fearoffish:
I'm going to update this with signed credentials, but I don't see the point to be honest, because you clearly couldn't give a crap about shell-plugins! The last update was 7 months ago and it was a simple dependabot one. If you aren't going to support something, maybe stop selling it in your docs.

PR #465, opened 2024-05-17
fix(ci): Use 1Password-bot token for flake update PRs so approval job…

mrjones2014:
… runs

## Overview

Changes the token used for the flake update CI jobs to use the `1Password-bot` users token, so that the approval job runs and `github-actions[bot]` can apply an approval to the flake.lock update PRs. 

## Type of change
 

- [ ] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [ ] Improved contributor utilities or experience

## Related Issue(s)
  

* Resolves: #
* Relates: #459  #456 

## How To Test

Merge this change, then run a flake.lock update CI job. The PR should be created by 1Password-bot, and github-actions[bot] should provide one approval for it.

## Changelog
  

Use 1Password-bot instead of github-actions[bot] to run the flake.lock update CI jobs.

PR #533, opened 2025-08-02
[Nix] Update flake.lock (it is very out of date)

RyanPrussin:
## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1710146030,
-        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
         "type": "github"
       },
       "original": {
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1716358718,
-        "narHash": "sha256-NQbegJb2ZZnAqp2EJhWwTf6DrZXSpA6xZCEq+RGV1r0=",
+        "lastModified": 1754122398,
+        "narHash": "sha256-CyIrZ68gVLSZGa2ERK7zc2dGabvWp2A8PM6PII9M7gI=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "3f316d2a50699a78afe5e77ca486ad553169061e",
+        "rev": "c02d05bcf73fb496c604798c2268ed424a09e73e",
         "type": "github"
       },
       "original": {


.github/workflows/check-flake.yml
```diff
@@ -1,6 +1,6 @@
 name: Check Nix flake
 on:
-  pull_request_target:
+  pull_request:
```
RyanPrussin, line 3: The purpose of this change is to fix the automated Nix CI workflow.

**The Problem**: The old workflow used `pull_request_target`, which caused it to run checks against the outdated `flake.lock` file from the `main` branch, leading to consistent failures.

**The Solution**: This change switches the workflow to the `pull_request` trigger, which correctly runs checks against the code in this pull request's branch. This is the standard practice for CI checks.

**Why It's Still Failing**: Because the workflow specifications are read from the `main` branch, the new `pull_request` trigger will not take effect until this change is merged. The checks will continue to fail until then.

.github/workflows/update-flake-dependencies.yml
```diff
@@ -15,11 +15,14 @@ jobs:
...
       run: nix flake update
+    - name: Define Unique COMMIT_BRANCH by Timestamp
+      run: |
+        TIMESTAMP=$(date +'%Y-%m-%d-%H%M%S')
+        echo "COMMIT_BRANCH=automation/update-flake-dependencies-$TIMESTAMP" >> $GITHUB_ENV
     - name: Create signed commit with flake.lock changes
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         FILE_TO_COMMIT: flake.lock
-        COMMIT_BRANCH: automation/update-flake-dependencies
```
RyanPrussin, line 22: The purpose of this change is to fix the broken weekly [Update flake dependencies](https://github.com/1Password/shell-plugins/actions/workflows/update-flake-dependencies.yml) GitHub workflow.

### Problem
Once a week and on manual workflow executions, the same branch name `automation/update-flake-dependencies` is being created even though it is never being updated or cleaned up, which causes it to fail on on each subsequent run of this GitHub Action.
image
See full error log from screenshot [here](https://github.com/1Password/shell-plugins/actions/runs/16679744505/job/47215244470).

### Solution
Add a timestamp to the end of the branch name so it is unique.

flake.lock
RyanPrussin, line 1: Normally I would not have updated the lock file in the same PR as CI changes, but your entire CI pipeline for Nix is not working as expected and that is partially because it is dependent on an up-to-date `flake.lock` file in order to pass checks. See **The Problem** in [this comment](https://github.com/1Password/shell-plugins/pull/533/files#r2250031240)) for more details.

mrjones2014:
Hi, thanks for your contribution!

I've previously attempted to solve part of this problem by making it [update the existing PR if one exists](https://github.com/1Password/shell-plugins/pull/468).

I've also attempted to [reduce the approval burden](https://github.com/1Password/shell-plugins/pull/465) by automating one of two required approvals.

Unfortunately, this stuff has just not merged because the repo policy requires two approvals, including one from a user with write access to the repo (which I don't have).

As I've also been unable to get any of the Nix related PRs merged, I've unfortunately resorted to [just vendoring the Nix 1Password Shell Plugins functionality into my own dotfiles](https://github.com/mrjones2014/dotfiles/blob/0d5742af43d829e12234c76da3c00d028ab01150/home-manager/modules/_1password-shell-plugins.nix).

.github/workflows/update-flake-dependencies.yml
RyanPrussin: Note: the reason I made a separate branch was under the assumption that if a new branch and PR is created every week, it will notify the 2nd reviewer more frequently and may result in more of the PRs getting the 2 approvals they need.

mrjones2014:
Hi @RyanPrussin thanks for your contributions! We've kickstarted Nix maintenance again internally, partly thanks to your PR!

This PR has been superseded now by the combination of:

- #469 
- #542 
- #541 
- #545

Closing this one as the issues are now resolved by the above PRs, thanks again!

RyanPrussin:
Glad to hear it, @mrjones2014 !

I did just want to mention that you should probably still make these changes:
- Replace `pull_request_target` trigger with `pull_request`, as per [this comment](https://github.com/1Password/shell-plugins/pull/533/files#r2250031240)
  - This makes sure `nix flake check` is run on the most recent commit in this PR rather than on the base of the PR 
  - GitHub docs state:  `Avoid using [the pull_request_target] event if you need to build or run code from the pull request.` (see [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target))
  - Otherwise, the `nix flake check` isn't really testing the new changes on a PR, but rather the current state of main at the time this branch was created
  - Without this, people could break things in their PRs and you won't find out until after it is merged into main
- Update `*.nix` to `**.nix`, as per [this comment](https://github.com/1Password/shell-plugins/pull/533/files#r2250041456)
  - This ensures that `nix flake check` is triggered by changes to `.nix` files anywhere in the project
  - Currently, changes to any of the `.nix` files in the `nix/` directory will not trigger `nix flake check` to run, which is why you can see no automated checks were run in your PRs [here](https://github.com/1Password/shell-plugins/pull/541) and [here](https://github.com/1Password/shell-plugins/pull/545)

mrjones2014:
Gotcha, if you submit a PR with those changes I'm happy to review!

Issue #544, opened 2025-08-26
Issue in shell-plugins broken with fish


BJSummerfield:
A recent change has made it so nixos will not compile.

The option `home-manager.users.{user}.fish' does not exist. Definition values:
       - In `/nix/store/cfm118zfdjnhjinsr4djqjjq3ir627sd-source/modules/apps/_1password.nix':
           {
             _type = "if";
             condition = true;
             content = {
               interactiveShellInit = ''

If i disable shell plugins I can compile.  The only plugin I use is github.

I'm very tired but i think it should be grouped in with programs similar to bash or zsh

https://nixos.wiki/wiki/Fish

oscarvarto:
I'm affected by this problem as well.

PR #545, opened 2025-08-26
Pass args to plugin shell functions

BJSummerfield:
## Overview

Moved Fish to programs.fish -> Nixos would not compile as user.fish does not exist

added arguments to the new functions.  Shell-plugin args were not passed to OP.  I only tested fish, however the issue was present in bash and zsh.

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #544 
* Relates: #

## How To Test

Code compiles now when using fish shell.  Arguments are now passed to plugins through `op`

### DIFF
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -1,10 +1,4 @@
-{
-  pkgs,
-  lib,
-  config,
-  is-home-manager,
-  ...
-}:
+{ pkgs, lib, config, is-home-manager, ... }:
 with lib;
 let
   cfg = config.programs._1password-shell-plugins;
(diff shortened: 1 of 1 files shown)


mrjones2014 (approved):

BJSummerfield:
Fixed merge conflicts

nix/shell-plugins.nix
```diff
@@ -60,75 +60,64 @@ in
...
-        end
-      '') pkg-exe-names;
-      packages = lib.optional (cfg.package != null) cfg.package ++ cfg.plugins;
-    in
-    mkIf cfg.enable (mkMerge [
-      (
-        {
-          # for Fish its the same option path between NixOS vs. home-manager
-          fish.interactiveShellInit = strings.concatStringsSep "\n" fishFunctions;
-        }
```
dudeofawesome, line 104: This applied to both user & host configs, so removing it will require adding this config down on line 121 as well.

I think the better solution would just be updating the key to `programs.fish.interactiveShellInit` though.

BJSummerfield: I tried to change the key to `programs.fish.interactiveShellInit` where it was initially and the functions were not added to my fish config.

Adding it `is-home-manager` does add the functions to my fish config.

Updated the pr to include `fishFunctions` to host configs

BJSummerfield: After some reading on the // operator (https://nix.dev/manual/nix/2.28/language/operators#update). The right side was overriding the left.  `programs.fish` was being left out after evaluation.

The `mkMerge` only had one item in it. A single config derived from the // operators. https://nixos.org/manual/nixos/stable/#sec-option-definitions-merging

The most recent update removes the // (update) operators and creates 3 items in the mkMerge.  The fish config, the conditional home manager config and the conditional nixos config.  This correctly leverages `mkMerge` for this use case.

Marton6 (approved):

PR #552, opened 2025-08-31
[Nix] CI Fixes for `nix flake check`

RyanPrussin:
## Overview
There are two changes here:

1. Ensure CI for `nix flake check` is triggered by changes to any .nix file 
1. Ensure that CI for `nix flake check` runs against the most recent commits to the current branch (currently runs against the base branch used to create the PR)

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
* Relates: #533

## How To Test
I don't have the permissions to test this myself, but theoretically a maintainer can test this by doing the following:
1. Create a new branch, and do something that intentionally causes `nix flake check` to fail locally in the file `nix/shell-plugins.nix`
1. Push your changes, and notice that the CI workflow checks for `nix flake check` does not trigger (you can check this [here](https://github.com/1Password/shell-plugins/actions/workflows/check-flake.yml))
1. Push the line change that include `'**.nix'` from this PR, and observe [here](https://github.com/1Password/shell-plugins/actions/workflows/check-flake.yml)) again that now there should be a successful run of `nix flake check` (even though we are expecting an error. This occurs because they are essentially run against `main` rather than the current branch's most recent commits).
1. Push the other line change from this PR (the line that contains `pull_request`), and watch as the checks now should run against the most recent commit on this branch. We now expect them to fail `nix flake check`, like it does locally.
1. Undo the intentional bug introduced in step 1 causing `nix flake check` to fail, and watch as the checks should run successfully.

Note: the above test may not work, as sometimes GitHub PRs won't allow the actions to be updated until after the PR is merged, so if there are any issues this should be considered as well.

A maintainer can run this action manually in order for the checks on this PR to pass: https://github.com/1Password/shell-plugins/actions/runs/17350571621 (or see "workflow awaiting approval" at the bottom of this PR description page where the CI checks are shown).

## Changelog
GitHub CI for `nix flake check` now triggers for any `.nix` file, not just the ones in the project root. Additionally, it now tests the most recent commits on the current branch rather than the branch the PR is based from.

### DIFF
--- a/.github/workflows/check-flake.yml
+++ b/.github/workflows/check-flake.yml
@@ -1,8 +1,8 @@
 name: Check Nix flake
 on:
-  pull_request_target:
+  pull_request:
     paths:
-      - '*.nix'
+      - '**.nix'
       - flake.lock
   push:
     branches:
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -47,7 +47,7 @@ in
           map
             (
               package:
-              if (elem (getExeNameasdasdas package) supported_plugins) then
+              if (elem (getExeName package) supported_plugins) then
                 package
               else
                 abort "${getExeName package} is not a valid 1Password Shell Plugin. A list of supported plugins can be found by running `op plugin list` or at: https://developer.1password.com/docs/cli/shell-plugins/"
--- a/nix/shell-plugins.nix
+++ b/nix/shell-plugins.nix
@@ -47,7 +47,7 @@ in
           map
             (
               package:
-              if (elem (getExeName package) supported_plugins) then
+              if (elem (getExeNameasdasdas package) supported_plugins) then
                 package
               else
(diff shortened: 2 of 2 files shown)


RyanPrussin:
I made this Nix fix PR for you @mrjones2014, as discussed [here](https://github.com/1Password/shell-plugins/pull/533#issuecomment-3239410951)!

Thanks for all your efforts recently! 🙏🏻

mrjones2014:
Just to add a little extra context for whoever else reviews this to provide 2nd approval, I used `pull_request_target` originally because of something around the efforts we made to try to get GitHub Actions bot to provide one of the approvals on the flake.lock PRs.

Reason being that we were using a token for that and `pull_request_target` only allows the job to run on the target (this repo) whereas `pull_request` allows the job to run in the fork where the PR originates, if I understand the docs correctly. 

But since we recently learned that not approvals actually unfortunately don't count towards branch protection rules anyway and that job was deleted, and **there are no tokens in the jobs anymore**, `pull_request` is the right thing to use here.

mrjones2014 (approved):

mrjones2014:
Maybe @volodymyrZotov can take a 2nd look

volodymyrZotov (approved):
Looks good!

PR #562, opened 2025-11-27
Add Doppler CLI shell plugin

CodeByZach:
## Overview

This PR adds a shell plugin for the Doppler CLI, enabling secure authentication using 1Password and biometric unlock.

The plugin allows Doppler CLI users to:
- Store a Doppler token (a Personal Token `dp.pt.` or CLI token `dp.ct.`) securely in 1Password
- Authenticate CLI commands with Touch ID/Face ID and other 1Password unlock methods
- Import existing tokens from the `DOPPLER_TOKEN` environment variable and the Doppler CLI config file (`~/.doppler/.doppler.yaml`)
- Provision the credential automatically as `DOPPLER_TOKEN`

## Type of change

- [x] Created a new plugin

## Related Issue(s)

N/A

## How To Test

1. Build and install the plugin:
  ```bash
  make doppler/build
  ```
2. Initialize the plugin:
  ```bash
  op plugin init doppler
  ```
  You'll be prompted to import a token the plugin detects, or to paste a Doppler token (Personal or CLI) and save it to 1Password.
3. Confirm authentication works (you should get a 1Password/biometric prompt):
  ```bash
  doppler me
  ```
4. Verify it works for other commands too:
  ```bash
  doppler projects list
  ```

## Changelog

Authenticate the Doppler CLI using Touch ID and other unlock options with 1Password Shell Plugins.


CodeByZach:
Friendly follow up
@edif2008
@Marton6
@volodymyrZotov
@SimonBarendse

CodeByZach:
Friendly follow up
@edif2008
@Marton6
@volodymyrZotov
@SimonBarendse

zachary-onix:
Kindly following up
@AndyTitu
@edif2008
@volodymyrZotov

zachary-onix:
@scottisloud

scottisloud (changes requested):
Thanks for putting this together — the overall plugin structure looks right, and using `DOPPLER_TOKEN` matches the Doppler CLI docs.

A few things I'd like to see before we merge:

- **Remove `plugins/plugins.go` from the PR.** That file is generated by `make registry` and is gitignored in this repo. Adding it will conflict with other plugin changes; CI/maintainers regenerate it when needed.
- **Align naming with the token type you're modeling.** The file is `service_token.go` and the constructor is `ServiceToken()`, but the credential uses `credname.PersonalAccessToken`, the description says "Personal Token", and the prefix is `dp.pt.` (personal). Please rename to match (e.g. `personal_access_token.go` / `PersonalAccessToken()`) or split credentials if you intend to support both.
- **Fix the management URL.** `https://dashboard.doppler.com/workplace/<workplace-id>/tokens/personal` won't work as a link — use a real dashboard or docs URL without placeholders.
- **Service tokens (`dp.st.`).** Doppler documents service tokens as the common choice for non-interactive use via `DOPPLER_TOKEN` ([token formats](https://docs.doppler.com/reference/auth-token-formats)). Right now the schema only accepts `dp.pt.`. Could you add support for service tokens (second credential or broader composition) or call out clearly if this plugin is personal-token-only?

**nit:** consider adding a fixed length hint (40–44 chars after the prefix) per Doppler's format docs, and a config-file importer for `~/.doppler` if you want smoother imports after `doppler login`.

Happy to take another look once these things are addressed! :)

CodeByZach:
Thanks for the review! Addressed everything:

- Removed the generated `plugins/plugins.go` (gitignored `make registry` output).
- Renamed to `PersonalAccessToken` (`personal_access_token.go`), covering both Personal (`dp.pt.`) and CLI (`dp.ct.`) tokens, which is the developer's account credential. Matches how
`gh`/`glab` model their `auth login` CLIs.
- Fixed the management URL to the dashboard root (the personal-tokens page is workplace-slug-gated, so it can't be a static link).
- Added the `~/.doppler/.doppler.yaml` importer, filtered to `dp.pt.`/`dp.ct.`.
- Scoped to personal/CLI tokens rather than service tokens (local-dev use case; a second credential type isn't supported by the schema). Happy to revisit.

CodeByZach:
Kindly bumping @scottisloud

plugins/doppler/doppler.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func DopplerCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Doppler CLI",
+		Runs:    []string{"doppler"},
+		DocsURL: sdk.URL("https://docs.doppler.com/docs/cli"),
+		NeedsAuth: needsauth.IfAll(
```
scottisloud, line 15: Not sure if your previous comment addresses this but my reading of the logic here suggests that `doppler login` would require doppler creds. 

If the expectation is that a user should be able to run `doppler login` without already-configured plugin, then it's best to add `needsauth.NotWhenContainsArgs("login")` so users can complete that flow. 

Implementing that exclusion, or clarifying that this is a deliberate limiting of scope for the plugin, would tip this over to merge ready!

Thanks for your quick turnaround on the revisions!

scottisloud (changes requested):

plugins/doppler/doppler.go
CodeByZach: I went back and forth on this, and the implication is to only support doppler personal tokens. Doppler login is used exclusively to generate tokens for it's first-party cli.

scottisloud: Okay perfect, so in our release notes we'll just make sure to call out the scope along these lines:
`Authenticate to Doppler CLI with Touch ID and other unlock options with 1Password Shell Plugins (currently only Personal and CLI tokens supported).` or something to that effect. 

I'll give it another once-over and merge.

scottisloud (approved):
LGTM.

scottisloud:
This will appear (along with many other recently-contributed plugins) in a future release of the 1Password CLI.

zachary-onix:
@scottisloud I couldn't seem to find it in the latest release. Is it still queued?

scottisloud:
> @scottisloud I couldn't seem to find it in the latest release. Is it still queued?

Hi @zachary-onix this is, indeed, still queued. We make a best-effort to circle back and link to the relevant release notes once it hits the stable channel, but don't hurt to keep your eyes on the release page just in case! 

Thanks for your patience!

zachary-onix:
@scottisloud I still can't find it in `op plugin list` 🙈

PR #569, opened 2025-12-29
fix: correct environment variable for output directory

y-cg:
## Overview

A env var should be set to the output directory if `SetOutputDirAsEnvVar` is called. The current implementation override the wrong env var. This PR corrects the behavior.

Note: Non of the existing plugins will be effected by this fix (at the time of writing), I am the only one who try to call this function.

## Type of change
 

- [ ] Created a new plugin

## Related PR(s)
  

* Relates: #183

## How To Test

```bash
go test -count=1 ./sdk/provision -run . -v
```

## Changelog
  

Correct environment variable for output directory (SDK)

### DIFF
--- a/sdk/provision/file_provisioner.go
+++ b/sdk/provision/file_provisioner.go
@@ -130,7 +130,7 @@ func (p FileProvisioner) Provision(ctx context.Context, in sdk.ProvisionInput, o
 	if p.outdirEnvVar != "" {
 		// Populate the specified environment variable with the output dir.
 		dir := filepath.Dir(outpath)
-		out.AddEnvVar(p.outpathEnvVar, dir)
+		out.AddEnvVar(p.outdirEnvVar, dir)
 	}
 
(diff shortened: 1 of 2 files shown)


scottisloud:
Thanks for the fix! Your one-line change lines it up with the field semantics, and the test matches the `plugintest` temp-dir convention. I also confirmed no current plugin uses that option, so nothing shipping today changes behavior.

Two things before this can merge, neither about the code:

- Because the change is in `sdk/provision` (the shared provisioning path that compiles into the CLI), it needs an internal review on our side before merge. That's standard for any `sdk/` change; I'll take it forward, and given how contained this is I don't expect it to be involved.
- Could you rebase onto `main`? The branch is a few months old and only the signed-commits check has run, so I'd like Test and Lint green on a current base before we land it.

Thanks again for tracking this down, @y-cg.

scottisloud (approved):

PR #570, opened 2025-12-31
Add Descope Shell Plugin

scottisloud:
## Overview

This PR adds a shell plugin for Descope (https://www.descope.com). Currently it supports the authentication of an admin using a Project ID and Management Key. 

I am not a Descope user, so the nuances of that service are not well known to me. This may need to be reverted to draft if the main value lies in non-Descope admin CLI users. If that's the case I'll have to add functionality to handle Access Keys as well. 

References:
* https://docs.descope.com/cli/descope
* https://www.descope.com

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

* Resolves: #
* Relates: #

## How To Test

Reference: https://docs.descope.com/cli/descope
* Create a free account at descope.com
* Obtain the Project ID for the default project
* Generate a management key from the Company page of the Descope Admin Console
* Store the values in a 1Password item or provide them to `op plugin init descope` as prompted. 

## Changelog
  

Authenticate into the Descope CLI as a Descope account admin.


plugins/descope/management_key.go
```diff
@@ -0,0 +1,82 @@
...
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"DESCOPE_PROJECT_ID":     fieldname.ProjectID,
+	"DESCOPE_MANAGEMENT_KEY": fieldname.ManagementKey,
+}
+
+// TODO: Check if the platform stores the Management Key in a local config file, and if so,
+// implement the function below to add support for importing it.
+func TryDescopeConfigFile() sdk.Importer {
+	return importer.TryFile("~/path/to/config/file.yml", func(ctx context.Context, contents importer.FileContents, in sdk.ImportInput, out *sdk.ImportAttempt) {
+		// var config Config
```
Marton6: Based on the docs, the descope CLI doesn't store credentials on disk and doesn't support provisioning credentials from a file on disk, therefore we can leave out this function, the config struct below and the related test.

These are added to encourage contributors to double check if config files can also store credentials, but should be removed if not used, since having some unused code makes it harder to interpret the meaning of code and increases the risk of bugs.

plugins/descope/management_key.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func ManagementKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.ManagementKey,
+		DocsURL:       sdk.URL("https://docs.descope.com/cli/descope"),
```
Marton6, line 15: **Note:** Here we would usually put a link to documentation explaining what descope management keys are and how they work. Unfortunately, the descope docs don't seem to have a separate page explaining what descope management keys are, so there is nothing good for us to link here.

So in this case, I believe it's fine to use this link in the case of this shell plugin.

sdk/schema/fieldname/names.go
```diff
@@ -35,6 +35,7 @@ const (
 	Host            = sdk.FieldName("Host")
 	HostAddress     = sdk.FieldName("Host Address")
 	Key             = sdk.FieldName("Key")
+	ManagementKey   = sdk.FieldName("Management Key")
```
Marton6, line 38: **Note:** Usually I would have suggested using an existing field name instead of adding new ones, however, since the descope CLI has both "Management Keys" and "Access Keys" it is important to be clear to the descope and 1Password users with regards to which of these keys we're referring to. Therefore I think it is a great call to introduce this new field name here.

Marton6 (commented):
Looks great! Thank you for the contribution.

Left one blocking comment, and some general notes on the review process.

plugins/descope/management_key.go
scottisloud: I'll poke around their docs again just in case there's a better canonical resource but if not, we'll leave it as is.

plugins/descope/management_key.go
scottisloud: REMOVED!

sdk/schema/fieldname/names.go
scottisloud: I was on the fence about this one but ended up in the same place as you.

scottisloud:
Sorry @Marton6 need a fresh approval to merge this in after pushing the commit that addressed your comment. Thanks!

JillRegan (approved):

PR #589, opened 2026-03-26
Add PyPI plugin for twine, flit, and hatch

cpierce:
Adds a shell plugin that provisions PyPI API tokens for Python packaging tools. One credential in 1Password powers all three tools, with each getting its native env vars (TWINE_PASSWORD, FLIT_PASSWORD, HATCH_INDEX_AUTH). Also imports existing credentials from .pypirc files.

## Overview
Adds a PyPI shell plugin that provisions API tokens for Python packaging tools. A single credential in 1Password powers `twine`, `flit`, and `hatch`, with each tool receiving its native environment variables. Also imports existing credentials from `~/.pypirc` files and `TWINE_PASSWORD`, `FLIT_PASSWORD`, and `HATCH_INDEX_AUTH` environment variables respectively.

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
none 

* Resolves: #
* Relates: #

## How To Test
`twine upload dist/*`
`flit publish`
`hatch publish`

## Changelog
Authenticate twine, flit, and hatch to PyPI using Touch ID and other unlock options with 1Password Shell Plugins.


plugins/pypi/pypirc_importer.go
```diff
@@ -0,0 +1,36 @@
...
+		if err != nil {
+			return
+		}
+
+		// Try [pypi] section first, then [server-login]
+		for _, section := range []string{"pypi", "server-login"} {
+			s := cfg.Section(section)
+			if s == nil {
+				continue
+			}
```
scottisloud: Might be safer to 
```suggestion
            if !cfg.HasSection(section) {
                continue
            }
            s := cfg.Section(section)
            ...
```

since with current implementation will never receive a nil value (Section never returns nil, always returns a pointer)

plugins/pypi/api_token.go
```diff
@@ -0,0 +1,38 @@
...
+				Name:                fieldname.Token,
+				MarkdownDescription: "API token used to authenticate to PyPI.",
+				Secret:              true,
+				Composition: &schema.ValueComposition{
+					Prefix: "pypi-",
+					Charset: schema.Charset{
+						Uppercase: true,
+						Lowercase: true,
+						Digits:    true,
+						Specific:  []rune{'-', '_'},
```
scottisloud, line 27: Any concern that `=` is not in the charset, given the possibility of `=`-padded tokens?

scottisloud (changes requested):
Minor changes to discuss/implement.

plugins/pypi/api_token.go
cpierce: PyPI tokens are URL-safe base64-encoded macaroons; PyPI strips = padding before emitting them, so = shouldn't show up in a real token. Happy to add it if you think it's best though.

plugins/pypi/pypirc_importer.go
cpierce: good call - changed.

plugins/pypi/api_token.go
scottisloud: Not concerned if you aren't. If it's a problem we can address it then!

scottisloud (approved):
LGTM (...upon a second reviewer's approval)

JillRegan (approved):

scottisloud:
Thanks so much for your contribution! It should appear in an upcoming release of the 1Password CLI.

PR #595, opened 2026-05-06
Update workflows to fix node 20 deprecation and update golangci-lint-action version and config

scottisloud:
## Overview

This PR:
* Updates `actions/checkout` and `actions/setup-go` to v6, upping Node dependency to Node 24. 
* Updates `golangci-lint-action` to v9 and golangci version to 2.12. 
* Migrates the golangci config to conform to version 2
* Removes deprecated linters. 

## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  
* Resolves: OS-140
* Resolves: OS-141
* Resolves: OS-145

## Changelog
  

- Updates the following actions:
  - `actions/checkout` v4 -> v6
  - `actions/setup-go` v4 -> v6
  - `golangci/golangci-lint-action` v3 -> v9
  - `DeterminateSystems/nix-installer-action` main -> v22
  - `DeterminateSystems/magic-nix-cache-action` main -> v13
- Updated `Makefile` with new version of golangci-lint
- Updated `.golangci.yml` to conform to version 2 of the config file spec
- Added `use-flakehub: false` to silence flakehub account errors and default to GHA cache
- Disabled linters newly recognizing existing errors (to be re-enabled and detected errors fixed in a future PR)

### DIFF
--- a/.github/workflows/check-flake.yml
+++ b/.github/workflows/check-flake.yml
@@ -1,12 +1,13 @@
 name: Check Nix flake
 on:
-  pull_request:
-    paths:
-      - "**.nix"
-      - flake.lock
-  push:
-    branches:
-      - main
+  workflow_dispatch:
+    # pull_request:
+    #   paths:
+    #     - "**.nix"
+    #     - flake.lock
+    # push:
+    #   branches:
+    #     - main
 jobs:
   check-flake:
     runs-on: ubuntu-latest
--- a/.github/workflows/check-flake.yml
+++ b/.github/workflows/check-flake.yml
@@ -1,5 +1,6 @@
 name: Check Nix flake
 on:
+  workflow_dispatch:
   pull_request:
     paths:
       - "**.nix"
@@ -12,8 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-      - uses: DeterminateSystems/nix-installer-action@main
-      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - uses: DeterminateSystems/nix-installer-action@v22
+      - uses: DeterminateSystems/magic-nix-cache-action@v13
       - name: Check Nix flake inputs
         uses: DeterminateSystems/flake-checker-action@v5
         with:
--- a/.github/workflows/check-flake.yml
+++ b/.github/workflows/check-flake.yml
@@ -1,13 +1,12 @@
 name: Check Nix flake
 on:
-  workflow_dispatch:
-    # pull_request:
-    # paths:
-    # - "**.nix"
-    # - flake.lock
-    # push:
-    # branches:
-    # - main
+  pull_request:
+    paths:
+      - "**.nix"
+      - flake.lock
+  push:
+    branches:
+      - main
 jobs:
   check-flake:
     runs-on: ubuntu-latest
--- a/.github/workflows/check-flake.yml
+++ b/.github/workflows/check-flake.yml
@@ -5,9 +5,10 @@ on:
     #   paths:
     #     - "**.nix"
     #     - flake.lock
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yaml
```diff
@@ -31,14 +31,14 @@ jobs:
...
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v6
         with:
           go-version: 1.19
 
-      - name: Check out code
-        uses: actions/checkout@v3
-
       - name: Lint with golanci-lint
         uses: golangci/golangci-lint-action@v3
```
JillRegan: `golangci-lint-action` can be updated as well to [latest](https://github.com/golangci/golangci-lint-action/releases/tag/v9.0.0).

.github/workflows/update-flake-dependencies.yml
```diff
@@ -2,13 +2,13 @@
...
+    - cron: "0 16 * * 5"
   workflow_dispatch: # for allowing manual triggers of the workflow
 jobs:
   update-dependencies:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - uses: DeterminateSystems/nix-installer-action@main
       - uses: DeterminateSystems/magic-nix-cache-action@main
```
JillRegan: I know this is not in scope of these changes but I think we should update both `nix-installer-action` and `magic-nix-cache-action` to point at their latest tagged release and not main. Main branch changes frequently and can lead to day to day inconsistencies. 

Example: 
[latest](https://github.com/DeterminateSystems/nix-installer-action/releases/tag/v22) for `DeterminateSystems/nix-installer-action@v22`
[latest](https://github.com/DeterminateSystems/magic-nix-cache-action/releases/tag/v13) for `DeterminateSystems/magic-nix-cache-action@v13`

.github/workflows/check-flake.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   check-flake:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - uses: DeterminateSystems/nix-installer-action@main
       - uses: DeterminateSystems/magic-nix-cache-action@main
```
JillRegan: Like the comment below I think we should update these as well here to use latest tag version if we can.

JillRegan (commented):
Left a few comments (2 not in scope of this ticket but think we should update anyways). Besides that everything else looks good!

.github/workflows/test.yaml
scottisloud: `golangci-lint-action` might be tricky for two reasons:
* Breaking changes between the version we're using and the latest release (seemingly largely around the config file). 
* hard-coded references to the currently-used `golangci-lint-action` in `Makefile` (self-inflicted tech debt)

opened a separate work item for `golangci-lint` since it needs a bit of special care.

scottisloud: Manually migrating `.golangci.yml` to the v2 spec seemed to be all that was required, which took less time than expected!

scottisloud:
Some things out of scope for this PR:
* update-flake-dependencies workflow will [continue to error](https://github.com/1Password/shell-plugins/actions/runs/25516661899) on `gh api --method PUT /repos/1Password/shell-plugins/contents/$FILE_TO_COMMIT \` (likely due to expired OP_BOT_TOKEN)
* Fixes to golangci-lint-action revealed linting issues in existing code. Those linters have been temporarily disabled and need to be re-enabled once the issues are fixed.

JillRegan (approved):
Thanks for these updates @scottisloud 👍 LGTM!

rishiy15 (approved):

PR #596, opened 2026-05-06
Add VeraCrypt shell plugin

gdellis:
## Summary
Add VeraCrypt shell plugin for 1Password CLI, enabling secure mount operations with VeraCrypt volumes.

## Plugin Details
- **Plugin name**: veracrypt
- **Platform**: VeraCrypt
- **Credential**: Volume Password

## Features
- Custom provisioner injects password via `-p <password> --non-interactive` flags
- Supports `VERACRYPT_PASSWORD` environment variable for credential import
- Volume field included for future multi-volume support
- Text-mode CLI (`-t`) for Linux

## Testing
- All validation checks pass (14/14)
- All tests pass (5/5)

## Example Usage
```bash
# Mount a VeraCrypt volume
op plugin run veracrypt -- veracrypt -t --mount /path/to/volume /mnt/mount
```

### DIFF
--- a/plugins/veracrypt/plugin.go
+++ b/plugins/veracrypt/plugin.go
@@ -1,0 +1,22 @@
+package veracrypt
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/schema"
+)
+
+func New() schema.Plugin {
+	return schema.Plugin{
+		Name: "veracrypt",
+		Platform: schema.PlatformInfo{
+			Name:     "VeraCrypt",
+			Homepage: sdk.URL("https://www.veracrypt.fr"),
+		},
+		Credentials: []schema.CredentialType{
+			VolumePassword(),
+		},
+		Executables: []schema.Executable{
+			VeraCryptCLI(),
+		},
+	}
+}
--- a/plugins/veracrypt/veracrypt.go
+++ b/plugins/veracrypt/veracrypt.go
@@ -1,0 +1,25 @@
+package veracrypt
+
+import (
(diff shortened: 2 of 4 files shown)


scottisloud:
Thanks for the submission @gdellis! 

The custom provisioner means we're going to have do do a bit more thorough of a review. 

We may be a bit slow to get to that deeper review, so we appreciate your patience! 

In the meantime, a few things to chip away on/consider:

**Provisioning approach**
Using `-p` with `--non-interactive` is a reasonable v1 fit for shell plugins: the user is already at the terminal and unlocks 1Password first, and VeraCrypt documents `-p` even though it prefers `--stdin` for unattended scripting. I wouldn't block merge on switching to `--stdin` alone — that likely needs SDK/runtime stdin support anyway — but please add a brief comment in the provisioner noting why `-p` was chosen (no documented env-var path; attended single-user use). 

**Config-file importer (blocker)**
Please remove `TryVeraCryptConfigFile()` — it registers `~/.VeraCrypt/Config` but the callback is empty, and VeraCrypt does not store volume passwords on disk in config files (only preferences/favorites). If you want file-based import later, keyfiles (`-k`) would be a separate credential type, not a password importer.

**Auth gating (blocker)**
`NeedsAuth` should also skip commands that don't need a volume password — e.g. `--list` / `-l`, `--version`, and unmount (`-u` / `--dismount`) — not just help/version flags.

**Import env var (minor)**
`VERACRYPT_PASSWORD` is fine for import as a script convention, but it isn't a VeraCrypt CLI env var — worth a one-line note in the credential description so importers know it's import-only.

Cheers,
Scott

PR #598, opened 2026-05-09
feat(aws): add support for AWS IAM Identity Center (SSO)

### LINKED ISSUE #210: AWS - SSO authentication provider
### Platform or tool

AWS

AWS shell plugins supports SSO authentication

### Current behavior

AWS shell plugin only supports authentication via the long lived credentials or authentication via Sts temporary credentials for the MFA and Assume Role use cases.

### Relevant log output

_No response_

### op CLI version

_No response_

rjwhitworth:
## Overview

Adds AWS IAM Identity Center (SSO) authentication to the AWS plugin so users with SSO-only AWS access can authenticate the AWS CLI via 1Password. The plugin reads the access token cached by `aws sso login` from `~/.aws/sso/cache/<sha1>.json` and exchanges it for short-lived role credentials via `sso:GetRoleCredentials`. Both the legacy `sso_start_url` form and the consolidated `sso_session` form are supported. Role credentials get cached in the plugin's encrypted store, so subsequent invocations within the credential TTL skip the remote call.

Access Key and SSO Profile coexist on the same plugin: each provisioner yields silently when the active profile is configured for the other type, so users can link both items against the same plugin without conflict. The importer scans `~/.aws/config` for both forms and surfaces a candidate per SSO-bearing profile. The `cdk`, `eksctl`, `awslogs`, and `sam` executables also accept the SSO Profile credential.

`aws sso login`, `aws sso logout`, `aws configure sso`, `aws configure sso-session`, and the read-only `aws configure list/list-profiles/get/set` subcommands skip credential provisioning so they can run without an active session.

Two SDK changes:

- Plugins are now allowed to expose more than one credential type. The `op` runtime already accepts this; the validator was the only block.
- A new `AllowsExternalSecretCache` flag on `CredentialType` lets a credential opt out of the "must have at least one secret field" check. Only the SSO Profile uses it (its bearer token lives in the AWS SDK's external cache, not in the vault item). Every other credential continues to fail validation if no secret field is declared.

`aws-sdk-go-v2/credentials`, `aws-sdk-go-v2/service/sso`, and `smithy-go` were already on the dependency graph via `aws-vault/v7`; all three move from indirect to direct.

## Security hardening

The second commit on this branch addresses findings from a pre-PR adversarial security review:

- Importer (`plugins/aws/sso_importer.go`): validates `sso_start_url` (https, non-empty host), `sso_account_id` (12-digit regex), and `sso_region`. Rejects NUL bytes mid-value. Refuses non-regular, symlinked, or non-owned `AWS_CONFIG_FILE` overrides. Uses strict botocore-parity INI options: `KeyValueDelimiters="="`, `IgnoreContinuation=true`, `Loose=true`.
- Provisioner (`plugins/aws/sso_provisioner.go`): asserts `~/.aws/sso/cache/<sha1>.json` is a regular non-symlink file owned by the current uid with no group/world read bits before passing the path to `ssocreds.New`. AWS SSO API errors flow through a smithy code whitelist; unknown codes get a generic plugin-controlled message so server-controlled error text doesn't reach user-visible output. The `sso:GetRoleCredentials` call runs under a 30-second `context.WithTimeout`.
- The `AllowsExternalSecretCache` flag (above) replaces an earlier severity downgrade so the secret-field check stays an Error for the rest of the catalogue.

## Type of change

## Related Issue(s)

* Resolves: #210

## How To Test

1. Configure an SSO profile in `~/.aws/config` (either form). Legacy:

   ```
   [profile my-sso]
   sso_start_url = https://your-org.awsapps.com/start
   sso_region = us-east-1
   sso_account_id = 123456789012
   sso_role_name = AdministratorAccess
   region = us-east-1
   ```

   Or consolidated:

   ```
   [sso-session corp]
   sso_start_url = https://your-org.awsapps.com/start
   sso_region = us-east-1

   [profile my-sso]
   sso_session = corp
   sso_account_id = 123456789012
   sso_role_name = AdministratorAccess
   region = us-east-1
   ```

2. `aws sso login --profile my-sso` (populates `~/.aws/sso/cache/<sha1>.json`).
3. `op plugin init aws`, pick **SSO Profile**, accept the auto-detected candidate.
4. `op plugin run -- aws sts get-caller-identity --profile my-sso` — prints the assumed-role ARN.
5. Re-run within the credential TTL — the second invocation hits the plugin's encrypted cache (no SSO endpoint round-trip).
6. (Optional) Wipe `~/.aws/sso/cache/<sha1>.json` and re-run — the plugin should surface: `AWS SSO token is missing or expired; run \`aws sso login --profile my-sso\` and try again`.
7. (Optional, cache-file safety) `chmod 644 ~/.aws/sso/cache/<sha1>.json` and re-run on a cache-miss — the plugin should refuse with `AWS SSO token cache "..." is group/world readable (mode 644); chmod 600 it and re-run`. Restore `chmod 600` afterwards.

## Changelog

Authenticate the AWS CLI with AWS IAM Identity Center (SSO) profiles using cached SSO tokens.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,14 +25,14 @@ require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
 	github.com/aws/aws-sdk-go-v2/config v1.18.15 // indirect
-	github.com/aws/aws-sdk-go-v2/credentials v1.13.15 // indirect
+	github.com/aws/aws-sdk-go-v2/credentials v1.13.15
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.12.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/configsources v1.1.29 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.4.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/ini v1.3.30 // indirect
 	github.com/aws/aws-sdk-go-v2/service/iam v1.19.4 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.9.23 // indirect
-	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4 // indirect
+	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.4 // indirect
 	github.com/aws/smithy-go v1.13.5 // indirect
 	github.com/danieljoos/wincred v1.1.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.9.23 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.4 // indirect
-	github.com/aws/smithy-go v1.13.5 // indirect
+	github.com/aws/smithy-go v1.13.5
 	github.com/danieljoos/wincred v1.1.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dvsekhvalnov/jose2go v1.6.0 // indirect
--- a/plugins/aws/access_key_test.go
+++ b/plugins/aws/access_key_test.go
@@ -545,6 +545,77 @@ func TestSourceProfileLoop(t *testing.T) {
 	})
 }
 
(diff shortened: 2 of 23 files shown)


scottisloud:
Hi @rjwhitworth thanks for the contribution. 

Preliminary review and testing completed, along with a couple tweaks to handle some edge-cases. 

Some additional review and documentation is likely required here to help users leverage the new functionality here. We will discuss internally the best approach here. 

Thanks for the review and I will update this when we have more to share.

scottisloud:
Two concerns:
* Access Key is always required, even for SSO. That may be as close as we can get, but will have to decide if we are willing to accept that SSO users who cannot provide an Access Key will not be able to use this. Personally I'm not _that_ worried about this being the case. 
* May need to make both creds `Optional: true` to minimize impact on existing users. 

There's also a minor CI failure that should be fixed up:
`QF1008: could remove embedded field "stsCacheWriter" from selector (staticcheck)` on line 183 of https://github.com/1Password/shell-plugins/pull/598/changes#diff-5f9c4fbb934c4a88db98623880a9d514b76b6d7546ec128e590cb8068aee3b66

PR #599, opened 2026-05-13
Fix linting errors and re-enable golangci linters

scottisloud:
## Overview

Following up on #595, the `errcheck`, `depguard`, and `staticcheck` linters were previously disabled in `.golangci.yml` because they were producing errors after the linter had been nonfunctional for an extended period of time. This PR resolves all outstanding issues and re-enables them so every configured linter runs in CI.

Closes OS-148

## Type of change
 

- [ ] Created a new plugin

## Changelog
  
**Changes**
`depguard` (1 fix)
- sdk/plugintest/example_secrets.go — swapped math/rand for the safer crypto/rand.

`errcheck` (4 fixes)
- sdk/plugintest/validation_report.go — checked the return values of the Printf calls on Heading, Success, Warning, and Error formatters.

`staticcheck` (7 fixes / suppressions)
- plugins/readme/api_key.go — dropped the redundant string type declaration (ST1023).
- sdk/plugintest/validation_report.go — removed the unnecessary nil check before len() (S1009).
- sdk/schema/executable.go — simplified a boolean expression via De Morgan's law (QF1001).
- plugins/aws/sts_provisioner.go and sdk/rpc/server/server.go — added targeted //nolint exclusions for QF1008 (embedded-field selectors) where the explicit form was preferred for readability.

`.golangci.yml`
- Removed the `disable:` block; `errcheck`, `depguard`, and `staticcheck` are now in the `enable:` list alongside the existing defaults.

**Testing**

golangci-lint run now exits cleanly with all three linters active.

### DIFF
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -271,7 +271,7 @@ func (p assumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, erro
 		return aws.Credentials{}, err
 	}
 
-	err = p.stsCacheWriter.Put(credentials)
+	err = p.stsCacheWriter.Put(credentials) //lint:ignore QF1008 explicit reads more clearly here
 	if err != nil {
 		return aws.Credentials{}, err
 	}
@@ -308,7 +308,7 @@ func (p mfaSessionTokenProvider) Retrieve(ctx context.Context) (aws.Credentials,
 		return aws.Credentials{}, err
 	}
 
-	err = p.stsCacheWriter.Put(credentials)
+	err = p.stsCacheWriter.Put(credentials) //lint:ignore QF1008 explicit reads more clearly here
 	if err != nil {
 		return aws.Credentials{}, err
 	}
--- a/plugins/aws/sts_provisioner.go
+++ b/plugins/aws/sts_provisioner.go
@@ -271,7 +271,7 @@ func (p assumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, erro
 		return aws.Credentials{}, err
 	}
 
-	err = p.stsCacheWriter.Put(credentials) //lint:ignore QF1008 explicit reads more clearly here
+	err = p.stsCacheWriter.Put(credentials) //nolint:staticcheck // QF1008 explicit reads more clearly here
 	if err != nil {
 		return aws.Credentials{}, err
 	}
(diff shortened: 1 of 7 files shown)


plugins/aws/sts_provisioner.go
```diff
@@ -271,7 +271,7 @@ func (p assumeRoleProvider) Retrieve(ctx context.Context) (aws.Credentials, erro
 		return aws.Credentials{}, err
 	}
 
-	err = p.stsCacheWriter.Put(credentials)
+	err = p.stsCacheWriter.Put(credentials) //nolint:staticcheck // QF1008 explicit reads more clearly here
 	if err != nil {
 		return aws.Credentials{}, err
 	}
```
JillRegan, line 277: On the fence about this ones and below. I do agree it reads more clearly this way though... but I don't love disabling the lint check. 

I'll leave it up to second reviewer on this. If others feel it is also easier to read this way let's keep :)

JillRegan (approved):
Left a small comment that I don't have a strong opinion about so approving! 🙌

plugins/aws/sts_provisioner.go
scottisloud: I was torn on this one too.... happy for review #2 to be the tie breaker.

bertrmz: Agreed that this style reads more clearly.

If we'd like to accept this as the in-house style, golangci-lint can be configured to omit it. The only annoying/tricky part is having to restate their defaults, but it'd look something like:
```diff
diff --git a/.golangci.yml b/.golangci.yml
index a1b6dd4..3256ca9 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,6 +14,18 @@ linters:
     - staticcheck
 
   settings:
+    staticcheck:
+      # Mirrors golangci-lint's default staticcheck check set, plus -QF1008
+      # so we can keep explicit embedded-field reads as house style.
+      checks:
+        - all
+        - -ST1000 # Incorrect or missing package comment
+        - -ST1003 # Poorly chosen identifier
+        - -ST1016 # Use consistent method receiver names
+        - -ST1020 # The documentation of an exported function should start with the function's name
+        - -ST1021 # The documentation of an exported type should start with type's name
+        - -ST1022 # The documentation of an exported variable or constant should start with variable's name
+        - -QF1008 # Omit embedded fields from selector expression
     depguard:
       rules:
         main:
```

scottisloud: As much as I'm a fan of the explicitness of the more verbose code, I'm happy to save us a bunch of fuss, do the more idiomatic Go thing, and nuke those embedded fields. 

If we have to re-state the defaults in the golanci config, then I think we're just creating maintenance overhead for ourselves.  

But still happy to defer to your preference here!

bertrmz: Yeah, that makes sense. Since there are only 4 cases right now, it seems unlikely this was an intentional stylistic choice.

& having the rule in place should make it easier for contributors to stay consistent.

scottisloud: Removed exceptions and corrected violations in https://github.com/1Password/shell-plugins/pull/599/commits/1391bd9e9680112d273e8d0a726121d1f778f278

bertrmz (approved):

PR #604, opened 2026-05-29
Add JetBrains Junie CLI shell plugin

eddumelendez:
## Overview

Adds a new JetBrains Junie shell plugin for authenticating the `junie` CLI with a Junie API key from 1Password.

The plugin provisions and imports `JUNIE_API_KEY` only. Provider-specific BYOK variables such as `JUNIE_OPENAI_API_KEY` and `JUNIE_ANTHROPIC_API_KEY` are intentionally not supported by this plugin.

## Type of change

- [x] Created a new plugin

## Related Issue(s)

N/A

## How To Test

For functional testing, initialize the plugin with a Junie API key and run an authenticated Junie command, for example:

```sh
junie "Summarize this project"
```

## Changelog
Authenticate JetBrains Junie CLI using Touch ID and other unlock options with 1Password Shell Plugins.

### DIFF
--- a/plugins/junie/api_key.go
+++ b/plugins/junie/api_key.go
@@ -1,0 +1,31 @@
+package junie
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+func APIKey() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.APIKey,
+		DocsURL:       sdk.URL("https://junie.jetbrains.com/docs/environment-variables.html"),
+		ManagementURL: sdk.URL("https://junie.jetbrains.com/cli"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "Junie API key used to authenticate to JetBrains Junie CLI.",
+				Secret:              true,
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
+		Importer:           importer.TryEnvVarPair(defaultEnvVarMapping),
+	}
+}
+
+var defaultEnvVarMapping = map[string]sdk.FieldName{
+	"JUNIE_API_KEY": fieldname.APIKey,
+}
--- a/plugins/junie/api_key_test.go
+++ b/plugins/junie/api_key_test.go
@@ -1,0 +1,72 @@
+package junie
+
+import (
+	"testing"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/plugintest"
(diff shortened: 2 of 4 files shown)


plugins/junie/api_key.go
```diff
@@ -0,0 +1,31 @@
...
+		DocsURL:       sdk.URL("https://junie.jetbrains.com/docs/environment-variables.html"),
+		ManagementURL: sdk.URL("https://junie.jetbrains.com/cli"),
+		Fields: []schema.CredentialField{
+			{
+				Name:                fieldname.APIKey,
+				MarkdownDescription: "Junie API key used to authenticate to JetBrains Junie CLI.",
+				Secret:              true,
+			},
+		},
+		DefaultProvisioner: provision.EnvVars(defaultEnvVarMapping),
```
eddumelendez, line 24: Junie have specific env vars to authenticate against providers via BYOK. See https://junie.jetbrains.com/docs/environment-variables.html#provider-api-keys
Is there any way to implement it in the shell plugins? I have been also looking into support OpenCode which relies on env vars from providers as well.

scottisloud: There may be some clever way to implement this in the shell plugin, but I wonder if this use-case may be better served by a different approach? 

[1Password Developer Environments](https://www.1password.dev/environments), for example, could be used, and accessed at a scope that makes sense to by:
* using a local file mount destination in the project root that includes any env vars your project needs along with the desired AI provider tokens you want to use with Junie on that project.

[1Password Secret References](https://www.1password.dev/cli/secret-reference-syntax) 
* export environment variables for the AI provider in the normal way you do, setting the value of those variables to secret references to those provider API tokens in 1Password. 
* prefix any commands requiring those tokens with `op run [--env-file=path/to/env/if/relevant] -- some command` 

Those are just a few ideas. Certainly there are many other ways to leverage the above 1Password features, or other features entirely, that might offer a bit more flexibility.

scottisloud:
Thanks for the contribution! 
Overall looks solid: `JUNIE_API_KEY` provisioning and import path matches Junie's documented auth, and the tests (including the negative cases for BYOK vars) look solid. One blocker before merge:

- **Bare `junie` auth gating:** Junie's quickstart starts interactive mode with bare `junie`, where users can sign in with a JetBrains Account, enter a Junie API key in the TUI, or configure BYOK. `NeedsAuth` currently only exempts help and version, so bare `junie` would ask 1Password for an API key before the TUI can start. Could you update `junie.go` to combine `NotForHelpOrVersion()` with `NotWithoutArgs()` via `needsauth.IfAll(...)`? `plugins/vercel/vercel.go` and `plugins/ngrok/ngrok.go` are close patterns to follow. Should be largely the same fix as in the other PR you submitted recently. 

**optional:** On the BYOK provider env vars — keeping this PR scoped to `JUNIE_API_KEY` only is reasonable; separate credential types could be a follow-up if there's demand.

I'll take another look once that's in. Thanks again, @eddumelendez.

scottisloud (changes requested):

eddumelendez:
> On the BYOK provider env vars — keeping this PR scoped to JUNIE_API_KEY only is reasonable; separate credential types could be a follow-up if there's demand.

I took a look at #610 and thought we can apply something similar. But, agree this can be a follow-up.

scottisloud (approved):
LGTM! Keep an eye out for this plugin to be in a future release of the 1Password CLI! 

Thanks for the contribution!

PR #608, opened 2026-06-02
appending advanced configuration to readme

### LINKED ISSUE #378: Allow plugins.sh to be stowed or tracked in dotfiles as symlinks
### op CLI version

2.19.0-beta.01

### Goal or desired behavior

Currently, the CLI expects `~/.config/op/plugins.sh` to be a file. I would like to keep it in my dotfiles and symlink it to the requisite location.

### Current behavior

`[ERROR] 2023/09/03 01:15:00 Can't continue. We can't safely access "/Users/mrugesh/.config/op/plugins.sh" because it's a symbolic link. Replace the linked file with the original file and try again.`

### Relevant log output

_No response_

scottisloud:
## Overview

Adds documentation for a workaround for people who choose to stow or symlink their `op`-related configuration files. 

The existing header levels didn't necessarily make it easy to slot this in, and as is, won't scale if we add docs for any additional "advanced configurations". But since we've got some custom stuff going on here I didn't want to step too far out of my lane and refactor the header levels in the readme. 
## Type of change
 

- [ ] Created a new plugin

## Related Issue(s)
  

* Resolves: #378

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Still not sure where or how to begin? We're happy to help! You can:
 By default, `op plugin init` writes its configuration to `~/.config/op/plugins.sh`
 and manages that file for you. If you'd rather keep your plugin configuration in
 your own dotfiles (for example, tracked in version control), you can define the
-shell functions yourself instead.
+shell functions yourself and source them in your shell's configuration file. 
 
 ### Step 1: Define the plugin functions
 Shell functions for each shell should follow this pattern. In this case, 
@@ -92,18 +92,18 @@ end
 
 ### Step 2: Tell `op` your plugins are already configured
 
-Set `OP_PLUGINS_SOURCED=1` so 1Password CLI knows the shell plugins have
+Set `OP_PLUGIN_ALIASES_SOURCED=1` so 1Password CLI knows the shell plugins have
 already been set up:
 
 ### bash / zsh
 
 ```bash
-export OP_PLUGINS_SOURCED="1"
+export OP_PLUGIN_ALIASES_SOURCED="1"
 ```
 
 ### fish
 ```fish
-set -x OP_PLUGINS_SOURCED 1
+set -x OP_PLUGIN_ALIASES_SOURCED 1
 ```
 
 > **Note:** `op plugin init <plugin>` edits `~/.config/op/plugins.sh` directly and won't update your own dotfiles, so configure plugins manually as shown above.
--- a/README.md
+++ b/README.md
@@ -63,6 +63,52 @@ Still not sure where or how to begin? We're happy to help! You can:
 - Book a free [pairing session](https://calendly.com/d/grs-x2h-pmb/1password-shell-plugins-pairing-session) with one of our developers
 - Join the [Developer Slack workspace](https://developer.1password.com/joinslack), and ask us any questions there
 
+## Advanced Configuration
+### Managing shell plugins in your own dotfiles
+
+By default, `op plugin init` writes its configuration to `~/.config/op/plugins.sh`
+and manages that file for you. If you'd rather keep your plugin configuration in
+your own dotfiles (for example, tracked in version control), you can define the
+shell functions yourself instead.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,8 +63,54 @@ Still not sure where or how to begin? We're happy to help! You can:
...
 
+## Advanced Configuration
+### Managing shell plugins in your own dotfiles
+
+By default, `op plugin init` writes its configuration to `~/.config/op/plugins.sh`
+and manages that file for you. If you'd rather keep your plugin configuration in
+your own dotfiles (for example, tracked in version control), you can define the
+shell functions yourself instead.
+
+### 1. Define the plugin functions
```
libutcher:
```suggestion

### Step 1: Define the plugin functions
```

To follow our typical style for formatting steps

README.md
```diff
@@ -63,8 +63,54 @@ Still not sure where or how to begin? We're happy to help! You can:
...
+Shell functions for each shell should follow this pattern. In this case,
```
libutcher, line 75: Is there a specific action people need to take here? I'm not super familiar with this, but wondering if we could make it a bit more beginner friendly by giving an example of when/where to define these functions.

README.md
```diff
@@ -63,8 +63,54 @@ Still not sure where or how to begin? We're happy to help! You can:
...
+
+```fish
+function gh --wraps gh --description "1Password shell plugin for GitHub CLI"
+  op plugin run -- gh $argv
+end
+```
+
+### 2. Tell `op` your plugins are already configured
+
+Set `OP_PLUGINS_SOURCED=1` so the 1Password CLI knows the shell plugins have
```
libutcher:
```suggestion
Set `OP_PLUGINS_SOURCED=1` so 1Password CLI knows the shell plugins have
```

README.md
```diff
@@ -63,8 +63,54 @@ Still not sure where or how to begin? We're happy to help! You can:
...
+
+### fish
+
+```fish
+function gh --wraps gh --description "1Password shell plugin for GitHub CLI"
+  op plugin run -- gh $argv
+end
+```
+
+### 2. Tell `op` your plugins are already configured
```
libutcher:
```suggestion
### Step 2: Tell 1Password CLI your plugins are already configured
```

To follow our typical style for how we refer to the CLI

libutcher (approved):
This looks great! Approved with some style suggestions and one question. It would be awesome to add an article for this under guides in the shell plugin docs on 1password.dev too.

README.md
scottisloud: This is a workaround for people who insist on maintaining their dotfiles using some management system, which usually requires storing the files in one place and sym-linking them to the directories applications (like 1Password) expect to find them. 

There are security reasons we are a bit careful with symlinked files so anyone taking this approach to their dotfiles has to jump through some hoops. This can mean manually defining the shell functions/aliases that wrap the CLI command (e.g., `gh`) in `op plugin run` since, if they're symlinking files, we're unable to write to the symlinked file. 

So this effectively documents manually performing what `op plugin init` would otherwise do automatically but can't in environments configured this way.

scottisloud: Let me review the existing directions here and see if we can be a bit more hands-on with the guidance!

scottisloud: I've made a minor change to clarify things a bit. 
TL;DR: as long as we are clear about what we expect, any user wanting to do this should _know_ how to do this according to their preference and the requirements of whatever they use to manage dotfiles. 

Happy to discuss this though! I might be a bit off the mark! 


Longer read: 
This is a rather power user need, and there's a lot of variation in exactly how people who want to centrally manage their dotfiles might do so. On our end, we just need to make sure users are clear on what 1Password expects (shell functions that follow the pattern documented, and are `sourced` by their shell config). Where they write those functions, how they source that file in their shell config, and where they store their shell config are largely user-defined, and anyone embarking on such an effort will likely have opinions about how to do this, or their dotfile manager will have opinions.

PR #613, opened 2026-06-18
Add OpenCode shell plugin

eddumelendez:
## Overview

Adds a new OpenCode shell plugin for the `opencode` CLI.

The plugin provisions an auth token through `OPENCODE_API_KEY`. It also supports importing an existing token from `OPENCODE_API_KEY`.

## Type of change

- [x] Created a new plugin

## Related Issue(s)

N/A

## How To Test

Run plugin validation:

```shell
make opencode/validate
```

Run the OpenCode plugin tests:

```shell
go test ./plugins/opencode
```

Example authenticated command for functional testing:

```shell
opencode run "explain this repo"
```

## Changelog
Authenticate the OpenCode CLI using Touch ID and other unlock options with 1Password Shell Plugins.

### DIFF
--- a/plugins/opencode/api_key.go
+++ b/plugins/opencode/api_key.go
@@ -17,7 +17,7 @@ func APIKey() schema.CredentialType {
 		Fields: []schema.CredentialField{
 			{
 				Name:                fieldname.APIKey,
-				MarkdownDescription: "API Key used to authenticate to opencode.",
+				MarkdownDescription: "API Key used to authenticate to OpenCode.",
 				Secret:              true,
 			},
 		},
--- a/plugins/opencode/api_key.go
+++ b/plugins/opencode/api_key.go
@@ -1,0 +1,31 @@
+package opencode
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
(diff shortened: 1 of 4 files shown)


scottisloud (changes requested):
Thanks for adding OpenCode support!

One blocking change before merge:

- **`NeedsAuth` should skip `opencode auth` subcommands.** Right now `opencode auth login`, `list`, and `logout` would still require a 1Password item first, which blocks the documented setup flow for users who have not configured the plugin yet. Please add `needsauth.NotWhenContainsArgs("auth")` alongside the existing helpers in `opencode.go` (same pattern as `plugins/cline/cline.go` and `plugins/upstash/upstash.go`).

Optional polish (not blocking):

- The field description could note this is an **OpenCode Zen** API key from [opencode.ai/auth](https://opencode.ai/auth) and that `OPENCODE_API_KEY` covers _only_ the  Zen gateway, not BYOK keys like `ANTHROPIC_API_KEY`.
- Credential `DocsURL` could point at the [Providers / OpenCode Zen](https://opencode.ai/docs/providers/) section instead of the intro page.

Once the `auth` exclusion is in, I'll take another look!

scottisloud (approved):
Thanks again for making these changes and for the contribution @eddumelendez ! LGTM

Look out for this in a future version of the 1Password CLI.

PR #620, opened 2026-07-03
fix: Gittea write config to XDG_CONFIG_HOME when set

### LINKED ISSUE #601: [gitea] Plugin writes config to a path tea won't read when $XDG_CONFIG_HOME is set
On macOS with `$XDG_CONFIG_HOME` exported, `tea` always errors with `Error: no available login` after `op plugin init tea`.

The plugin writes to `~/Library/Application Support/tea/config.yml` ([source](https://github.com/1Password/shell-plugins/blob/ba97aedc26410886efa92d7cbfda3ba73d37fd29/plugins/gitea/personal_access_token.go#L16-L22), via `os.UserConfigDir()`), but tea resolves its config through `adrg/xdg`, which prefers `$XDG_CONFIG_HOME` when set. Go's `os.UserConfigDir()` doesn't read `$XDG_CONFIG_HOME` on darwin, so the two diverge and tea finds an empty config.

Without `$XDG_CONFIG_HOME` both sides default to `~/Library/Application Support`, so it only bites users who export it (common when syncing dotfiles across Linux + macOS).

### Repro
macOS, tea 0.14.1, `export XDG_CONFIG_HOME=$HOME/.config`, `op plugin init tea`, then `tea pulls` → `Error: no available login`.

### Suggested fix
Honor `$XDG_CONFIG_HOME` first in `ConfigPath()`:

```go
if xdg := os.Getenv("XDG_CONFIG_HOME"); xdg != "" {
    return filepath.Join(xdg, "tea", "config.yml")
}
```

### Workaround
`ln -s "$HOME/Library/Application Support/tea" "$XDG_CONFIG_HOME/tea"`

rr3khan:
## Overview

On macOS with `$XDG_CONFIG_HOME` exported, `op plugin init tea` writes tea's config to a path tea never reads, so every `tea` command fails with `Error: no available login`.

The plugin resolves the config path with `os.UserConfigDir()`, which ignores `$XDG_CONFIG_HOME` on darwin, while tea resolves it through [`adrg/xdg`](https://github.com/adrg/xdg), which prefers `$XDG_CONFIG_HOME` on every platform. When the variable is set (common for users syncing dotfiles across Linux and macOS), the two diverge: the plugin writes to `~/Library/Application Support/tea/config.yml` while tea reads from `$XDG_CONFIG_HOME/tea/config.yml`.

`ConfigPath()` now checks `$XDG_CONFIG_HOME` first and falls back to `os.UserConfigDir()`, matching tea's own resolution order. Added tests covering both branches.

## Type of change

- [ ] Created a new plugin

## Related Issue(s)

* Resolves: #601

## How To Test

```sh
export XDG_CONFIG_HOME=$HOME/.config
op plugin init tea   # link a Gitea personal access token
tea pulls            # previously: Error: no available login
```

Without `$XDG_CONFIG_HOME` set, behavior is unchanged (`os.UserConfigDir()` fallback). `go test ./plugins/gitea/...` and `make gitea/validate` pass.

## Changelog

The Gitea plugin now writes tea's configuration to the location tea reads when `XDG_CONFIG_HOME` is set.

### DIFF
--- a/plugins/gitea/personal_access_token.go
+++ b/plugins/gitea/personal_access_token.go
@@ -3,6 +3,7 @@ package gitea
 import (
 	"context"
 	"os"
+	"path/filepath"
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
@@ -14,11 +15,16 @@ import (
 )
 
 func ConfigPath() string {
+	// tea's config resolver honors XDG_CONFIG_HOME on macOS; os.UserConfigDir does not.
+	if xdgConfigHome := os.Getenv("XDG_CONFIG_HOME"); xdgConfigHome != "" {
+		return filepath.Join(xdgConfigHome, "tea", "config.yml")
+	}
+
 	configDir, err := os.UserConfigDir()
 	if err != nil {
 		return "~/.config/tea/config.yml"
 	}
-	return configDir + "/tea/config.yml"
+	return filepath.Join(configDir, "tea", "config.yml")
 }
 
 func PersonalAccessToken() schema.CredentialType {
--- a/plugins/gitea/personal_access_token_test.go
+++ b/plugins/gitea/personal_access_token_test.go
@@ -1,12 +1,42 @@
 package gitea
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


scottisloud (approved):
Looks good! LGTM.

PR #622, opened 2026-07-06
feat(aws): add support for AWS IAM Identity Center (SSO)

### LINKED ISSUE #210: AWS - SSO authentication provider

rr3khan:
## Overview

Follow-up to #598, which added AWS IAM Identity Center (SSO) support to the AWS plugin but was closed. See the #598 description for the full design; in short, the plugin reads the access token cached by `aws sso login` and exchanges it for short-lived role credentials via `sso:GetRoleCredentials`, supporting both the legacy `sso_start_url` and consolidated `sso_session` config forms.

On top of that branch, this PR adds:

- `Optional: true` on both credential usages (Access Key and SSO Profile) across all five executables, addressing the review concern that SSO-only users, who have no IAM access key at all, would otherwise be locked out.
- https://github.com/1Password/shell-plugins/pull/622/changes/e8accd80c34a8735a879ee553b55465c8cfa9814
- https://github.com/1Password/shell-plugins/pull/622/changes/fac5e1ee5feb615e0d27457914ebcddb616f1ea8

## Local testing results

Tested against a live IAM Identity Center org with op 2.30.0 and a local plugin build:

- Full SSO flow works: `op plugin init aws` imports the SSO profile detected in `~/.aws/config`, and `op plugin run -- aws sts get-caller-identity --profile <sso-profile>` returns the assumed-role ARN. A repeat invocation is served from the plugin's encrypted credential cache without a round-trip to the SSO endpoint.
- The importer handled the consolidated `sso_session` form, including a session whose name is the start URL itself (what `aws configure sso` writes if you paste the URL at the session-name prompt).
- Both credential types coexist on the plugin: an Access Key item and an SSO Profile item linked side by side, with the right provisioner engaging per profile.
-  Both credential usages are marked Optional: true. With op 2.30.0, plugin init then offers a credential-type chooser with "Stop choosing credentials", so SSO-only users can link just an SSO Profile (and access-key users just an Access Key). Without the flag, init requires linking an item for every credential type.

One limitations observed requiring CLI-side changes rather than plugin changes:

- The init prompt labels every credential step "Locate your AWS Access Key", including the SSO Profile step. The structure validation runs against the correct credential type, only the label is wrong.

Screenshot 2026-07-07 at 1 01 00 PM

Not included here: the pre-existing SDK findings deferred in `b90170b`'s commit message (F-2, F-6, F-7, F-9, F-15). Worth noting that F-2 (`GetCredentialType` ignores `credentialName`) and F-6 (`MarshalJSON` reads `Credentials[0]`) become reachable now that this is the first plugin to expose two credential types. Happy to pick those up in a follow-up PR.

## Type of change
 

## Related Issue(s)
  

* Resolves: https://github.com/1Password/shell-plugins/issues/210
* Continues off from https://github.com/1Password/shell-plugins/pull/598 

## How To Test

The steps from #598 still apply:

Tested with op 2.30.0 and a local build (`make aws/build`) against a live IAM Identity Center org:

1. Configure an SSO profile in `~/.aws/config` and run `aws sso login --profile <name>`.
   The login command passes through without provisioning, by design.
2. Run `op plugin init aws`. At "Choose a credential type", pick **AWS SSO Profile** and
   link an SSO Profile item, then pick **Stop choosing credentials** — no Access Key item
   is required. Note: the locate prompt is mislabeled "Locate your AWS Access Key"
   (see limitation above); it still validates against the SSO Profile item.
3. Run `aws sts get-caller-identity --profile <name>` — prints the assumed-role ARN.

To retest from a clean slate, `op plugin clear aws` resets the plugin's saved credential
configuration.

## Changelog
  

Authenticate the AWS CLI with AWS IAM Identity Center (SSO) profiles using cached SSO tokens.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,14 +25,14 @@ require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
 	github.com/aws/aws-sdk-go-v2/config v1.18.15 // indirect
-	github.com/aws/aws-sdk-go-v2/credentials v1.13.15 // indirect
+	github.com/aws/aws-sdk-go-v2/credentials v1.13.15
 	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.12.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/configsources v1.1.29 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.4.23 // indirect
 	github.com/aws/aws-sdk-go-v2/internal/ini v1.3.30 // indirect
 	github.com/aws/aws-sdk-go-v2/service/iam v1.19.4 // indirect
 	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.9.23 // indirect
-	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4 // indirect
+	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.4 // indirect
 	github.com/aws/smithy-go v1.13.5 // indirect
 	github.com/danieljoos/wincred v1.1.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.9.23 // indirect
 	github.com/aws/aws-sdk-go-v2/service/sso v1.12.4
 	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.14.4 // indirect
-	github.com/aws/smithy-go v1.13.5 // indirect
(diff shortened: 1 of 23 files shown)


scottisloud (commented):
Overall things look great. Going to bring this to the team to talk about the potential changes required to `op` to accommodate certain use-cases and whether those changes are a pre-requisite to shipping this plugin, and for a second set of eyes on the new provisioner. 

Nothing is immediately standing out to me, and I'm not inclined to worry about the use cases requiring changes to `op` necessarily, since this is a net-improvement overall. 

Provisional approval, will have a second set of eyes on this before actual approval and merge.

scottisloud:
After some internal discussion we want to do a bit more work to understand the implications of some of the bigger changes this plugin introduces. No showstoppers or blockers, just want to ensure that we are deliberate about introducing some new concepts, and some understanding of the changes that would need to at least eventually be made to `op`.

PR #625, opened 2026-07-16
[DRAFT] feat(plugin) : Add Apono plugin

rr3khan:
## Overview

Adds a shell plugin for the [Apono CLI](https://github.com/apono-io/apono-cli) (`apono`), which requests and receives just-in-time access to services, databases, and applications.

Apono personal API tokens are created in the Apono web UI (Settings → Personal API Token) and are long-lived bearer credentials, so they're a good fit for 1Password storage. This plugin:

- Stores the **Personal API Token** credential in 1Password (`credname.PersonalAPIToken`, single `Token` field).
- Provisions it as the `--personal-token` flag of `apono login` via a custom arg provisioner, scoped with `needsauth.ForCommand("login")`.
- Imports existing tokens from the CLI's `config.json` (`~/Library/Application Support/apono-cli/config.json` on macOS, `~/.config/apono-cli/config.json` on Linux), one candidate per profile that has a `personal_token` set, skipping OAuth-only profiles.

### Why the provisioner targets `apono login` instead of environment variables

The Apono CLI currently reads no credentials from environment variables and has no per-command token flag — the only token injection point is `apono login --personal-token <TOKEN>`, which validates the token and persists it to the CLI's own `config.json`. This plugin therefore keeps the token out of shell history and clipboards and enables 1Password import/storage, but after login the CLI still stores the token on disk by its own design.

apono-io/apono-cli#116 (draft) adds `APONO_PERSONAL_TOKEN` support to the CLI. Once that ships, this plugin can switch to a standard `provision.EnvVars` provisioner covering every `apono` command per-invocation, with nothing written to disk. Happy to do that as a follow-up here.

`ManagementURL` is intentionally omitted for now; will add the exact settings-page URL once confirmed.

## Type of change

- [x] Created a new plugin

## Related Issue(s)

* Relates: apono-io/apono-cli#116

## How To Test

1. Create a personal API token in the Apono web UI (Settings → Personal API Token).
2. `make apono/build && op plugin init apono` and save the token as a Personal API Token item.
3. Run `apono login` — the token should be injected as `--personal-token` and the CLI should create a profile without opening a browser.
4. Run `apono access list` to confirm the authenticated session works.
5. Importer: with an existing `config.json` containing a `personal_token`, `op plugin init apono` should offer to import it.

Unit tests: `go test ./plugins/apono/...`; schema: `make apono/validate` (all checks pass; single warning for the omitted management URL).

## Changelog

Authenticate the Apono CLI using Touch ID and other unlock options with 1Password Shell Plugins.

### DIFF
--- a/plugins/apono/apono.go
+++ b/plugins/apono/apono.go
@@ -1,0 +1,29 @@
+package apono
+
+import (
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/needsauth"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+)
+
+func AponoCLI() schema.Executable {
+	return schema.Executable{
+		Name:    "Apono CLI",
+		Runs:    []string{"apono"},
+		DocsURL: sdk.URL("https://docs.apono.io/docs/access-requests-and-approvals/cli/install-and-manage-the-apono-cli"),
+		NeedsAuth: needsauth.IfAll(
+			// The Apono CLI only accepts a token on "apono login"; all other
+			// commands authenticate through the session it stores itself.
+			needsauth.ForCommand("login"),
+			needsauth.NotForHelp(),
+			needsauth.NotWhenContainsArgs("--personal-token"),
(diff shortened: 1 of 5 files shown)


rr3khan:
On pause while I wait for a test account from Apono

PR #638, opened 2026-07-21
[fix] fix update-flake-dependencies failures

scottisloud:
This PR:
* checks for the presence of a PR on the `automation/update-flake-dependencies` branch, rather than the previous logic which checked whether a branch of that name exists. 
  * If a PR doesn't exist, it opens a PR. 
* Updated the list of reviewers to be current and recent members of the Open Source team.  

**Background**
update-flake-dependencies would only open a PR with the updated flake.lock file if a branch called `automation/update-flake-dependencies` did not exist. Since that branch was not reliably deleted, and hasn't been for years, the workflow would find the branch and thus not open a PR. 

See this workflow test run which is executed off of this branch: https://github.com/1Password/shell-plugins/pull/656

### DIFF
--- a/.github/workflows/update-flake-dependencies.yml
+++ b/.github/workflows/update-flake-dependencies.yml
@@ -28,11 +28,9 @@ jobs:
           # fetch remote state
           git fetch
           # if branch exists on remote already
-          BRANCH_EXISTS=false
           if git checkout "$COMMIT_BRANCH" > /dev/null 2>&1; then
             # pull changes
             git pull
-            BRANCH_EXISTS=true
           else
             # otherwise, create the branch and push it to remote
             git checkout -b "$COMMIT_BRANCH"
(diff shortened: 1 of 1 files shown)


scottisloud:
https://github.com/1Password/shell-plugins/pull/639 suggests the workflow logic is now correct, however, the commits it produces aren't signed. Investigating.....

scottisloud:
Ah, okay it looks like commit bafbf363 switched from a GITHUB_TOKEN to a PAT. However, commits by automations using a PAT are _not_ signed, and since that swap, we've never had signed commits from this automation. (I assume the switch from a GH token to a PAT was intended to work around GH's suppression of `on: pull_request` workflow triggers when using a GH token). 

A few ways forward that I'll raise internally.

scottisloud:
This is blocked until a new GH app is approved, which would allow workloads to make signed commits and PRs.

rishiy15 (approved):

PR #422, opened 2024-01-26
Add NPM plugin

alondahari:
## Overview
This is a second attempt at creating an NPM plugin, following https://github.com/1Password/shell-plugins/pull/168. 

Currently, this plugin supports:
- Importing existing project and user config files
- Multiple registries and scopes

It does not yet support:
- MFA
- importing global and builtin config files

## Type of change
 

- [x] Created a new plugin
- [ ] Improved an existing plugin
- [ ] Fixed a bug in an existing plugin
- [ ] Improved contributor utilities or experience

## How To Test

- `npm login` to generate an `.npmrc` file if you don't have an existing one
- init the plugin
- remove your `.npmrc` file from your home folder
- `npm whoami`

## Changelog
  
Authenticate the NPM CLI using Touch ID and other unlock options with 1Password Shell Plugins.

PR #658, opened 2026-08-08
Add npm and pnpm support

dalldrit13:
## Overview

This change introduces a 1Password plugin supporting npm and pnpm. Currently both these systems default is to store credentials on disk in plaintext. 1Password to the rescue!

## Type of change
 

- [x] Created a new plugin

## Related Issue(s)
  

Didn't see any related issues however similar PR [here](https://github.com/1Password/shell-plugins/pull/422) that looks to be stale from ~2 years ago.

## How To Test

Once built run `op plugin init npm` or `op plugin init pnpm`. Follow the steps and then you can verify functionality with `npm whoami` or `pnpm whoami`. 

_Note: You do need an npm account for testing this._

_File import should work if you've run `npm login` or `pnpm login` as these generate the plaintext file the plugin looks for. You can also generate granular access tokens in npm and paste those in the paste flow_

## Changelog
  

`Authenticate with npm (or pnpm) using the newly supported plugins`

Some additional notes for the technical writing team in the case this gets accepted and published.
- If a user is importing from a file they should probably be advised to remove the plaintext file once they've configured the plugin, (not strictly necessary functionally but kinda the whole idea of using the plugin is not having plaintext api keys lying about and the plugin ecosystem doesn't seem to provide any way of prompting the removal of these files at the CLI)
- Running `npm login` or `pnpm login` after configuring the plugin will not update your token in 1P but rather it will rewrite the token back into a plaintext file. `npm` will continue to use what's in 1P, and `pnpm` puts higher precedence on `auth.ini` so that will override our 1P token. I could not think of a clean way to manage this nor did I see an existing pattern in any other plugin. For something like `gh`, the CLI itself handles this case
  ```sh
  gh auth login 
  ? Where do you use GitHub? GitHub.com
  The value of the GH_TOKEN environment variable is being used for authentication.
  To have GitHub CLI store credentials instead, first clear the value from the environment.
  ```
  There should probably be guidance around this quirk and encourage users to create granular tokens in the web UI instead of the login commands. It's also worth noting in the past year npm has cranked down their default token expiry on tokens created using `npm login` to I believe 1 hour, so users running this often would continually hit this limitation.

### DIFF
--- a/plugins/npm/access_token.go
+++ b/plugins/npm/access_token.go
@@ -1,0 +1,232 @@
+package npm
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/1Password/shell-plugins/sdk"
+	"github.com/1Password/shell-plugins/sdk/importer"
+	"github.com/1Password/shell-plugins/sdk/provision"
+	"github.com/1Password/shell-plugins/sdk/schema"
+	"github.com/1Password/shell-plugins/sdk/schema/credname"
+	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
+)
+
+const defaultRegistry = "https://registry.npmjs.org/"
+
+func AccessToken() schema.CredentialType {
+	return schema.CredentialType{
+		Name:          credname.AccessToken,
+		DocsURL:       sdk.URL("https://docs.npmjs.com/about-access-tokens"),
+		ManagementURL: sdk.URL("https://www.npmjs.com/settings/~/tokens"),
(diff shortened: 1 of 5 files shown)


scottisloud (approved):
Hi @dalldrit13 
Thanks for your contribution! I've given it a review and all looks good. 

Fixed a minor lint issue and made a small change to overcome a CI-related failure you'd have no way of knowing about in your local environment. 

Approved!

This plugin will appear in a future release of the 1Password CLI! Keep an eye on [the changelogs](https://app-updates.agilebits.com/product_history/CLI2), and we _try_ to circle back and let you know when it's going live. 

LGTM! 🎉

PR #662, opened 2026-08-31
Update npm plugin to fix double temp file read issue

dalldrit13:
## Overview

I originally introduced this plugin in https://github.com/1Password/shell-plugins/pull/658

While recently using the plugin I discovered an issue with the temp file approach related to consuming more than once. 

This arose in version pinning of package managers.

When a version of npm or pnpm is specified inside the package.json

e.g.
```json
{
  "packageManager": "pnpm@11.24.0"
}
```

and this version differs from the global version on the system, the default behaviour is corepack downloads that missing version and spawns a child process using that version.

So the process looks like
```
1Password
  └─ pnpm 11.24 bootstrap process
       └─ pnpm 11.22 actual command process
```

The issue here is the temp file 1Password creates is only readable once as a stream. The parent process consumes it, then the child attempts to read it also and hangs indefinitely.

This also occurs with other double read cases (see testing)

This resolves that issue by instead leveraging env vars which do persist to the child process.

## Type of change

- [ ] Created a new plugin

## How To Test

With the previous version of the plugin built you can reproduce the issue with the following

### Repro Example 1

1. Create an empty dir
2. Add a package.json file
3. Populate with
    ```json
    {
      "name": "npm-fifo-repro",
      "version": "1.0.0",
      "private": true,
      "scripts": {
        "preinstall": "npm whoami"
      }
    }
    ```
4. Run `npm install`

### Repro Example 2 (pnpm)

assuming your global pnpm version !== 11.22.0 (otherwise just alter packageManager so not equal to global version)

1. Create an empty dir
2. Add a package.json file
3. Populate with
    ```json
    {
      "name": "pnpm-fifo-repro",
      "version": "1.0.0",
      "private": true,
      "packageManager": "pnpm@11.22.0"
    }
    ```
4. Run `pnpm whoami`

Both of these configurations should not hang after rebuilding this version of the plugin

## Changelog
  
Switch to env based approach for npm plugin, resolving file drain process hang issue.

### DIFF
--- a/plugins/npm/access_token.go
+++ b/plugins/npm/access_token.go
@@ -3,6 +3,7 @@ package npm
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/url"
 	"os"
@@ -11,7 +12,6 @@ import (
 
 	"github.com/1Password/shell-plugins/sdk"
 	"github.com/1Password/shell-plugins/sdk/importer"
-	"github.com/1Password/shell-plugins/sdk/provision"
 	"github.com/1Password/shell-plugins/sdk/schema"
 	"github.com/1Password/shell-plugins/sdk/schema/credname"
 	"github.com/1Password/shell-plugins/sdk/schema/fieldname"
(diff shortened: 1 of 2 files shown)


scottisloud (approved):
No concerns - Custom provisioner looks straightforward and aligns with many other existing provisioner patterns, so not requesting additional security review. 

LGTM

PR #666, opened 2026-09-17
Add workspace to tofu plugin needsauth and new tests

### LINKED ISSUE #618: tofu plugin: workspace subcommand does not trigger credential injection
The tofu plugin's credential NeedsAuth rules are missing needsauth.ForCommand("workspace"), which is present in the equivalent Terraform plugin.

```
 Current plugins/tofu/terraform.go:
 NeedsAuth: needsauth.IfAny(
     needsauth.ForCommand("refresh"),
     needsauth.ForCommand("init"),
     needsauth.ForCommand("state"),
     needsauth.ForCommand("plan"),
     needsauth.ForCommand("apply"),
     needsauth.ForCommand("destroy"),
     needsauth.ForCommand("import"),
     needsauth.ForCommand("test"),
     // "workspace" is missing
 ),
```

Because tofu workspace can switch between workspaces that use different provider credentials, it should behave identically to terraform workspace and inject credentials. The fix is to add needsauth.ForCommand("workspace") to the list.

AJaccP:
## Overview
The Terraform plugin requires the project credentials for `terraform workspace`, but the OpenTofu plugin's otherwise identical list of subcommands omits it, so `tofu workspace select staging` ran with nothing injected and no warning. This adds the missing `ForCommand("workspace")` rule, which covers every `workspace` subcommand, making the two plugins behave the same.

Neither plugin had any test coverage. This adds matching `NeedsAuth` tables to both, identical on purpose: the two command lists are hand-maintained copies of each other and this bug was them drifting apart, so either side drifting now fails a test.

## Type of change

- [ ] Created a new plugin

## Related Issue(s)

* Resolves: #618

## How To Test

Unit tests, including a regression test for the reported bug:

```bash
go test ./plugins/tofu/ ./plugins/terraform/ -v
```

`TestTofuCLIProjectCredentialsNeedsAuth` covers three `workspace` forms and guards that the nine already-working remote subcommands stay gated while `fmt` and `validate` stay exempt. `TestTerraformCLIProjectCredentialsNeedsAuth` asserts the same table against Terraform, as the parity guard. `TestTofuCLINeedsAuth` pins the executable-level gate, which passes both before and after this change — it is there so a future fix for a similar bug is not misapplied at that level.

End to end with the CLI:

```bash
op plugin init tofu
```

No specific credential should be needed. What we're checking is *when* 1Password is asked for the credential, not whether the OpenTofu command itself succeeds.

From a directory with a `.tf` configuration, these should now resolve the project credentials, where before this change they ran with nothing injected and no prompt:

```bash
tofu workspace list
tofu workspace select staging
tofu workspace new staging
```

Terraform has always behaved this way, so the two should now agree:

```bash
terraform workspace list
```

And these should still run with no credentials, which is the guard against the rule being too broad:

```bash
tofu fmt
tofu validate
```

## Changelog

The OpenTofu plugin now provides the project credentials to `tofu workspace` commands, matching the Terraform plugin's behaviour.


JillRegan (approved):
LGTM! Thank you!

PR #670, opened 2026-09-23
Fix example secrets for credential fields without a length

### LINKED ISSUE #215: Blank example credential when the field doesn't have a length defined
When working [the clilol shell plugin](https://github.com/1Password/shell-plugins/pull/214), I noticed that not defining the length for a field name results in an empty example credential (field to be precise):

```
➜  shell-plugins git:(add/clilol) make clilol/example-secrets
go run cmd/contrib/main.go clilol/example-secrets
API Key:
  Address: t1CNH9Qa1q3mKgpCIS4Lc2iXqj7itZDivf3eA6sjbs4KqhTdObx53AgFEXAMPLE
  Email:
  API Key: e0z0quc1iwq8fgf8ub7dos4xrexample
```

Here's how my field name is configured:

```go
{
	Name:                fieldname.Email,
	MarkdownDescription: "Your omg.lol account email address.",
	Secret:              false,
	Composition: &schema.ValueComposition{
		Charset: schema.Charset{
			Lowercase: true,
			Uppercase: true,
			Digits:    true,
			Specific:  []rune{'@', '.'},
		},
	},
},
```

Configuring the field name to have a length definitely fixes it, but I think it'd be a nice improvement to have some defaults.

AJaccP:
## Overview

`ExampleSecretFromComposition` sizes the random part of an example secret as `Length - len(prefix) - len(suffix)`, which assumes `Length` is always set. It's optional in `ValueComposition`, and correctly left unset for values that vary in length. With no `Length`, a field without a prefix produced an empty value, and a field with a prefix produced a negative length that panicked in `make([]byte, n)`. On current `main`, `<plugin>/example-secrets` prints blank values for 14 plugins (e.g. Snowflake) and crashes for 11 (e.g. PyPI).

When `Length` is unset, the value is now built additively: prefix, 24 random characters, then the `EXAMPLE` suffix when the charset has letters. The random part no longer depends on the prefix, so a long prefix can't consume it. When `Length` is set, output is unchanged except at the edges: `EXAMPLE` is only added if at least one random character still fits, and the random part is clamped at zero so a `Length` shorter than the prefix can't panic. All 61 existing fields that declare a `Length` still produce exactly that length.

## Type of change

- [ ] Created a new plugin

## Related Issue(s)

* Resolves: #215

## How To Test

Unit tests, including regression tests for the reported bug:

```bash
go test ./sdk/plugintest/ -v
```

End to end with the contributor tooling. These printed blank values or panicked before this change, and should now print a full value for every field:

```bash
make snowflake/example-secrets
make pypi/example-secrets
```

These declare a fixed `Length` and should print values of exactly that length, as before:

```bash
make github/example-secrets    # 40 characters
make aws/example-secrets       # 20 and 40 characters
```

## Changelog

`<plugin>/example-secrets` now generates example values for credential fields without a `Length`, instead of printing blanks or crashing.


JillRegan (approved):
LGTM!!

